PR #23, opened 2023-08-21
Update README.md

dmikey:
missing init, then update will pull ggml


leejet:
In fact, for those who clone the repository using `git clone --recursive https://github.com/leejet/stable-diffusion.cpp`, the `init` doesn't really make sense. However, to prevent any potential mistakes from people not including the `--recursive` parameter, I've decided to merge this PR.

dmikey:
Thanks! That's me! The guy that only follows half the directions :)

PR #38, opened 2023-08-30
flushes after printf

dmikey:
add flushes after `printf` statements, so that stdout will throw events and log in real time. 

if not flushed, stuck in buffer for other processes until this finishes. 

this fixes for example using `sd | tee output.txt`

fixes building UI tools without need for ABI/RPC (formatted logging probably ideal next)

Screenshot 2023-08-29 at 10 25 28 PM


fwsGonzo:
I also had to do this for a strange environment I am working in. I didn't create a PR because I don't know if this is something that everyone wants, but it helped me see how far I got with the memory constraints I have.

PR #50, opened 2023-09-07
Add Euler, Heun and DPM++ (2M) samplers

ursg:
I don't like ancestral samplers, so I implemented the simplest deterministic ones: Euler and Heun sampling, based on their python equivalents from the k-diffusers package.

This also includes code to do the ```--sampling-method``` parameter parsing. And fixes the spelling of Euler. :)

PR #51, opened 2023-09-08
Added option to switch the sigma schedule.

ursg:
In close connection to PR #50, this adds an option to switch to a different sigma schedule in the denoising process.

Concretely, this allows switching to the "Karras" schedule from the Karras et al 2022 paper, equivalent to the samplers marked as "Karras" in the AUTOMATIC1111 WebUI. This choice is in principle orthogonal to the sampler choice and can be given independently.


leejet:
Nice!

PR #56, opened 2023-09-12
Add DPM2 and DPM++(2s) a samplers

ursg:
This adds two more samplers, built by reimplementing the python code from [k_diffusion](https://github.com/crowsonkb/k-diffusion/blob/master/k_diffusion/sampling.py) in C++.

I'm actually not sure why DPM++(2s) a works at all, as [I think it does a negative timestep if the sigma goes to zero](https://github.com/leejet/stable-diffusion.cpp/commit/8f53bb4ee9b685ba4599c7696da7348587a15e79#diff-a8e875e8f7684a2f9e35212d88ee9574001952eb2c049385e73d5e24fd6e2a83R3795-R3797), but [as k_diffusion does the same](https://github.com/crowsonkb/k-diffusion/blob/master/k_diffusion/sampling.py#L525), I just left it as-is.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,8 +22,10 @@ Inference of [Stable Diffusion](https://github.com/CompVis/stable-diffusion) in 
     - `Euler A`
     - `Euler`
     - `Heun`
+    - `DPM2`
     - `DPM++ 2M`
     - [`DPM++ 2M v2`](https://github.com/AUTOMATIC1111/stable-diffusion-webui/discussions/8457)
+    - `DPM++ 2S a`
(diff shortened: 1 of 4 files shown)


JohnClaw:
Hope that someday you will add DDIM sampler.

ursg:
> Hope that someday you will add DDIM sampler.

I looked at that already, but I have to admit I don't understand it's code. Its implementation looks very different from the k-diffusers ones (and uses different variable names for the same(?) things).

leejet:
Good! Thank you for your contribution.

Issue #26, opened 2023-08-22
Feature request: Exif metadata


h3ndrik:
I think it's a great feature that tools like automatic1111 write all the parameters from inference into the images' metadata. That way you can re-discover old pictures on your harddrive and see what model hash, seed and prompt you used.

The Readme of the stb library mentiones this small lib:
* https://github.com/mayanklahiri/easyexif

(Edit: I think that library is just to read metadata, not write it :disappointed: )

leejet:
In fact, I'm considering adding parameter information to the generated images. However, I haven't found a good method for doing so yet.

Green-Sky:
I think we should be able to "write" the image to memory (using `stb_image_write`) and modify the memory afterwards to insert a `tEXt` chunk (automatic's ui inserts them at the front after `IHDR`)

[wikipedia excerpt](https://en.wikipedia.org/wiki/PNG#Ancillary_chunks):
> `tEXt` can store text that can be represented in [ISO/IEC 8859-1](https://en.wikipedia.org/wiki/ISO/IEC_8859-1), with one [key-value](https://en.wikipedia.org/wiki/Attribute%E2%80%93value_pair) pair for each chunk. The "key" must be between one and 79 characters long. Separator is a null character. The "value" can be any length, including zero up to the maximum permissible chunk size minus the length of the keyword and separator. Neither "key" nor "value" can contain null character. Leading or trailing spaces are also disallowed.

example hex screencap from file generated with automatic's ui:
![image](https://github.com/leejet/stable-diffusion.cpp/assets/2938071/6d551c87-543a-4025-9c10-d32a1c4a58a5)

PR #57, opened 2023-09-13
Write generation parameter exif data into output pngs.

ursg:
This adds prompt, negative prompt (if nonempty), seed, sampler and other generation parameters to the output file as a tEXt PNG block, in the same format as AUTOMATIC1111 webui does.

In order to keep everything free of external library dependencies, I have somewhat dirtily hacked this into the stb_image_write implementation.

Things that might still need discussion here:

- Should this be a command line parameter, or be always-on?
- Would it be better to keep stb_image_write unmodified, use stbi_write_png_to_memory and fudge the tEXt block into that manually in main.cpp?

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ Inference of [Stable Diffusion](https://github.com/CompVis/stable-diffusion) in 
     - [`DPM++ 2M v2`](https://github.com/AUTOMATIC1111/stable-diffusion-webui/discussions/8457)
     - `DPM++ 2S a`
 - Cross-platform reproducibility (`--rng cuda`, consistent with the `stable-diffusion-webui GPU RNG`)
+- Embedds generation parameters into png output as webui-compatible text string
 - Supported platforms
     - Linux
     - Mac OS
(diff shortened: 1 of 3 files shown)


FSSRepo:
I think it should always be included in the image; it doesn't take up space anyway. Whether you want to read it or not will depend on each person, but as a preventive measure, that information should always be included.

Green-Sky:
> * Would it be better to keep stb_image_write unmodified, use stbi_write_png_to_memory and fudge the tEXt block into that manually in main.cpp?

The best case would be to have support upstream (stb), but that is unlikely, to happen soon, even if you open a perfect pr there now :disappointed: .

I am move of an advocate to keep dependencies clean and work around the limit in the client code, maybe wrap it. I proposed your dirty approach [here](https://github.com/leejet/stable-diffusion.cpp/issues/26#issuecomment-1692115541) too :smile:

> * Should this be a command line parameter, or be always-on?

how about opt-out?

h3ndrik:
Sure. Make that always-on or opt-out. Makes little sense to accommodate just the few people who don't want to share their workflow or keep it a secret they did their homework assignment with ChatGPT ;-)

leejet:
Great!

PR #82, opened 2023-11-19
fix: some tensors are in bf16

Green-Sky:
numpy does not support bf16, so we conservatively upcast to f32.

I encountered some loras with (some?) tensors in bf16.


leejet:
Yes, I forgot to do it for lora, I did it for model weights before.

Green-Sky:
ah right. thanks for correcting :)

Green-Sky:
ah, you seem to down cast it to f16, have you seen any issues with that? I went with the safe route of upcasting to f32, since bf16 is a chopped-off f32.

leejet:
Nothing special, just consistency with the model. I also forget why I switched to an f16 in the first place, it seems like an f32 would be a better choice.

PR #83, opened 2023-11-20
Allow Loras with negative weight, too.

ursg:
There are a couple of loras which serve to adjust certain concepts in both positive and negative directions (like exposure, detail level etc).

The current code rejects them if loaded with a negative weight, but I suggest that this check can simply be dropped.


Green-Sky (approved):
Technically, we could discard == 0.f, but that usually does not happen anyway...

PR #87, opened 2023-11-26
sd : taesd faster decoding implementation

FSSRepo:
Unused


leejet:
> I don't know why the quality degrades a bit, probably because the weights of the convolutions are in f16, there must be a slight drop in quality, or it must be inherent to the decoder model.

Because this is a distilled model and the structure is different, it is normal that there are differences.

PR #89, opened 2023-11-27
StableDiffusion::setup instead of constructor

Jonathhhan:
Just a suggestion, because I am not sure about the drawbacks. For me it is quite handy to have a setup function instead of a constructor. https://forum.openframeworks.cc/t/use-setup-vs-constructor/41106

### DIFF
--- a/common/common.cpp
+++ b/common/common.cpp
@@ -370,7 +370,7 @@ std::string basename(const std::string& path) {
     return path;
 }
 
-const char* get_image_params(SDParams params, int seed) {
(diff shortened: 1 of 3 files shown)


leejet:
For a class with a std::shared_ptr member, it's not advisable to leave the std::shared_ptr uninitialized within the constructor. I prefer to provide a c api instead of modifying the constructor.

Jonathhhan:
Alright, I understand. Thanks.

PR #90, opened 2023-11-27
libstable-diffusion.so relative path

Jonathhhan:
Currently `libstable-diffusion.so` points to the path where libggml.so was created.
Its possible to change that to the relative path with this PR.


Jonathhhan:
It seems to work without if I build `libstable-diffusion` not as a shared lib: `libstable-diffusion.a`

PR #88, opened 2023-11-26
stable-diffusion: TAESD implementation - faster autoencoder

FSSRepo:
Fixes [#36](https://github.com/leejet/stable-diffusion.cpp/issues/36)

This is a quick implementation, so I may have overlooked something. Initially, I thought about implementing it in a separate header, but there were some issues due to dependencies collisions with ggml, and I honestly didn't want to bother solving that. Therefore, I directly added it to the `stable-diffusion.cpp` file.

### Results

| Props | AutoEncoder KL | TinyAutoEncoder |
|--------------|--------------|--------------|
| Test 1  | output_akl |   output_taesd |
| Test 2   | output_kl    | output_tae  |
| Processing Time (CPU Backend) | 24 seconds | 1.8 seconds |
| Params memory | 95 MB | 2 MB |
| Compute memory (512 x 512) | 1664 MB | 416 MB |

How to use it:

Just add `--taesd TAESD_MODEL_PATH` to command line, just works for txt2img for now:

```bash
git clone https://github.com/FSSRepo/stable-diffusion.cpp.git
mkdir build
cd build
cmake ..
cmake --build . --config Release

sd -m model.gguf -p "a lovely cat" --taesd taesd-model.gguf
```

Tasks:

- [x] Implement the unary operation `ggml_tanh` in CUDA, for full offloading support.
- [x] Complete the implementation of the encoder and perform tests.

PR #91, opened 2023-11-27
fix: reading memory of stack allocated object past its scope

Green-Sky:
I was having random crashes and corrupted png files over in #88 .
This pr fixes those


leejet:
Thanks for your feedback, it looks like I missed it in my review.

Issue #85, opened 2023-11-20
bug: tokenizer only matches exact words and no subwords


Green-Sky:
eg `spacestation` should probably be tokenized as `space` + `station</w>`, but right now it is just an unhandled token.
```
[DEBUG] stable-diffusion.cpp:1077 - parse 'spacestation' to [['spacestation', 1], ]
[DEBUG] stable-diffusion.cpp:469  - split prompt "spacestation" to tokens ["<|endoftext|>", ]
```

(resulting images have nothing to do with a spacestation)

Green-Sky:
Additionally, alot of loras define trigger words they are trained on. Not sure if they are in the lora and should be added to the vocab at runtime, or are just rare words not correctly tokenized (issue).

leejet:
Yes, because I haven't implemented bpe yet. I marked this as TODO in the code before

PR #119, opened 2023-12-19
feat: add bpe

Cyberhan123:
fix: https://github.com/leejet/stable-diffusion.cpp/issues/85

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -3,11 +3,10 @@ UseTab: Never
 IndentWidth: 4
 TabWidth: 4
 AllowShortIfStatementsOnASingleLine: false
-IndentCaseLabels: false
 ColumnLimit: 0
 AccessModifierOffset: -4
 NamespaceIndentation: All
-FixNamespaceComments: false 
+FixNamespaceComments: false
 AlignAfterOpenBracket: true
 AlignConsecutiveAssignments: true
 IndentCaseLabels: true
--- a/.clang-format
+++ b/.clang-format
@@ -3,6 +3,7 @@ UseTab: Never
 IndentWidth: 4
 TabWidth: 4
 AllowShortIfStatementsOnASingleLine: false
+IndentCaseLabels: false
 ColumnLimit: 0
 AccessModifierOffset: -4
 NamespaceIndentation: All
--- a/README.md
+++ b/README.md
@@ -51,7 +51,6 @@ Inference of [Stable Diffusion](https://github.com/CompVis/stable-diffusion) in 
     - The current implementation of ggml_conv_2d is slow and has high memory usage
     - Implement Winograd Convolution 2D for 3x3 kernel filtering
 - [ ] Continuing to reduce memory usage (quantizing the weights of ggml_conv_2d)
-- [ ] Implement BPE Tokenizer
 - [ ] Implement [Real-ESRGAN](https://github.com/xinntao/Real-ESRGAN/tree/master) upscaler
 - [ ] k-quants support
 
--- a/model.cpp
+++ b/model.cpp
@@ -1192,6 +1192,7 @@ ggml_type ModelLoader::get_sd_wtype() {
     return GGML_TYPE_COUNT;
 }
 
+
 bool ModelLoader::load_vocab(on_new_token_cb_t on_new_token_cb) {
     char* vocab_buffer          = reinterpret_cast<char*>(vocab_json);
     nlohmann::json vocab        = nlohmann::json::parse(vocab_buffer);
@@ -1204,6 +1205,22 @@ bool ModelLoader::load_vocab(on_new_token_cb_t on_new_token_cb) {
             token += decoder[c];
         }
         on_new_token_cb(token, token_id);
+    }
+    return true;
+}
+
+bool ModelLoader::load_merges(on_new_merges_cb_t on_new_merges_cb) {
+    char* vocab_buffer          = reinterpret_cast<char*>(vocab_json);
+    nlohmann::json vocab        = nlohmann::json::parse(vocab_buffer);
+    std::map<char, int> decoder = unicode_to_byte();
+    for (auto& it : vocab.items()) {
+        int token_id          = it.value();
+        std::string token_str = it.key();
+        std::string token     = "";
+        for (char c : token_str) {
+            token += decoder[c];
+        }
+        on_new_merges_cb();
     }
     return true;
 }
--- a/model.cpp
+++ b/model.cpp
@@ -1192,37 +1192,9 @@ ggml_type ModelLoader::get_sd_wtype() {
     return GGML_TYPE_COUNT;
 }
 
-
-bool ModelLoader::load_vocab(on_new_token_cb_t on_new_token_cb) {
-    char* vocab_buffer          = reinterpret_cast<char*>(vocab_json);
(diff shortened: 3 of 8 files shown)


leejet:
I don't think it's necessary to load the merge file, vocab is already ready, just use it. The only thing we need to do is implement the actual bpe function.

Cyberhan123:
Okay, I'll adjust it

Cyberhan123:
> I don't think it's necessary to load the merge file, vocab is already ready, just use it. The only thing we need to do is implement the actual bpe function.

I actually want to implement customized clip loading. For example, chinese-clip, I wonder if this idea is feasible?

leejet:
This clip model works if it can produce output with similar distribution to the clip model used by stable diffusion or if there is a dedicated unet model. I don't know if chinese-clip is suitable for stable diffusion, because I haven't studied its training process in detail.

Cyberhan123:
> I don't think it's necessary to load the merge file, vocab is already ready, just use it. The only thing we need to do is implement the actual bpe function.

@leejet  I just checked the built-in vocab.json. The vocab.json cannot repalce the function of merge.txt because it does not provide subword order. Another problem is that the token id of the bpe algorithm is the frequency of word occurrence, not a fixed mapping, so I am a bit worried that unexpected errors may occur when using the built-in vocabulary list.

leejet:
I've improved the tokenizer's handling of Unicode, generating vocab using merges information, similar to https://github.com/openai/CLIP/blob/main/clip/simple_tokenizer.py. Now you can easily the latest master branch into your PR, making it convenient to implement BPE. Please refer to https://github.com/openai/CLIP/blob/main/clip/simple_tokenizer.py for guidance.

FSSRepo:
I already have an implementation of embeddings (textual inversion) ready, and it works fine. However, it requires a specific way of loading the embedding from the prompt `bad quality, ugly, face malformed, bad anatomy, embd:bad-art-neg, embd:<embedding filename>` since custom embeddings are added to the end of `token_embed_weight`, and to create ''new tokens" that point to the position of the new embeddings is required.

Cyberhan123:
> I've improved the tokenizer's handling of Unicode, generating vocab using merges information, similar to https://github.com/openai/CLIP/blob/main/clip/simple_tokenizer.py. Now you can merge the latest master branch into your PR, making it convenient to implement BPE. Please refer to https://github.com/openai/CLIP/blob/main/clip/simple_tokenizer.py for guidance.

Thx

Cyberhan123:
Now We have completed all the logic, but I feel that the performance is not very good. After investigation, I found that `load_from_merges` does a lot of loops. Now we need to make a trade-off, whether we want faster loading speed or less memory.

If we want to improve the speed, we can add a built-in vocab.json, which will increase the memory by about 1MB. If we need to less memory, we can keep the current code.

Cyberhan123:
> I already have an implementation of embeddings (textual inversion) ready, and it works fine. However, it requires a specific way of loading the embedding from the prompt `bad quality, ugly, face malformed, bad anatomy, embd:bad-art-neg, embd:<embedding filename>` since custom embeddings are added to the end of `token_embed_weight`, and to create ''new tokens" that point to the position of the new embeddings is required.
> 
> Another issue is that it is unclear how other projects handle very long prompts of more than 77 tokens. Do they only use the first 77 and ignore the rest? There are some embeddings that require up to 75 tokens, and that covers almost the entirety of the available context.

@FSSRepo  Please check out the link below:
https://github.com/AUTOMATIC1111/stable-diffusion-webui/blob/cf2772fab0af5573da775e7437e6acdca424f26e/modules/sd_hijack_clip.py#L203

In summary, it will be split into multiple chunks

leejet:
> Now We have completed all the logic, but I feel that the performance is not very good. After investigation, I found that `load_from_merges` does a lot of loops. Now we need to make a trade-off, whether we want faster loading speed or less memory.
> 
> If we want to improve the speed, we can add a built-in vocab.json, which will increase the memory by about 1MB. If we need to less memory, we can keep the current code.

On my computer, load_merges takes 82ms, and load_vocab takes 53ms. There isn't a significant difference. These operations, akin to loading model weights, run only once, and the time spent compared to loading model weights is negligible. I believe there's no need to do this.

Cyberhan123:
We can merge this pr

FSSRepo:
My pull request is ready; however, a user is experiencing issues with the VAE tiling function for reasons unknown to me. I have tested it with images of various resolutions and haven't encountered problems with incomplete images.

Cyberhan123:
> My pull request is ready; however, a user is experiencing issues with the VAE tiling function for reasons unknown to me. I have tested it with images of various resolutions and haven't encountered problems with incomplete images.

I tested on mac and windows cuda and didn't encounter this problem

leejet:
> My pull request is ready; however, a user is experiencing issues with the VAE tiling function for reasons unknown to me. I have tested it with images of various resolutions and haven't encountered problems with incomplete images.

I've finished reviewing your PR and I'm ready to merge it, but I'm unsure whether Green-Sky will still encounter the issue.

leejet:
Once I've reviewed the changes in this PR, I might merge it first.

mwt11a:
I am finding this PR results in images that have very little relation to the prompt. Running on Linux and using CPU only. 

Even if a simple prompt like "photo of a cat next to a christmas tree, wearing a christmas hat".  Master will generate a image of a cat generally sitting next to a christmas tree and wearing a red santa type hat. While this PR will generate a woman wearing a woolly hat. Generally a close up head shot of her.

Cyberhan123:
> I am finding this PR results in images that have very little relation to the prompt. Running on Linux and using CPU only.
> 
> Even if a simple prompt like "photo of a cat next to a christmas tree, wearing a christmas hat". Master will generate a image of a cat generally sitting next to a christmas tree and wearing a red santa type hat. While this PR will generate a woman wearing a woolly hat. Generally a close up head shot of her.

I just tried it and it happened to me too. Let me see what caused it.

FSSRepo:
@Cyberhan123 Is this a complete implementation of BPE, or are there still things missing, as you could start deleting "Implement BPE Tokenizer" from the TODO section of the README?

Cyberhan123:
> @Cyberhan123 Is this a complete implementation of BPE, or are there still things missing, as you could start deleting "Implement BPE Tokenizer" from the TODO section of the README?

@FSSRepo  This is the complete implementation, I just made a low-level mistake, which has now been fixed.

Cyberhan123:
@mwt11a 
If you want, please pull the latest code and test it, it works fine for me.
```bash
./sd -m ../../models/sd-v1-4.ckpt -p "photo of a cat next to a christmas tree, wearing a christmas hat" --type q8_0
```
output

Cyberhan123:
emmm, the cat generated by this model looks a bit like Gargamel...

Cyberhan123:
@leejet 
image

Cyberhan123:
I am more accustomed to using JetBrains products, such as webstorm, goland, and clion. Maybe due to lack of experience, I didn't find a suitable formatting method, which caused extra work for you. I'm sorry.

FSSRepo:
I'm working with VSCode 😂😂😂 and a lot of extensions

leejet:
@Cyberhan123 I've completed reviewing the code in this PR, and it looks good to me. I'm planning to merge it now. Thank you for your contribution.

PR #127, opened 2023-12-27
Additional functionality

Cyberhan123:
I want to add functionality to our current library, Do you have any suggestions? @leejet @FSSRepo

- [x] Add custom log output callback
- [ ] Add termination logic during image generation
- [ ] Other models can be reloaded

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -3,7 +3,6 @@ UseTab: Never
 IndentWidth: 4
 TabWidth: 4
 AllowShortIfStatementsOnASingleLine: false
-IndentCaseLabels: false
 ColumnLimit: 0
 AccessModifierOffset: -4
 NamespaceIndentation: All
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,6 @@ test/
 .vscode/
 .cache/
 *.swp
-.vscode/
 *.bat
 *.bin
 *.exe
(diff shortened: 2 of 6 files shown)


Cyberhan123:
SDLogLevel::ERROR conflicts with the windows macro ERROR

Jonathhhan:
> SDLogLevel::ERROR conflicts with the windows macro ERROR

Not a very elegant solution, but in earlier versions (when it was part of stable-diffusion.cpp/.h) I replaced it with `SDLogLevel::SD_ERROR`.

Cyberhan123:
> > SDLogLevel::ERROR conflicts with the windows macro ERROR
> 
> Not a very elegant solution, but in earlier versions (when it was part of stable-diffusion.cpp/.h) I replaced it with `SDLogLevel::SD_ERROR`.

But this is the most practical solution. If we keep the original ERROR writing method, there will be bugs in set_sd_log_level on Windows.

Cyberhan123:
The latest code has achieved my needs

Cyberhan123:
Sorry, I think I missed the pretty_progress integration sd_set_log_callback thing, so I need to reopen pr to implement it

Cyberhan123:
This is not an urgent matter at the moment, I'll move it to a draft

PR #129, opened 2023-12-29
feat: Textual Inversion (embeddings)

FSSRepo:
I made a quick implementation to support embeddings. It's not the best way to do it, but when the code is refactored, the handling of embeddings could be improved. In some cases, there are very large embeddings that could consume beyond the available context and could lead to overflow or errors.

You can specify the embedding to use by adding the embedding filename to the prompt (usually negative prompt), separate it with `,`. Like this:

```bash
sd -p "a lovely cat" -n "EasyNegative, bad image" --embd-dir models/embeddings
```

**NOTE:** `EasyNegative` It uses a space of 75 tokens, so there are only 2 tokens left.

@leejet Any suggestions to improve this?

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ Inference of [Stable Diffusion](https://github.com/CompVis/stable-diffusion) in 
 - Faster and memory efficient latent decoding with [TAESD](https://github.com/madebyollin/taesd)
 - Upscale images generated with [ESRGAN](https://github.com/xinntao/Real-ESRGAN)
 - VAE tiling processing for reduce memory usage
-- VAE tiling processing for reduce memory usage
+- Textual Inversion support (embeddings)
 - Sampling method
     - `Euler A`
     - `Euler`
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ Inference of [Stable Diffusion](https://github.com/CompVis/stable-diffusion) in 
 - Faster and memory efficient latent decoding with [TAESD](https://github.com/madebyollin/taesd)
 - Upscale images generated with [ESRGAN](https://github.com/xinntao/Real-ESRGAN)
 - VAE tiling processing for reduce memory usage
+- VAE tiling processing for reduce memory usage
 - Sampling method
     - `Euler A`
     - `Euler`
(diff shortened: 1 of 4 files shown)


FSSRepo:
@Green-Sky Try to see if the errors are fixed with the last commit I made. You don't necessarily have to use the feature addressed in this PR.

diimdeep:
Why is max_position_embeddings = 77; limited like  this and how pytorch based versions deal with it, I remember that used ridiculous long prompts without problem  ?

diimdeep:
`-n 'embd:baddream, embd:unrealisticdream'` from https://civitai.com/models/72437/baddream-unrealisticdream-negative-embeddings

```ggml_new_object: not enough space in the context's memory pool (needed 57216, available 32768)```

FSSRepo:
@diimdeep As mentioned in the description, excessively large embeddings are not supported. I don't know how A1111 or ComfyUI manage very long prompts. Could you truncate them, something like splitting the embeddings in half and only including the first half?

leejet:
> Any suggestions to improve this?

I think we can take a cue from how sd-webui operates. Upon detecting that a token in the prompt corresponds to a local embedding file, we can replace the token with its embedding rather than explicitly specifying it using `embd:`. By the way, I'm currently refactoring the project, and once it's done, it should support long prompts.

FSSRepo:
@leejet I am thinking of merging this pull request with #131 to incorporate the new refactoring changes all at once.

Green-Sky:
@FSSRepo @slaren Since the sync AFTER set fixes the issue, i am pretty sure that slow-ish PCIe speeds or similar seem to break it. From what i can find online, it seems the sync needs to happen after the memcpy, since a different stream(?) might already be accessing the data before it is ready.

slaren:
That's probably the cause. I expected `cudaMemcpy` to be synchronous in all cases, but that's not what happens. The documentation says this:
>For transfers from pageable host memory to device memory, a stream sync is performed before the copy is initiated. The function will return once the pageable buffer has been copied to the staging memory for DMA transfer to device memory, but the DMA to final destination may not have completed. 

So `cudaMemcpy` may indeed return before the copy is completed, and a slow PCIe bus could cause the DMA transfer to not complete before the kernel is launched.

FSSRepo:
This changes were merged to #131

PR #133, opened 2023-12-31
refactor: reorganize code and use c api

leejet:
This PR made a lot of changes, including code reorganization and the addition of a C API. Despite the extensive modifications, I believe they are necessary to improve code readability and scalability.


util.cpp
```diff
@@ -164,40 +169,90 @@ std::string path_join(const std::string& p1, const std::string& p2) {
...
+    }
+
+    static char log_buffer[LOG_BUFFER_SIZE];
+
+    int written = snprintf(log_buffer, LOG_BUFFER_SIZE, "[%s] %s:%-4d - ", level_str, sd_basename(file).c_str(), line);
+
+    if (written >= 0 && written < LOG_BUFFER_SIZE) {
+        vsnprintf(log_buffer + written, LOG_BUFFER_SIZE - written, format, args);
+        strncat(log_buffer, "\n", LOG_BUFFER_SIZE - strlen(log_buffer) - 1);
+    }
```
Cyberhan123, line 220: Will this way of writing bury errors for the future?

Cyberhan123:
Wow, this change required a lot of effort

util.cpp
```diff
@@ -164,40 +169,90 @@ std::string path_join(const std::string& p1, const std::string& p2) {
...
+
+    if (sd_log_cb) {
+        sd_log_cb(level, log_buffer, sd_log_cb_data);
     }
 
     va_end(args);
 }
```
Cyberhan123, line 227: This way of writing will cause bugs in the future

util.cpp
```diff
@@ -164,40 +169,90 @@ std::string path_join(const std::string& p1, const std::string& p2) {
...
+        level_str = "WARN ";
+    } else if (level == SD_LOG_ERROR) {
+        level_str = "ERROR";
+    }
+
+    static char log_buffer[LOG_BUFFER_SIZE];
+
+    int written = snprintf(log_buffer, LOG_BUFFER_SIZE, "[%s] %s:%-4d - ", level_str, sd_basename(file).c_str(), line);
+
+    if (written >= 0 && written < LOG_BUFFER_SIZE) {
```
Cyberhan123, line 217:
```c++
void log_printf(sd_log_level_t level, const char* file, int line, const char* format, ...) {
    va_list args;
    va_start(args, format);

    const char* level_str = "DEBUG";
    if (level == SD_LOG_INFO) {
        level_str = "INFO ";
    } else if (level == SD_LOG_WARN) {
        level_str = "WARN ";
    } else if (level == SD_LOG_ERROR) {
        level_str = "ERROR";
    }

    static char log_buffer[LOG_BUFFER_SIZE];

    int written = snprintf(log_buffer, LOG_BUFFER_SIZE, "[%s] %s:%-4d - ", level_str, sd_basename(file).c_str(), line);

    if (written >= 0 && written < LOG_BUFFER_SIZE) {
        vsnprintf(log_buffer + written, LOG_BUFFER_SIZE - written, format, args);
        strncat(log_buffer, "\n", LOG_BUFFER_SIZE - strlen(log_buffer) - 1);
    }

    if (sd_log_cb) {
        sd_log_cb(level, log_buffer, sd_log_cb_data);
    }

    va_end(args);
}
```
This way of writing will cause bugs in the future

Cyberhan123: Because the current writing method only supports 1024-sized buffers, I want to replace the print in `pretty_progress` with the standard `log_printf` implementation in my pr, so that the log of pretty_progress can be received in the GUI

leejet: What kind of bugs will it cause? Could you provide more details?

Cyberhan123: If it is greater than 1024, the correct result will not be output, but currently it should be less than

leejet: When the log length exceeds 1023 characters, it gets truncated. This behavior is expected; it's not feasible to print logs of unlimited length.

Cyberhan123: ok

leejet:
I've tested the use of SD1.x/SD2.x/SDXL models on Windows/Linux and haven't encountered any issues. I'll merge this PR because I have other work depending on it.

PR #148, opened 2024-01-15
Enable `taesd` option again.

Jonathhhan:
This PR enables the `taesd` option again. Maybe it would be nice, to select `taesd` and `vae_tiling` without reloading the sd_ctx / model? `taesd` also seems to overwrite the vae model. `sdxl_vae-fp16-fix`, for example, has no effect, if `taesd` is selected.


leejet:
If you don't want to use taesd, just pass NULL to taesd_path. I think there's no need to add this parameter.

Jonathhhan:
@leejet not sure, if I make something wrong. But defining the `taesd_path` without setting `use_tiny_autoencoder` to true does not enable `taesd` for me.

Jonathhhan:
It actually works with the main branch, it just does not work with the control-net branch. Sorry for that.

PR #150, opened 2024-01-19
chore: improve ci

Cyberhan123:
This PR is mainly to improve the construction scope of ci, mainly for windows and mac. And I belive most of the people who use linux are developers, most of them are more professional than me. 

There are also some reasons：
1. There are also compatibility issues with glibc.
2. There are many linux distributions.
3. cuda construction will consume a lot of time (I also lack experience in implementing CUDA on Linux). 

Of course, if someone is willing, they can continue to improve in subsequent PRs. I have no objection.


.github/workflows/build.yml
```diff
@@ -30,26 +30,42 @@ jobs:
         with:
           submodules: recursive
 
-
       - name: Dependencies
         id: depends
         run: |
           sudo apt-get update
-          sudo apt-get install build-essential
+          sudo apt-get install build-essential zip
```
Green-Sky, line 37: you don't need zip, upload-artifact can do that for you.
```
    - uses: actions/upload-artifact@v3
      with:
        name: ${{ github.event.repository.name }}-x86_64
        path: |
          ${{github.workspace}}/build/bin/
```

.github/workflows/build.yml
```diff
@@ -30,26 +30,42 @@ jobs:
...
+
+      - name: Pack artifacts
+        id: pack_artifacts
+        if: ${{ ( github.event_name == 'push' && github.ref == 'refs/heads/main' ) || github.event.inputs.create_release == 'true' }}
+        run: |
+          cp ggml/LICENSE ./build/bin/ggml.txt
+          cp LICENSE ./build/bin/stable-diffusion.cpp.txt
+          zip -r sd-${{ env.BRANCH_NAME }}-${{ steps.commit.outputs.short }}-bin-${{ steps.system-info.outputs.OS_TYPE }}-${{ steps.system-info.outputs.OS_NAME }}-${{ steps.system-info.outputs.OS_VERSION }}-${{ steps.system-info.outputs.CPU_ARCH }}.zip ./build/bin/*
+
+
```
Green-Sky, line 68: you actually forgot the upload step here

.github/workflows/build.yml
```diff
@@ -140,9 +198,21 @@ jobs:
...
+          } elseif (Test-Path ".\build\bin\stable-diffusion.dll") {
+          $filePath = ".\build\bin\*"
+            echo "Exists at path $filePath"
+            Copy-Item ggml/LICENSE .\build\bin\ggml.txt
+            Copy-Item LICENSE .\build\bin\stable-diffusion.cpp.txt
+          } else {
+            ls .\build\bin
+            throw "Can't find stable-diffusion.dll"
+          }
+          7z a rwkv-${{ env.BRANCH_NAME }}-${{ steps.commit.outputs.short }}-bin-win-${{ matrix.build }}-x64.zip $filePath
```
Green-Sky, line 215: "rwkv", I guess we now know where the code is from :smile:

CMakeLists.txt
```diff
@@ -63,16 +63,21 @@ endif()
...
-add_library(${SD_LIB} stable-diffusion.h stable-diffusion.cpp model.h model.cpp util.h util.cpp upscaler.cpp
-             ggml_extend.hpp clip.hpp common.hpp unet.hpp tae.hpp esrgan.hpp lora.hpp denoiser.hpp rng.hpp rng_philox.hpp)
+set(SD_LIB_INCLUDE stable-diffusion.h stable-diffusion.cpp model.h model.cpp util.h util.cpp upscaler.cpp
+        ggml_extend.hpp clip.hpp common.hpp unet.hpp tae.hpp esrgan.hpp lora.hpp denoiser.hpp rng.hpp rng_philox.hpp)
 
-if(BUILD_SHARED_LIBS)
+# we can get only one share lib
+if(SD_BUILD_SHARED_LIBS)
     message("Build shared library")
+    set(BUILD_SHARED_LIBS OFF)
```
Green-Sky, line 72: this looks very counter intuitive.

CMakeLists.txt
```diff
@@ -63,16 +63,21 @@ endif()
...
+    message(${SD_LIB_INCLUDE})
+    add_library(${SD_LIB} SHARED ${SD_LIB_INCLUDE})
     add_definitions(-DSD_BUILD_SHARED_LIB)
     target_compile_definitions(${SD_LIB} PRIVATE -DSD_BUILD_DLL)
     set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 else()
     message("Build static library")
+    add_library(${SD_LIB} SHARED ${SD_LIB_INCLUDE})
```
Green-Sky, line 80: same here?

.github/workflows/build.yml
Cyberhan123: 😄

CMakeLists.txt
Cyberhan123: Hahaha, but this is the key to packaging all the code into a dynamic library, which will help me perform black magic in golang.
Another very important benefit is that you don’t have to worry about what version of ggml (avx2? cuda? commit hash?) it is. You only need to replace the dynamic library. Ideally, I can just download the dynamic library in the desktop app. Errors about sd.cpp can be fixed without having to worry about other dynamic libraries.

Cyberhan123:
Hi, @leejet , I have completed the upgrade to ci. It's waiting to be merged. But one thing I want to explain is that this build logic contains some **if judgment logic**. They are ignored in this PR, so there may be some problems merged into the master will be corrected. Of course I will follow up and make changes

.github/workflows/build.yml
Green-Sky: this might actually be wrong, since download artifact extracts them again, sorry for the confusion.

leejet:
Have you tested it in your master branch?

Cyberhan123:
> Have you tested it in your master branch?

Yes, I've tried this on my main branch, and with some adjustments, the build product is now being generated normally.
here is the latest release on my main branch: https://github.com/Cyberhan123/stable-diffusion.cpp/releases/tag/master-bd56734

leejet:
I think it's best to include the cudart dependency for Windows in the packaged build, just like what llama.cpp did, as shown in https://github.com/ggerganov/llama.cpp/releases.

Cyberhan123:
> There is another issue: stable-diffusion.h is a pure C header file, so it should not include anything C++. Therefore, the declarations of sd_basename and preprocess_canny need to be removed from this file. If your project relies on these two functions, I suggest implementing them separately.

The build products of ci are dynamic libraries and cli programs. If the dynamic library does not expose the two functions sd_basename and preprocess_canny, they need to be migrated to the cli program.

Cyberhan123:
> I think it's best to include the cudart dependency for Windows in the packaged build, just like what llama.cpp did, as shown in https://github.com/ggerganov/llama.cpp/releases.

There's certainly nothing wrong with that, I can make adjustments

leejet:
> The build products of CI are dynamic libraries and CLI programs. If the dynamic library does not expose the two functions sd_basename and preprocess_canny, they need to be migrated to the CLI program.

There is indeed an issue here. I merged the latest changes from the master branch and made some adjustments.

Cyberhan123:
The ci build of cuda and hip runs very slowly.  ：）Let me get up tomorrow morning and see the effect of adding cudart.

Cyberhan123:
https://github.com/Cyberhan123/stable-diffusion.cpp/releases/tag/master-346416b
cudart has in release now

PR #166, opened 2024-01-30
Img2img batch generation.

Jonathhhan:
It seems possible to do batch generation with `img2img` with those changes (better double check it, I just adapted it from `txt2img`).


leejet:
In the latest master branch, this has already been supported. Nevertheless, thank you for your PR.

PR #131, opened 2023-12-29
feat: Control Net support + Textual Inversion (embeddings)

FSSRepo:
I implemented a Control Net here; it's somewhat similar to the encoder phase of stable diffusion, but with some convolutional layers.

### Usage

```bash
./bin/sd --mode txt2img -p "a person" -m ../models/sd-v1-4.ckpt --control-net models/control_openpose-fp16.safetensors --control-image assets/control.png
```

### Some examples

| Conditioning | Result 1 | Result 2 |
|----------|----------|----------|
| none | output | output |
| control | output_controlnet | output_net2 |
| control_2 | output | output_net |
| control_3 | output | | 

**NOTE**: Requires more of 4 GB of VRAM, peak memory usage 3.6 GB (Out Of Memory in Windows). Use `--control-net-cpu` to keep Control Net with CPU backend

PR #174, opened 2024-02-12
Add in capability to repeatedly run the upscaler in a row

sean-bailey:
While the ESRGAN upscaler models compatible with stable-diffusion.cpp at this time are fixed in their upscaling rate, they do not discriminate on input image data. This means that, hypothetically, we could continuously run that upscaling model on the outputs of the previous upscaling run until we run out of memory.

I've tested this locally on my M1 MBP, and have successfully done txt2img and img2img generation resulting in image resolutions of 8192x8192. Bigger than that and I get an out-of-memory error, but hypothetically a large enough machine should be able to do this. 

This does *not* affect the image upscale multiplier inherent to the model, it simply runs the same fixed-multiplier model repeatedly (so for example, an `x4` model would take a 512x512 image upscaled to 2048x2048 on one pass, and then from 2048x2048 to 8192x8192 on a second pass, then 32k x 32k, then 128...)

The changes were designed to have minimal impact on any other code, so anything which currently leverages legacy versions should continue to function fine. If you want to use the new upscale repeating function, add

`--upscale-repeats <int >= 1>` 

to your command. Otherwise, by default, it will default to run the upscaler once.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,8 +94,6 @@ git submodule update
 #### Build from scratch
 
 ```shell
-git submodule init
-git submodule update
 mkdir build
 cd build
 cmake ..
--- a/README.md
+++ b/README.md
@@ -94,6 +94,8 @@ git submodule update
 #### Build from scratch
 
 ```shell
+git submodule init
+git submodule update
 mkdir build
 cd build
 cmake ..
@@ -148,7 +150,7 @@ cmake --build . --config Release
 ### Run
 
 ```
-usage: ./bin/sd [arguments]
+usage: ./build/bin/sd [arguments]
 
 arguments:
   -h, --help                         show this help message and exit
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -186,6 +189,7 @@ arguments:
                                      <= 0 represents unspecified, will be 1 for SD1.x, 2 for SD2.x
   --vae-tiling                       process vae in tiles to reduce memory usage
   --control-net-cpu                  keep controlnet in cpu (for low vram)
+  --canny                            apply canny preprocessor (edge detection)
```
phudtran, line 192: Is this supported in the PR?

sean-bailey: Canny wasn't something I modified in any way, I think it's how I modified the README itself that is showing the line addition for `--upscale-repeats`. Generally, functionality which was working before should still be working now as the upscaling process happens at the end, and this only repeats the upscale process.

leejet: This is introduced by https://github.com/leejet/stable-diffusion.cpp/pull/131

leejet:
Great! Thank you for your contribution.

PR #200, opened 2024-03-14
Addressed a double free bug when running on CPU introduced in the last commit as well as an issue loading SDXL LoRAs

grauho:
Fixed a double free issue when running multiple backends on the CPU, eg: CLIP and the primary backend, as this would result in the *_backend pointers both pointing to the same thing resulting in a segfault when calling the StableDiffusionGGML destructor.


az13js:
I have also encountered this issue. When using it on Windows 11, I encountered a "Segmentation fault".

leejet:
Great! Thank you for fixing this problem.

PR #204, opened 2024-03-18
do not free sd_ctx in load_from_file

Jonathhhan:
With this change I can load another model, when the sd_ctx is freed in load_from_model it crashes while loading a new model (I free the sd_ctx with `free_sd_ctx` first and create a new one for loading a new model).


Jonathhhan:
That did not solve my issue. First it was working, and now it crashes again...

Issue #207, opened 2024-03-19
photoMaker issue with two or more generated images / SDXL sample steps


Jonathhhan:
Very nice feature, thanks. It seems that only the first generated image works after loading the `sd_ctx` (multiple images work with batch size > 1).

I used the Newton images from the example and the prompt: "man img, man with futuristic clothes".

This is the first image:
![ofxStableDiffusion-2024-03-19-03-33-18](https://github.com/leejet/stable-diffusion.cpp/assets/41275844/e3906450-59bd-4cb4-8116-b879e4b5f265)

And this is the second image:
![ofxStableDiffusion-2024-03-19-03-34-24](https://github.com/leejet/stable-diffusion.cpp/assets/41275844/9e49113f-1e1c-49ef-a9d9-539fa6d7b040)

And with SDXL-Turbo photoMaker seems to need less than the fixed sample 50 steps...

Green-Sky:
It looks like the cfg scale was too high for the first image.

Jonathhhan:
@Green-Sky yes, I used 7 und not the recommended 5.

bssrdf:
@Jonathhhan, could you provide the full command line with SDXL and Photomaker model files? In particular, did you use the file from [https://huggingface.co/bssrdf/PhotoMaker](https://huggingface.co/bssrdf/PhotoMaker)? 

Here are what I can generate using Newton example images and your prompt with batch size 2. 

```
bin/sd -m ../models/RealVisXL_V3.0.safetensors  --stacked-id-embd-dir ../models/photomaker-v1.safetensors --input-id-images-dir examples/newton_man -p "man img, man with futuristic clothes"  --cfg-scale 7 --sampling-method euler -H 1024 -W 1024  -b 2 -o newton_issu01.png
``` 
![newton_issu01](https://github.com/leejet/stable-diffusion.cpp/assets/689043/03221e2b-5c94-4916-acfb-437d4024331e)
![newton_issu01_2](https://github.com/leejet/stable-diffusion.cpp/assets/689043/3dae9691-b15d-40fa-b1ba-1efc298eb5bc)

They look fine.

Jonathhhan:
@bssrdf batch processing works fine. The issue appears, if I run txt2img for a second time without reloading the sd_ctx. The console output looks exactly the same for both runs:
```
System Info:
    BLAS = 1
    SSE3 = 1
    AVX = 1
    AVX2 = 1
    AVX512 = 0
[...]

PR #208, opened 2024-03-21
apply pmid lora only once for multiple txt2img calls

bssrdf:
This PR fixed an [issue](https://github.com/leejet/stable-diffusion.cpp/issues/207) found by a user. The use case is to call txt2img multiple times and Photomaker lora's should **ONLY** be applied **once**.


leejet:

PR #214, opened 2024-04-02
Add Jellybox as UI using sd.cpp

phudtran:
I created a UI using sd.cpp

Only txt2img for now, but it supports both AMD and Nvidia with a 1 click install,


leejet:
This UI looks pretty good.

PR #219, opened 2024-04-04
fix tiles_ctx not freed in sd_tiling

null-define:
fix tiles_ctx not freed in sd_tiling.
if use esrgan, this will cause`ggml_init` in` sd_tiling` return failed after multiple times `txt2img` and `upscale` .


leejet:

PR #241, opened 2024-04-26
AYS, Align Your Steps, Scheduler Support

grauho:
Added NVIDIA's new "Align Your Steps" style scheduler in accordance with their quick start guide. Currently has handling for SD1.5, SDXL, and SVD, using the noise levels from their paper to generate the sigma values. Can be selected using the --schedule ays command line switch. Updates the main.cpp help message and README to reflect this option, also they now inform the user of the --color switch as well.


leejet:

PR #263, opened 2024-05-12
Prevent LoRA alpha scalars from being skipped over during model loading

grauho:
Adjusted model.cpp such that LoRA alpha scalars are not skipped for having an encoded dimension of zero.

### DIFF
--- a/model.cpp
+++ b/model.cpp
@@ -796,6 +796,10 @@ ggml_type str_to_ggml_type(const std::string& dtype) {
     return ttype;
 }
 
+static bool name_ends_with(const std::string name, const std::string suffix) {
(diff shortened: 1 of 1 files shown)


grauho:
Interestingly enough the problem this addresses isn't a problem for most LoRAs as the LoRA loading logic only needs to access the "alpha" scalars when the "scale" tensor is missing. That said I believe this addition is still valuable for two reasons. Firstly, edge cases that don't have the scale tensor may yield undesirable results if the alpha scalar tensor isn't loaded as observed with the SDXL Lightning LoRA. Secondly, this keeps the dimension check from spamming the users terminal with 'ERROR' log messages that might lead to confusion.

leejet:
Thank you for your contribution, but I believe the latest master branch has already addressed this issue and is more versatile, even functioning in Debug mode.

PR #269, opened 2024-05-21
Log fix

ring-c:


leejet:
These logs are used for debugging, and the relevant log printing has already been commented out. I think it's not necessary to delete them.

ring-c:
@leejet reverted "cleanup" part

leejet:

PR #274, opened 2024-05-28
Use std::from_chars instead of std::stof

snakelizzard:
It is better for locales where decimal separator differs from '.'


leejet:
`std::from_chars` was introduced in C++17, but sd.cpp has a minimum requirement of C++11, so this change is not suitable.

PR #278, opened 2024-05-29
Make crc32 100x faster on x86-64

jart:
This change makes checkpoints load significantly faster by optimizing pkzip's cyclic redundancy check. This code was developed by Intel and Google and Mozilla. See Chromium's zlib codebase for further details.


leejet:
Amazing work! Thank you for your contribution.

PR #285, opened 2024-06-11
Support for BREAK pseudo-token, to break clip chunks.

ursg:
Similarly to the other major implementations, this supports the user manually breaking the prompt into 77-token-chunks by adding a manual "BREAK" token.


grauho:
With this modified current regex this seems to delete all the white space in the prompt tokens.

leejet:
Replaced by https://github.com/leejet/stable-diffusion.cpp/pull/422. Thank you for your contribution.

PR #293, opened 2024-06-21
Add enum type to sd_type_t

drasticactions:
When building the stable-diffusion.h C headers, I get an error complaining about a missing enum type in the convert method

```
/home/drasticactions/Development/NativeBindings/library/stable-diffusion/stable-diffusion.h:198:92: error: must use 'enum' tag to refer to type 'sd_type_t'
thread 'main' panicked at src/main.rs:4:21:
called `Result::unwrap()` on an `Err` value: ClangDiagnostic("/home/drasticactions/Development/NativeBindings/library/stable-diffusion/stable-diffusion.h:198:92: error: must use 'enum' tag to refer to type 'sd_type_t'\n")
```

Adding the type fixes the error, and I think it should work with different compilers.


Green-Sky (approved):
Yes C needs the `enum` keyword.

grauho:
Yep I had to add the keyword as well when I tried linking from a C program, could it not also be const?

leejet:

PR #316, opened 2024-07-30
Add ays, clip skip and vae, to image embedded params

one-lithe-rune:
### Motivation

Since my own little [frontend](https://github.com/one-lithe-rune/diffuser-dials) app reads the embedded parameter info from PNG images to display them ready for editing and generating again, I noticed that the `Sampler` tag includes 'karras' if you have set that scheduler, but doesn't include 'ays' if you have set that. Since 'ays' has been added recently I'm guessing this is unintentional?

Since I was already in there to fix that, I also added `VAE` and `Clip Skip` to be written if vae_path or clip_skip have been set, since those are also useful for my app.

### Changes

- If 'ays' is set as the scheduler include it in the 'Sampler' tag in the data embedded into the final image. This makes it the same as how Karras is handled
- If a custom VAE path is set, include the vae name (without path and extension, using the same function as for Model) in embedded image params under a `VAE:` tag.
- If a custom Clip skip is set, include that Clip skip value in embedded image params under a `Clip skip:` tag.


wbruna:
Tested it locally; looks good. The exponential and gits schedulers are missing (probably added after this PR...), but they can be easily added in a follow-up commit.

leejet:
Thank you for your contribution. The corresponding changes have been merged in [743](https://github.com/leejet/stable-diffusion.cpp/pull/743).

PR #324, opened 2024-08-02
Add install section to CMake to allow linking

65a:
Mark stable-diffusion.h as a public header
Mark stable-diffusion.a as a public artifact
Install both of them in install step


65a:
Closing due to non-responsiveness of maintainer

Issue #325, opened 2024-08-02
v-prediction support


daniandtheweb:
Right now the program seems to only support eps-prediction. Some models use v-prediction and the lack of support for it causes the inference to generate noisy and unusable images with those models.

EDIT: it seems like there's support for it but the detection for stable diffusion 1.X isn't implemented.

leejet:
Which sd1.x models use v-prediction?

daniandtheweb:
Mainly the yiffymix models but I've found other 1.5 experimental models that use it that I can't remember right now.

Since there aren't a lot of models using it maybe an option to force the denoising mode manually would be enough.

PR #334, opened 2024-08-07
feat: add prediction argument

daniandtheweb:
This PR adresses the issue https://github.com/leejet/stable-diffusion.cpp/issues/325 by adding an argument to force a specific prediction mode. 

Some experimental models (such as [EasyFluff](https://huggingface.co/zatochu/EasyFluff) and similar models) may introduce different prediction modes from what's expected (Stable Diffusion 1.5 uses EPS but in this case the mentioned model uses v-predict) and this change allows using them correctly by manually adjusting the prediction type.


stduhpf:
Maybe it would make sense to also add support for the recently added `EDM_V_PRED` that is used by CosXL?

daniandtheweb:
I haven't been up to date with the recent development so I'm trying to catch up for now. Has it already been added to the repo right? I may be able to add support for it later.

stduhpf:
Yes it has been added with https://github.com/leejet/stable-diffusion.cpp/pull/683

Also there's a distinction between Flux Flow and Sd3 (Discrete) Flow:
https://github.com/leejet/stable-diffusion.cpp/blob/master/stable-diffusion.cpp#L584C1-L606C1

daniandtheweb:
I've added the new prediction methods. I'm unable to check them as I'm on a laptop right now but the logic is the same as with the others so it should work.

daniandtheweb:
@stduhpf Feel free to push this PR on your bleedingedge fork or let me know if you think I could improve the code in any way (I'm still not sure if the description is clear enough).
I'd really love at least a comment from the main dev but unfortunately he doesn't seem to be very active around here.

stduhpf:
I think I will re-create the bleedingedge branch from current master with the new c API (#728), I will try to include this PR when I do so.

daniandtheweb:
Will you include the image preview PR? I'm building a webui (more like a command generator + gallery) and that would be very helpful. I've been thinking of adding a toggle to switch between a mode that works on the mainline branch and another that works on your fork (I still need to add all the new features from mainline but at least that's the plan).

leejet:

PR #343, opened 2024-08-13
feat: added GITS scheduler

daniandtheweb:
I've implemented the [GITS](https://github.com/zju-pi/diff-sampler/tree/main/gits-main) scheduler.

With Stable-Diffusion 1.5 models seems to work nicely, I can't test more models right now as I'm currently working on a laptop and the resources are quite limited.


denoiser.hpp
```diff
@@ -178,6 +178,423 @@ struct AYSSchedule : SigmaSchedule {
     }
 };
 
+const std::vector<std::vector<float>> NOISE_LEVELS_0_80 = { \
```
Green-Sky, line 181: a couple of c++ things there.
1. you dont needs the tailing `\` , since you dont want to escape the new line
2. this looks like you wanted it to be a macro/define, in that case the `\` would be needed yes
3. dont put large data in headers. `#include <file/path>` works by copy pasting the whole file into the current translation unit

denoiser.hpp
```diff
@@ -178,6 +178,423 @@ struct AYSSchedule : SigmaSchedule {
...
+    { 14.61464119f, 2.36326075f, 1.24153244f, 0.803307f, 0.59516323f, 0.45573691f, 0.38853383f, 0.34370604f, 0.32104823f, 0.29807833f, 0.27464288f, 0.25053367f, 0.22545385f, 0.19894916f, 0.17026083f, 0.13792117f, 0.09824532f, 0.02916753f }, \
+    { 14.61464119f, 2.45070267f, 1.32549286f, 0.86115354f, 0.64427125f, 0.50118381f, 0.41087446f, 0.36617002f, 0.34370604f, 0.32104823f, 0.29807833f, 0.27464288f, 0.25053367f, 0.22545385f, 0.19894916f, 0.17026083f, 0.13792117f, 0.09824532f, 0.02916753f }, \
+    { 14.61464119f, 2.45070267f, 1.36964464f, 0.92192322f, 0.69515091f, 0.54755926f, 0.45573691f, 0.41087446f, 0.36617002f, 0.34370604f, 0.32104823f, 0.29807833f, 0.27464288f, 0.25053367f, 0.22545385f, 0.19894916f, 0.17026083f, 0.13792117f, 0.09824532f, 0.02916753f }, \
+    { 14.61464119f, 2.45070267f, 1.41535246f, 0.95350921f, 0.72133851f, 0.57119018f, 0.4783645f, 0.43325692f, 0.38853383f, 0.36617002f, 0.34370604f, 0.32104823f, 0.29807833f, 0.27464288f, 0.25053367f, 0.22545385f, 0.19894916f, 0.17026083f, 0.13792117f, 0.09824532f, 0.02916753f } \
+};
+
+
+const std::vector<std::vector<std::vector<float>>> NOISE_LEVELS = {
+    NOISE_LEVELS_0_80,
```
Green-Sky, line 510: This copies them all while still living on on their own. It would make sense to use the defines here or insert them fully here in the first place. It is also ok to have such generated tables in an external `table.inl` and pull it in using `#include` to better organize it.

denoiser.hpp
```diff
@@ -178,6 +178,423 @@ struct AYSSchedule : SigmaSchedule {
...
+    NOISE_LEVELS_1_25,
+    NOISE_LEVELS_1_30,
+    NOISE_LEVELS_1_35,
+    NOISE_LEVELS_1_40,
+    NOISE_LEVELS_1_45,
+    NOISE_LEVELS_1_50
+};
+
+
+struct GITSScheduler : SigmaSchedule {
```
Green-Sky, line 528: small nit, everything else is "Schedule" while you went with "Scheduler".

daniandtheweb: Didn't notice this, thanks.

denoiser.hpp
daniandtheweb: Since it will include only the gits noise levels should the name be something like gits_noise.inl or is it better something more generic?

Green-Sky: > something like gits_noise.inl

Yea I think that is fine.

daniandtheweb: I've just updated the code, does that seem better? I've also removed the `\` in the noise levels.

Green-Sky: Yea looks good. I still think the whole `denoiser.hpp` should be split into `.hpp` and `.cpp`, but thats out of scope for this pr.

denoiser.hpp
```diff
@@ -178,6 +179,72 @@ struct AYSSchedule : SigmaSchedule {
...
+            total_steps = static_cast<int>(std::round(n * sigma_max));
+        }
+
+        std::vector<float> sigmas;
+        int idx = static_cast<int>(std::round(1.20f * 10)) - 8; // Example adjustment for coefficient 1.20
+        if (n <= 20) {
+            sigmas = GITS_NOISE[idx][n - 2];
+        } else {
+            sigmas = GITS_NOISE[idx].back();
+            sigmas = log_linear_interp(sigmas, n + 1);
```
Green-Sky, line 238: you can optimize a copy of `sigams` away here by feeding `GITS_NOISE[idx].back()` directly into `log_linear_interp()`

gits_noise.inl
```diff
@@ -0,0 +1,349 @@
...
+    { 14.61464119f, 2.36326075f, 1.24153244f, 0.803307f, 0.59516323f, 0.45573691f, 0.38853383f, 0.34370604f, 0.32104823f, 0.29807833f, 0.27464288f, 0.25053367f, 0.22545385f, 0.19894916f, 0.17026083f, 0.13792117f, 0.09824532f, 0.02916753f },
+    { 14.61464119f, 2.45070267f, 1.32549286f, 0.86115354f, 0.64427125f, 0.50118381f, 0.41087446f, 0.36617002f, 0.34370604f, 0.32104823f, 0.29807833f, 0.27464288f, 0.25053367f, 0.22545385f, 0.19894916f, 0.17026083f, 0.13792117f, 0.09824532f, 0.02916753f },
+    { 14.61464119f, 2.45070267f, 1.36964464f, 0.92192322f, 0.69515091f, 0.54755926f, 0.45573691f, 0.41087446f, 0.36617002f, 0.34370604f, 0.32104823f, 0.29807833f, 0.27464288f, 0.25053367f, 0.22545385f, 0.19894916f, 0.17026083f, 0.13792117f, 0.09824532f, 0.02916753f },
+    { 14.61464119f, 2.45070267f, 1.41535246f, 0.95350921f, 0.72133851f, 0.57119018f, 0.4783645f, 0.43325692f, 0.38853383f, 0.36617002f, 0.34370604f, 0.32104823f, 0.29807833f, 0.27464288f, 0.25053367f, 0.22545385f, 0.19894916f, 0.17026083f, 0.13792117f, 0.09824532f, 0.02916753f }
+};
+
+const std::vector<std::vector<std::vector<float>>> GITS_NOISE = {
```
Green-Sky, line 331: i think you can reduce binary size further here by not copying and saving references instead.
`const std::vector<const std::vector<std::vector<float>>&>`

daniandtheweb: Done

daniandtheweb: Ok, now it's done.

Green-Sky: are you sure the reference does not work and you need to use a pointer?
(notice the `&` in `const std::vector<const std::vector<std::vector<float>>&>`)

Green-Sky: oh yea, nvm you cant have direct reference.

denoiser.hpp
```diff
@@ -59,14 +59,16 @@ std::vector<double> linear_interp(std::vector<float> new_x,
...
-    /* serves as the bounds checking for the below while loop */
-    if ((new_x[0] < ref_x[0]) || (new_x[new_x.size() - 1] > ref_x[ref_x.size() - 1])) {
-        LOG_ERROR("Linear Interpolation Failed: bad bounds");
-        return new_y;
+    /* Adjusted bounds checking to ensure new_x is within ref_x range */
+    if ((new_x[0] < ref_x[0])) {
+        new_x[0] = ref_x[0];
+    }
+    if ((new_x[new_x.size() - 1] > ref_x[ref_x.size() - 1])) {
+        new_x[new_x.size() - 1] = ref_x[ref_x.size() - 1];
```
Green-Sky, line 71: instead of `vec[vec.size() - 1]`  you can simply use `vec.back()`.

leejet:

PR #344, opened 2024-08-13
feat: added ipndm and ipndm_v samplers

daniandtheweb:
I've implemented the iPNDM and iPNDM_V schedulers from [here](https://github.com/zju-pi/diff-sampler/tree/main/diff-solvers-main).

The samplers are set to order 4 by default so it'll take a lot of steps to get a usable image. I've left that as that's also the default on ComfyUI.


leejet:

PR #346, opened 2024-08-14
feat: added exponential scheduler

daniandtheweb:
This pr implements the exponential scheduler. I've tested it and gives good results at high steps.

I think this pr is already ready to be merged if no one finds any issue.


leejet:

PR #359, opened 2024-08-23
add inplace conversion support for f8_e4m3

Green-Sky:
In the same way it is done for bf16.
Like how bf16 converts losslessly to fp32, f8_e4m3 converts losslessly to fp16.

(merge base right now is `flux`, @leejet tell me if I should retarget onto master)


leejet:

PR #358, opened 2024-08-23
Support for PhotoMaker  Version 2

bssrdf:
This PR added support for PhotoMaker V2 (PMV2).

PMV2 requires/adds  a face detection step via insightface. I couldn't find a C++ lib to do equivalent in SD. (There is a c++ SDK called InspireFace but it is pretty heavy and I don't think SD wants to add it as dependency).  So a workaround is to add an additional manual step: a python script was created to do face detection and generate resultant id_emebding. This id_emebeding is then fed into SD along with input images. 

See more in updated README.

|          |         |
|----    |----   |
| image | image |
| image | image |

PR #367, opened 2024-08-25
Very simple http server

stduhpf:
This is a very simple server that I made to be able to generate different prompts without reloading the models everytime.

Edit (11/19/25):
> full API specs (could get out of date eventually, let me know if I forget to update it) available here https://raw.githubusercontent.com/stduhpf/stable-diffusion.cpp/refs/heads/server/examples/server/server_doc.yaml
> You can import it in swagger editor to preview it with a nice UI.

Edit (03/01/25):
> Now with a basic GUI (/index.html endpoint for now).
> The API has changed a bit, so the rest of the old examples are outdated , I'll fix it later.

<details>

<summary>Mostly outdated instructions</summary>

## Starting the server

The syntax is pretty much the same as the cli. 
```
.\build\bin\Release\sd-server.exe --diffusion-model  ..\ComfyUI\models\unet\flux1-schnell-Q3_k.gguf --vae ..\ComfyUI\models\vae\ae.q8_0.gguf --clip_l ..\ComfyUI\models\clip\clip_l.q8_0.gguf --t5xxl ..\ComfyUI\models\clip\t5xxl_q4_k.gguf  -p "Default prompt" --cfg-scale 1.0 --sampling-method euler -v --steps 4 -o "server_output.png"
```

### Using the example client script

1. Make sure you have python with the modules `requests` and `pillow`are installed 
`pip install requests pillow`
2. Launch client in interactive mode
`python -i examples/server/test_client.py`

### Simplest setup

1. Make sure you have python installed with the `requests` module:
`pip install requests`
2. Open a python REPL:
`python`
3. Import the requests module
`>>> import requests`
4. Post your prompt directly to the `/txt2img`endpoints
`>>> requests.post("http://localhost:8080/txt2img","a lovely cat holding a sign says 'flux.cpp'")`
5. Images will be saved to disk on the server side, and each generation will overwrite the previous one.

### Using json payloads

1. Make sure you have python installed with the `requests` module:
`pip install requests`
2. Open a python REPL:
`python`
3. Import the requests and json modules
`>>> import requests, json`
4. Construct your json payload with generation parameters
`>>> payload = {'prompt': """a lovely cat holding a sign says "flux.cpp" """,'height': 768, 'seed': 42, 'sample_steps': 4}`
5. Post your payload to the `/txt2img`endpoints
`>>> requests.post("http://localhost:8080/txt2img", json.dumps(payload))`

### Decoding response using pillow

1. Make sure both `requests` and `pillow`are installed 
`pip install requests pillow`
2. Open a python REPL:
`python`
3. Import the requests, json and base64 modules
`>>> import requests, json, base64`
4. Import io.BytesIO and PIL.Image
`>>> from io import BytesIO` 
`>>> from PIL import Image`
5. Get the response from server
`>>> response = requests.post("http://localhost:8080/txt2img","a lovely cat holding a sign says 'flux.cpp'")`
6. Parse the response text as Json
`>>> parsed = json.loads(response.text)`
7. Decode base64 image data
`>>> pngbytes = base64.b64decode(parsed[0]["data"])`
8. Convert to PIL Image
`>>> image = Image.open(BytesIO(pngbytes))`
9. Display the image in default viewer
`>>> image.show()`

### One-liner

1. First import the necessary modules
`>>> import requests, json, base64`
`>>> from io import BytesIO` 
`>>> from PIL import Image`
 2. Use this line to send the request and open all the generated images.
 `>>> [Image.open(BytesIO(base64.b64decode(img["data"]))).show() for img in json.loads(requests.post("http://localhost:8080/txt2img",json.dumps( {'seed': -1, 'batch_count':4, 'sample_steps':4, 'prompt': """a lovely cat holding a sign says "flux.cpp" """} )).text)]`
 3. To send another payload after it's finished, press up arrow and you can edit the payload. 

If you don't want the image viewer to pause the execution of your command, you can do the following (not needed on macOS for some reason):
`>>> from threading import Thread`
`>>> [Thread(target=Image.open(BytesIO(base64.b64decode(img["data"]))).show, args=()).start() for img in json.loads(requests.post("http://localhost:8080/txt2img",json.dumps( {'seed': -1, 'batch_count':4, 'sample_steps':4, 'prompt': """a lovely cat holding a sign says "flux.cpp" """} )).text)]`
</details>

### DIFF
--- a/examples/server/main.cpp
+++ b/examples/server/main.cpp
@@ -73,7 +73,7 @@ struct SDCtxParams {
 
     rng_type_t rng_type               = CUDA_RNG;
     rng_type_t sampler_rng_type       = CUDA_RNG;
-    prediction_t prediction           = DEFAULT_PRED;
+    prediction_t prediction           = PREDICTION_COUNT;
     lora_apply_mode_t lora_apply_mode = LORA_APPLY_AUTO;
 
     bool offload_params_to_cpu   = false;


theaerotoad:
I'm excited about this one, and was attempting to combine with the Vulkan WIP pull.

I'm seeing a compile time issue (around the pingpong function) in my merge, and seems it's in the original as well.

```
stable-diffusion.cpp/examples/server/main.cpp:572:24: error: non-local lambda expression cannot have a capture-default
  572 | const auto pingpong = [&](const httplib::Request &, httplib::Response & res) {
      |                        ^
/home/aerotoad/software/aicpp/sd_vulkan_flux/server/stable-diffusion.cpp/examples/server/main.cpp: In lambda function:
/home/aerotoad/software/aicpp/sd_vulkan_flux/server/stable-diffusion.cpp/examples/server/main.cpp:672:5: warning: control reaches end of non-void function [-Wreturn-type]
  672 |     };
```

stduhpf:
> around the pingpong function

Ah! This function should go, I just added it at the start of devlopment to see if I was able to connect to the server. If it's causing issues, just remove it, and all the few things that depend on it.

stduhpf:
@theaerotoad just out of curiosity, which C++ compiler are you using? MSVC had no issue with this code (which I believe was technically incorrect).

theaerotoad:
> @theaerotoad just out of curiosity, which C++ compiler are you using? MSVC had no issue with this code (which I believe was technically incorrect).

Tested it on gcc 12.2.0-14 on Debian.

theaerotoad:
Yup, removing the pingpong endpoint allows compilation.

Another thought--the default 'localhost' string didn't work on my end initially.  Looks like `llama.cpp` server defaults to using 127.0.0.1 instead of 'localhost', so it might be worth setting the default string that way.  Not a big deal. though.

I was able to generate an image via requests, but segfault immediately afterwards.

```
[DEBUG] ggml_extend.hpp:977  - flux compute buffer size: 397.27 MB(RAM)
  |==================================================| 4/4 - 79.22s/it
[INFO ] stable-diffusion.cpp:1295 - sampling completed, taking 316.44s
[INFO ] stable-diffusion.cpp:1303 - generating 1 latent images completed, taking 316.44s
[INFO ] stable-diffusion.cpp:1306 - decoding 1 latents
[DEBUG] ggml_extend.hpp:977  - vae compute buffer size: 1664.00 MB(RAM)
[DEBUG] stable-diffusion.cpp:967  - computing vae [mode: DECODE] graph completed, taking 14.59s
[INFO ] stable-diffusion.cpp:1316 - latent 1 decoded, taking 14.59s
[INFO ] stable-diffusion.cpp:1320 - decode_first_stage completed, taking 14.59s
[INFO ] stable-diffusion.cpp:1429 - txt2img completed in 341.62s
save result image to 'server_output.png'
Segmentation fault
```

I've played around a bit (not much of a c++ coder at this point, and can't reliably track down where it's coming from, though.  I'm running with batch 1 (so only one image), and the first image gets written properly, with tags, then the dreaded segfault.

stduhpf:
Maybe you could try on the CPU backend to see if the segfault is related to the Vulkan merge or to the server itself? (Also you should probably use a less demanding model than flux when testing)

theaerotoad:
> Maybe you could try on the CPU backend to see if the segfault is related to the Vulkan merge or to the server itself? (Also you should probably use a less demanding model than flux when testing)

Right--should have said I ran the earlier example with the CPU backend (tried with no BLAS just to confirm it wasn't in my merging it over that caused this!)  It's _much_ faster with Vulkan.

I can confirm I seem to throw a   segfault with the server everytime with:

* CPU (no BLAS), from server branch with SDXL
* CPU (no BLAS), from server branch with Flux Schnell, q8 quants
* Vulkan, merged into [SkuttleOleg's](https://github.com/SkutteOleg/stable-diffusion.cpp) with Flux Schnell and q8 quants

For each of the above, they run fine with the main cli example (although painfully slowly on CPU).

stduhpf:
Hmm it doesn't happen on my machine, that's annoying to debug. I'll try running it on WSL to see if it's a linux thing.

stduhpf:
@theaerotoad I belive it's fixed now.

theaerotoad:
@stduhpf Yup, that fixes it.  Thank you!

Sure nice not to have to reload everything each time.

theaerotoad:
@stduhpf -- This is working pretty well, I played around with it a bit this weekend.  I have a few tweaks, to enable other inputs to be specified (via html form inputs) and returning the image as part of the POST command, and reduce CPU usage--use t.join() rather than while(1) at the end.

Do you want them? I may just share as a gist, or can branch off your repo.  What's your preference?

stduhpf:
@theaerotoad Both options are fine with me, thanks for helping. 

I thought about returning the image in base64 after each generation, but I was too lazy to implement it.

ring-c:
Working on some web-server myself, but with go backend. Its very early stage, and not for public use, but you can borrow [html/js frontend](https://github.com/ring-c/go-web-diff/blob/main/pkg/generate/index.html) for your project, for example.

stduhpf:
I just spent hours trying to understand why the server wasn't sending the image metadata as it is supposed to, turns out PIL automatically strips out the metadata, the server was working fine 🙃.

Green-Sky:
There are some differences to the automatic111 v1 webui api.
You use
- `sampling_steps` instead of `steps`
- `batch_count` instead of `batch_size`
- `sample_method` instead of `sampler_index`

This info however might be outdated, I just wanted to make my bot work with your api, so this just jumped at me.
We should look into what the other api's do (automatic and ComfyUI), and base it on that, to not make it incompatible unnecessarily.

stduhpf:
> There are some differences to the automatic111 v1 webui api. You use
> 
> * `sampling_steps` instead of `steps`
> * `batch_count` instead of `batch_size`
> * `sample_method` instead of `sampler_index`
> 
> This info however might be outdated, I just wanted to make my bot work with your api, so this just jumped at me. We should look into what the other api's do (automatic and ComfyUI), and base it on that, to not make it incompatible unnecessarily.

I might look into making the API compatible with other standards in the future. For now, I just use the same arguments as the `txt2img()` function declaration in stable-diffusion.h : https://github.com/leejet/stable-diffusion.cpp/blob/14206fd48832ab600d9db75f15acb5062ae2c296/stable-diffusion.h#L148-L164

stduhpf:
Speaking of, shouldn't the schedule method be specified when calling txt2img() rather than when creating the context?

Green-Sky:
> I might look into making the API compatible with other standards in the future. For now, I just use the same arguments as the txt2img() function declaration in stable-diffusion.h

I see.

> Speaking of, shouldn't the schedule method be specified when calling txt2img() rather than when creating the context?

I suppose.

Green-Sky:
If anyone just wants to run a command:
```bash
curl -sv --json '{"prompt": "a lovely cat", "seed": -1}' 127.0.0.1:7860/txt2img | jq -r .[0].data | base64 -d - > api_result.png
```

NNDam:
@stduhpf thanks for your work. Currently I'm using this pr for photomaker v2 but have error when change the input embedding (I believed it called "input_id_images_path"). How to input the different face without reload the whole SDXL model or some function to reload the face embedding ?

stduhpf:
@NNDam You can try with my lastest commit. I can't test it on my end, but it *should* work now?

NNDam:
@stduhpf thanks, I tried but still not work. The main problem is, at the first time load model, I also need to preload ```--input-id-images-dir```  extracted from script ```face_detect.py``` in [this PR](https://github.com/leejet/stable-diffusion.cpp/pull/358). But the embedding won't reload if I change the ```input_id_images_path``` when perform requests to server. It still output the face same with preloaded face at first time (and also **Segmentation fault** if the number of the current face differ with number of preloaded face)

stduhpf:
> @stduhpf thanks, I tried but still not work. The main problem is, at the first time load model, I also need to preload `--input-id-images-dir` extracted from script `face_detect.py` in [this PR](https://github.com/leejet/stable-diffusion.cpp/pull/358). But the embedding won't reload if I change the `input_id_images_path` when perform requests to server. It still output the face same with preloaded face at first time (and also **Segmentation fault** if the number of the current face differ with number of preloaded face)
Oh I see. Well, even if [Support for PhotoMaker Version 2](https://github.com/leejet/stable-diffusion.cpp/pull/358/files#top) was merged, I couldn't get this to work with the current architecture of the server. Have you tried with photomaker v1?

NNDam:
Hi @bssrdf, can you help us ?

bssrdf:
> Hi @bssrdf, can you help us ?

@NNDam, I'll see what can be done to make it work. Photomaker was developed following control net's workflow. It needs to be adjusted to work with this sever setup.

stduhpf:
I think some changes need to be made in stable-diffusion.cpp/stable-diffusion.h. Some arguments like scheduler type, vae settings, and controlnets are passed to the `new_sd_ctx()` function that load the models, but they should probably be passed to functions like `txt2img()`, `img2img()` and `img2vid()` instead.
That's completely out of scope for this PR though, but it would allow the server to support controlnet and photomaker v2.

bssrdf:
@NNDam , @stduhpf , I briefly looked at the server code. There may be a simple workaround for photomaker.

https://github.com/stduhpf/stable-diffusion.cpp/blob/d0704a536bae4904f9133ef0f1076ac8f7c44f0b/examples/server/main.cpp#L774

stduhpf:
@bssrdf That's exactly what I did in the last commit (https://github.com/leejet/stable-diffusion.cpp/pull/367/commits/d0704a536bae4904f9133ef0f1076ac8f7c44f0b): https://github.com/stduhpf/stable-diffusion.cpp/blob/d0704a536bae4904f9133ef0f1076ac8f7c44f0b/examples/server/main.cpp#L696. In theory this should work for photomaker v1 support (though I haven't tried it).

But Photomaker v2 support from your PR requires passing `params.input_id_images_path` as an arument to `new_sd_ctx()`, instead of just `txt2img()` .

CrushDemo01:
Hello, my program stops after every load model.

`
[INFO ] stable-diffusion.cpp:491  - total params memory size = 11434.90MB (VRAM 6483.59MB, RAM 4951.31MB): clip 4951.31MB(RAM), unet 6389.02MB(VRAM), vae 94.57MB(VRAM), controlnet 0.00MB(VRAM), pmid 0.00MB(RAM)
[INFO ] stable-diffusion.cpp:510  - loading model from '' completed, taking 38.26s
[INFO ] stable-diffusion.cpp:527  - running in Flux FLOW mode
[DEBUG] stable-diffusion.cpp:581  - finished loaded file
root@ucloud-wlcb-gpu-010:/text2img/serve_sdcpp/stable-diffusion.cpp# 
`
I didn't get any error or wrong feedback.

bssrdf:
> @bssrdf That's exactly what I did in the last commit ([d0704a5](https://github.com/leejet/stable-diffusion.cpp/commit/d0704a536bae4904f9133ef0f1076ac8f7c44f0b)): https://github.com/stduhpf/stable-diffusion.cpp/blob/d0704a536bae4904f9133ef0f1076ac8f7c44f0b/examples/server/main.cpp#L696. In theory this should work for photomaker v1 support (though I haven't tried it).
> 
> But Photomaker v2 support from your PR requires passing `params.input_id_images_path` as an arument to `new_sd_ctx()`, instead of just `txt2img()` .

Thanks for the information, @stduhpf.
I updated loading id_embeds to using raw binary tensor file load (using ```load_tensor_from_file```). It is more efficient to load this way since there is only 1 tensor. Now it should change/update id_embed based on the request and feeds photomaker V2. @NNDam, please retry my PR and let me know if there is still problems.

NNDam:
It worked !!! Thanks @bssrdf @stduhpf

nmandic78:
Any news on merging this?

Green-Sky:
The server does not deliver `index.html` when you omit it right now.

so `http://127.0.0.1:8080/` results in `404`.

ServeurpersoCom:
I must test it ASAP :)

I edit also for a ❤️
Just added a little heart because this work absolutely deserves it.
I’m currently deep into llama.cpp development and MCP tooling, but I’m really excited about this server mode. I can’t wait to expose an image-generation MCP tool powered by this stable-diffusion.cpp server.
This PR is opening exactly the door needed. Amazing work!

Green-Sky:
@stduhpf a rebase on master would be nice (:

edit: :heart:

stduhpf:
Updated to lastest sd.h API with previews, but it still doesn't support edit models, Wan, Qwen, controlnets, eta, timestep shift,  and I'm not sure about photomaker. 

I might come back to it and refactor it a bit to support more stuff.

Green-Sky:
> Updated to lastest sd.h API with previews

Thanks :pray: , now I can deploy my chat bot again. :smile:

Green-Sky:
We should probably split out argument parsing into an examples common file, like what llama.cpp did.

examples/server/main.cpp
```diff
@@ -50,20 +50,25 @@ struct SDCtxParams {
     std::string embeddings_path;
     std::string stacked_id_embeddings_path;
 
-    bool vae_decode_only = false;
-    bool vae_tiling      = false;
-
+    bool vae_decode_only = false;  // Does it ever make sense to set it to true?
```
wbruna: FWIW, Koboldcpp always sets it to `false` too. I think it'd only make sense to change it if there was also a flag to disable image to image (for reducing VRAM usage as much as possible).

wbruna (commented):
I'm getting "Please select a valid value." tooltips if CFG Scale or Distill Guidance have non-integer values, although the values do get sent correctly:

Image

Happens both with Firefox 140.5esr and Chromium 142.0.7444.162, on Linux.

examples/server/main.cpp
```diff
@@ -0,0 +1,2527 @@
...
+    return updatectx;
+}
+
+std::vector<std::string> list_files(const std::string& dir_path) {
+    namespace fs = std::filesystem;
+    std::vector<std::string> files;
+    if (dir_path != "")
+        for (const auto& entry : fs::recursive_directory_iterator(dir_path)) {
+            if (entry.is_regular_file()) {
+                auto relative_path   = fs::relative(entry.path(), dir_path);
```
wbruna: `fs::relative` follows symlinks, so a symlink `model.safetensors` inside the directory gets resolved to `../../../real/path/to/model.safetensors`. Could this be changed to something like

```c++
auto relative_path = entry.path().lexically_relative(dir_path);
```

instead?

stduhpf:
> I'm getting "Please select a valid value." tooltips if CFG Scale or Distill Guidance have non-integer values, although the values do get sent correctly:
> Image
> 
> Happens both with Firefox 140.5esr and Chromium 142.0.7444.162, on Linux.

This UI will need to be throughly reworked eventually anyways.

examples/server/main.cpp
```diff
@@ -0,0 +1,2527 @@
...
+            {"chroma_use_dit_mask", &params->ctxParams.chroma_use_dit_mask},
+            {"chroma_use_t5_mask", &params->ctxParams.chroma_use_t5_mask},
+        },
+        // manual_options (oh boy there are a lot)
+        {
+            {"model", [&](const json& o) -> bool {
+                 bool change     = false;
+                 int model_index = o.get<int>();
+                 if (model_index >= 0 && model_index < params->models_files.size()) {
+                     std::string new_path = params->models_dir + params->models_files[model_index];
```
wbruna: I'm getting invalid paths due to a missing trailing '/' on `params->models_dir`. These should probably be concatenated with `std::filesystem::path` (maybe even kept as path objects)?

Same issue with the two `diffusion_models_dir` below.

stduhpf:
I've had Gemini come up with a frontend based on the API docs, it seems pretty good, but It's way too bloated to qualify for being included in the server executable, I think. 

If anyone wants to use it, you can download it from: https://gist.github.com/stduhpf/ead10d9163cc032e0d2b2fae0acdbc19

And then run the server with `--custom-frontend-path [path to the frontend.html]` argument.

xxnuo:
https://github.com/xxnuo/stable-diffusion.cpp/blob/jetson/examples/server/main.cpp

I've written several HTTP server parameter patches here to catch up with the latest Z-Image parameters supported by the main branch. You can compare and merge them here if needed.

xxnuo:
By the way, I discovered that this library now supports calling C++ services from a Python server to achieve the same effect.

https://github.com/shakfu/cyllama/issues/1

examples/server/main.cpp
```diff
@@ -0,0 +1,3119 @@
...
+        end_task_json["steps"]  = 0;
+        end_task_json["eta"]    = "?";
+        std::lock_guard<std::mutex> results_lock(g_results_mutex);
+        g_task_results[task.task_id] = end_task_json;
+    }
+    g_running_task_id = "";
+}
+
+void worker_thread() {
+    while (!g_stop_worker) {
```
wbruna, line 2083: Not that I think it's _wrong_ as is, but I'm not a fan of having atomics mixed with mutexes: I think it makes the concurrent behavior harder to understand. And as long as the globals are always manipulated while holding the mutex, they don't need to be atomic; something like:
```
	lock
	while keep_going:
		if work:
			grab work, mark worker busy
			unlock
			process
			lock
			mark worker free
		else:
			wait (implicit unlock+lock)
	unlock
```
The non-blocking code paths can simply lock and unlock as needed.

examples/server/b64.cpp
```diff
@@ -0,0 +1,42 @@
+
```
wbruna, line 1: Maybe rename to `b64.hpp`?

examples/server/main.cpp
```diff
@@ -0,0 +1,3119 @@
...
+
+#else
+    std::vector<char*> c_args;
+    for (const auto& arg : args)
+        c_args.push_back(const_cast<char*>(arg.c_str()));
+    c_args.push_back(nullptr);
+
+    pid_t pid = fork();
+    if (pid == 0) {
+        execvp(c_args[0], c_args.data());
```
wbruna, line 3060: exec isn't really needed here: you are already on the child process, so calling `run_worker_mode` directly would work pretty much the same. I don't think Windows has a way to do that, though.

mpulukkinen:
Hi, do you need hand with the server code? I downloaded the pr and did some fixes. It thought it's less work than modifying the cli version, like I did last time I, when I integrated this to my app. 

I've fixed at least these: 
App did not start properly if some parameters were missing (not needed for server
App loses the current task after first poll -> makes impossible to keep polling the task state

Still in progress: I can't get the damm app to shut down when I call process.kill() from c# XD

stduhpf:
@mpulukkinen You could open a PR that targets [my server branch](https://github.com/stduhpf/stable-diffusion.cpp/tree/server). If I find the time I'll review it

mpulukkinen:
> @mpulukkinen You could open a PR that targets [my server branch](https://github.com/stduhpf/stable-diffusion.cpp/tree/server). If I find the time I'll review it

Did I do it right? https://github.com/stduhpf/stable-diffusion.cpp/pull/8

CoderTCY:
Is it capable with SD WebUI API now?

stduhpf:
Nope. If it's something that interests a lot of people I could try implementing (at least a subset of) sdapi, but if I'm not mistaken, koboldcpp already supports it.

wbruna:
> Nope. If it's something that interests a lot of people I could try implementing (at least a subset of) sdapi, but if I'm not mistaken, koboldcpp already supports it.

It does.

It may be a good idea to follow the sdapi anyway, since it's already something resembling a standard. But of course nothing would prevent the server to eventually implement that in addition to its own API.

Issue #382, opened 2024-08-29
Non-ASCII characters in prompt crashes the tokenizer 


stduhpf:
Using any non-ascii character in the prompt always cause the program to crash during prompt processing. The crash seems to be hapenning during this function call: https://github.com/leejet/stable-diffusion.cpp/blob/58d54738e26c9b793478c3aceafbbee0acf29405/conditioner.hpp#L357

```
> ./build/bin/sd -m ../ComfyUI/models/checkpoints/dreamshaper_8LCM.q8_0.gguf --cfg-scale 1 --steps 8 --sampling-method lcm --seed 42 -p "Un petit chaton très mignon" -v
Option: 
    n_threads:         12
    mode:              txt2img
    model_path:        ../ComfyUI/models/checkpoints/dreamshaper_8LCM.q8_0.gguf
    wtype:             unspecified
    clip_l_path:
    t5xxl_path:
    diffusion_model_path:
    vae_path:
    taesd_path:
    esrgan_path:
    controlnet_path:
    embeddings_path:
    stacked_id_embeddings_path:
    input_id_images_path:
    style ratio:       20.00
    normzalize input image :  false
    output_path:       output.png
    init_img:
    control_image:
    clip on cpu:       false
    controlnet cpu:    false
    vae decoder on cpu:false
    strength(control): 0.90
    prompt:            Un petit chaton très mignon
    negative_prompt:
    min_cfg:           1.00
    cfg_scale:         1.00
    guidance:          3.50
    clip_skip:         -1
    width:             512
    height:            512
    sample_method:     lcm
    schedule:          default
    sample_steps:      8
    strength(img2img): 0.75
    rng:               cuda
    seed:              42
    batch_count:       1
    vae_tiling:        false
    upscale_repeats:   1
System Info:
    BLAS = 0
    SSE3 = 1
    AVX = 1
    AVX2 = 1
    AVX512 = 0
    AVX512_VBMI = 0
    AVX512_VNNI = 0
    FMA = 1
    NEON = 0
    ARM_FMA = 0
    F16C = 1
    FP16_VA = 0
    WASM_SIMD = 0
    VSX = 0
[DEBUG] stable-diffusion.cpp:180  - Using CPU backend
[INFO ] stable-diffusion.cpp:195  - loading model from '../ComfyUI/models/checkpoints/dreamshaper_8LCM.q8_0.gguf'
[...]

PR #383, opened 2024-08-29
Fix tokenizer

stduhpf:
The type `char` is signed by default, this causes problems with characters having ASCII code greater than 127.

Fixes https://github.com/leejet/stable-diffusion.cpp/issues/382


Green-Sky (approved):
very good observation

leejet:

Issue #262, opened 2024-05-12
Any GUI / webui for this?


f1am3d:
Topic

grauho:
https://github.com/fszontagh/sd.cpp.gui.wx

This was posted in discussions the other day, haven't tried it yet but looks promising.

daniandtheweb:
If you're interested I've made a simple webui for interacting with the program.

https://github.com/daniandtheweb/sd.cpp-webui

JohnAlcatraz:
koboldcpp also uses stablediffusion.cpp: https://github.com/LostRuins/koboldcpp

Jonathhhan:
Here my Open Frameworks Addon: https://github.com/Jonathhhan/ofxStableDiffusion
The example is made with ofxImGui.

daniandtheweb:
It could be a good idea to dedicate a space in the README for the GUIs as in llama.cpp.

phudtran:
> It could be a good idea to dedicate a space in the README for the GUIs as in llama.cpp.

There is one. At the very bottom

daniandtheweb:
> > It could be a good idea to dedicate a space in the README for the GUIs as in llama.cpp.
> 
> There is one. At the very bottom

Oh, hadn't noticed it, thanks.

PR #387, opened 2024-09-01
Add a simple GUI to examples to manage parameters

piallai:
- Windows and Linux compatible
- Single option in CMakeLists.txt : <code>SD_EXAMPLES_GLOVE_GUI</code>
- Set CLI argument <code>-glove</code> to <code>sd</code> executable
- Requires only Qt to be installed
- Autosave parameters
- The GUI is easily modifiable : does not require GUI knowledge
   - Add parameters
   - Order of parameters
   - Group parameters
- Compatible with Qt5 and Qt6
- On Windows, depending on Qt installation, some DLLs may not be found:
   - go to bin/Release or bin/Debug, or wherever is sd.exe
   - windeployqt.exe .\sd.exe

Related to #262 

SD-glove


leejet:
I think it's better to put this gui in a separate repo, I can add a link to it in the readme.

PR #393, opened 2024-09-03
Docs: specify `models` directory for location of model weights downloads

a4v2d4:
I mostly got things working by just copying+pasting Bash commands shown in the `README`.

Only issue was:
- After building, the first example command(`./bin/sd -m ../models/sd-v1-4.ckpt -p "a lovely cat"`) failed because I didn't have the `models` dir nor anything in it.

Fix was:
- `mkdir ../models`
- `mv ../sd-v1-4.ckpt ../models/`

This just makes that explicit in the weight downloading setup.

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,8 @@ git submodule update
     - Stable Diffusion 3 2B from https://huggingface.co/stabilityai/stable-diffusion-3-medium
 
     ```shell
+    mkdir models
+    cd models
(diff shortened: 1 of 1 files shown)


grauho:
This program doesn't do any model downloading for you, the model directory path given in the README.md is merely an example. It could have been written more generally as follows: 
`./bin/sd -m <MODEL PATH> --prompt "A lovely cat"`

Here you would replace <MODEL PATH> with either the relative or absolute path to the model you want to use (and have previously downloaded).

PR #403, opened 2024-09-08
fix double prefix `model.diffusion_model.`

Green-Sky:
This happens because the tensor name might already have the prefix, but we unconditionally add it. Some finetunes/quants don't work otherwise.
example model: https://civitai.com/models/645943?modelVersionId=768009


Green-Sky:
Turns out it is only prefixed, when the model is loaded via `--diffusion-model`. So with those models, you can just use normal `-m`.

Issue #354, opened 2024-08-20
Image Preview


MGTRIDER:
Hi there developers, first of a big thank you for making this technology accessible to more people. I have a question, is there a way to preview the image being rendered, just like in other UI's? It would really save a lot of time if i could cancel a render if the result is undesirable. Thank you for your time.

daniandtheweb:
I've tried implementing it myself on a private repo by forcing every step to output a decoded image but unfortunately I haven't been able to properly implement it and the generation always broke after 3 steps. It shouldn't be hard to implement correctly, if I have time I may give it a try again.

MGTRIDER:
> I've tried implementing it myself on a private repo by forcing every step to output a decoded image but unfortunately I haven't been able to properly implement it and the generation always broke after 3 steps. It shouldn't be hard to implement correctly, if I have time I may give it a try again.

Thanks for having taken your time to give it a go. I think that such a feature would benefit us all. It would also save us on electricity costs and time as we wouldn't have to render out a "bad" seed each time and pray for luck, just to get good results.

leejet:
Image preview is currently supported.

PR #416, opened 2024-09-20
Image Preview

ring-c:
For big batches, if something went wrong - we loose all generated images. I want to get images as they a sampled. This way I can save content to disk and/or show preview as it get done.

Its also a step to #354, we can just call a `sd_result_cb` with additional parameters. This will take time to modify each sample method in `sample_k_diffusion()` so i do plan to make it as separate PR (if i can get it working).

What do you think about this changes? This PR is not breaking, just replacing functionality if callback is set.


ring-c:
Well, i added a callback for a sampling steps.
Its inefficient, with steps callback:
`txt2img completed in 64.58s`
Without:
`txt2img completed in 18.03s`
(Euler A, 32 steps)

amblamps:
> Well, i added a callback for a sampling steps. Its inefficient, with steps callback: `txt2img completed in 64.58s` Without: `txt2img completed in 18.03s` (Euler A, 32 steps)
> 
> Can look something like this: Screenshot_20240928_112203

Does the 64.58s include saving the images to disk?

ring-c:
@amblamps images are saved in this time, but impact is minimal. I will try to test different approaches to minimize times,  but most of the time is `decode_first_stage` call.

leejet:
Replaced by https://github.com/leejet/stable-diffusion.cpp/pull/522. However, I still want to express my gratitude for your contribution.

PR #422, opened 2024-09-27
Break pseudo token

daniandtheweb:
I've noticed that #285 hasn't been updated for a while so I've upstreamed the PR and fixed the regex to correctly identify the BREAK token without deleting the white spaces.
@ursg I've still marked you as the author as most of the work was yours, if you still want to update your PR instead I'll close this.

### DIFF
--- a/util.cpp
+++ b/util.cpp
@@ -594,7 +594,7 @@ std::vector<std::pair<std::string, float>> parse_prompt_attention(const std::str
         }
     };
 
-    std::smatch m,m2;
+    std::smatch m, m2;
     std::string remaining_text = text;
 
     while (std::regex_search(remaining_text, m, re_attention)) {
--- a/conditioner.hpp
+++ b/conditioner.hpp
@@ -270,8 +270,8 @@ struct FrozenCLIPEmbedderWithCustomWords : public Conditioner {
             const std::string& curr_text = item.first;
             float curr_weight            = item.second;
             // printf(" %s: %f \n", curr_text.c_str(), curr_weight);
-            int32_t clean_index          = 0;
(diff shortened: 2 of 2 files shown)


daniandtheweb:
Apparently I've messed something up during the code refactoring so this still doesn't work right now.

wbruna:
That 'B' stuff on the regex took me a bit too long to understand, so a comment would've been appreciated 🙂 Apart from that, the code looks good to me, and worked as expected on a few tests.

daniandtheweb:
Actually I should refactor this code to properly adapt it to the master branch. Right now it's a bit messy. I'll still add the original author as coauthor.

daniandtheweb:
I've finally remembered that I had this PR pending. I've just upstreamed it and it seems to work nicely.
@wbruna I added a comment on what does B do for easier understanding, let me know if it seems clear enough, I'm not really good at explaining regex.

util.cpp
```diff
@@ -513,6 +514,8 @@ sd_image_f32_t clip_preprocess(sd_image_f32_t image, int target_width, int targe
 //   (abc) - increases attention to abc by a multiplier of 1.1
 //   (abc:3.12) - increases attention to abc by a multiplier of 3.12
 //   [abc] - decreases attention to abc by a multiplier of 1.1
+//   BREAK - separates the prompt into conceptually distinct parts for sequential processing
+//   B - internal helper pattern; prevents 'B' in 'BREAK' from being consumed as normal text
```
wbruna, line 552: I think this is enough to make the regex a little less magical. Thanks!

leejet:

PR #426, opened 2024-09-30
Fix sample progress showing 

dsignarius:
Pretty_progress is called from sample within stablediffusion.cpp at line 707 It is called with a value of step=0

This shows that the sampling has been started and step 1 is running.. with this line here that does not happen.


dsignarius:
Withdrawing

PR #454, opened 2024-10-30
fast latent image preview

stduhpf:
This adds a callback to quickly convert latents to low resolution RGB images by projecting latent pixels to RGB directly. 

The preview images are low resolution, but they take almost no extra time or memory to compute. 

(not the same thing as https://github.com/leejet/stable-diffusion.cpp/pull/416)


electricazimuth:
+1 for this feature being in main

leejet:
Replaced by https://github.com/leejet/stable-diffusion.cpp/pull/522.

Issue #439, opened 2024-10-19
Corrupted images on Vukan backend


wbruna:
I'm getting either corrupted or inconsistent images on Vulkan backend, for any resolution other than 512x512.

My system is a Linux PC with a Ryzen 3400G, almost-vanilla Debian 12 (using the distro graphic stack). All following tests with:
`--type f16 --lora-model-dir ./LoRA --model ./SD/dreamshaper_8.safetensors --prompt 'a fantasy character, detailed background, colorful<lora:lcm-lora-sdv1-5:1>' --cfg-scale 1.0 --sampling-method lcm --steps 4 --rng cuda --seed 42 -b 1 --color` 
, and a script alternating resolution and compiled binary (Vulkan or CPU backend).

320x512:

Vulkan 1 |  Vulkan 2 | CPU
:----:|:----:|:----:
![320x512_vulkan1](https://github.com/user-attachments/assets/a4d15af4-4440-460f-b285-da8b10b2e083) | ![320x512_vulkan2](https://github.com/user-attachments/assets/53573927-a2d0-4e4c-a43e-05a93fc6a593) | ![320x512_cpu](https://github.com/user-attachments/assets/da803094-e2f1-40c4-91ab-15bb5fd728f3)

Vulkan images look ok-ish (for such a small resolution anyway), but the same seed should produce the same image. And the CPU render looks quite different.

Second test, 384x384; similar behavior (changes between Vulkan 1 and 2 may not be apparent on the thumbnail):

Vulkan 1 |  Vulkan 2 | CPU
:----:|:----:|:----:
![384x384_vulkan1](https://github.com/user-attachments/assets/fd4487fa-50ac-43b3-851b-a468584a996c) | ![384x384_vulkan2](https://github.com/user-attachments/assets/e43984f8-4d57-4177-bdd9-02847babfdb0) | ![384x384_cpu](https://github.com/user-attachments/assets/b000997a-a473-483f-aaa2-c63a88ca8957)

The third test, 448x448, gets weird:

Vulkan 1 |  Vulkan 2 | CPU
:----:|:----:|:----:
[...]

PR #455, opened 2024-11-01
Refactor: wtype per tensor from file instead of global

stduhpf:
I'm not sure if it makes a significant difference yet.

### DIFF
--- a/clip.hpp
+++ b/clip.hpp
@@ -545,9 +545,12 @@ protected:
     int64_t vocab_size;
     int64_t num_positions;
 
-    void init_params(struct ggml_context* ctx, ggml_type wtype) {
-        params["token_embedding.weight"]    = ggml_new_tensor_2d(ctx, wtype, embed_dim, vocab_size);
-        params["position_embedding.weight"] = ggml_new_tensor_2d(ctx, GGML_TYPE_F32, embed_dim, num_positions);
+    void init_params(struct ggml_context* ctx, std::map<std::string, enum ggml_type>& tensor_types, const std::string prefix = "") {
+        enum ggml_type token_wtype    = (tensor_types.find(prefix + "token_embedding.weight") != tensor_types.end()) ? tensor_types[prefix + "token_embedding.weight"] : GGML_TYPE_F32;
+        enum ggml_type position_wtype = GGML_TYPE_F32;  //(tensor_types.find(prefix + "position_embedding.weight") != tensor_types.end()) ? tensor_types[prefix + "position_embedding.weight"] : GGML_TYPE_F32;
+
+        params["token_embedding.weight"]    = ggml_new_tensor_2d(ctx, token_wtype, embed_dim, vocab_size);
+        params["position_embedding.weight"] = ggml_new_tensor_2d(ctx, position_wtype, embed_dim, num_positions);
     }
 
 public:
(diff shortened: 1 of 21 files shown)


stduhpf:
Trying master again after having these changes implemented for a while it feels like this reduces the loading times, but maybe there's somthing else affecting it.

Green-Sky:
well, it removes conversions. I had this today where i loaded flux with q8_0 t5 and f16 clip, and was wondering why t5 was using f16 (including ram usage). Turns out sd.cpp can only have one conditioner wtype rn...

fszontagh:
Maybe this is the wrong place where i write this, but i started a flux diffusion, which is loaded the model into VRAM, and started to make the diffusion on CPU. Please the the screenshot below: 

image

Green-Sky:
You probably saw the embedding models.

stduhpf:
I can confirm now, this PR makes loading weights very much faster for larger models.

(Results including warm runs only, so the models are always in disk cache)
| model (types) | master (model loading time) | PR (model loading time) |
| --- | --- | ---|
| flux1-schnell (q3_k, ae: q8_0, clip_l: q8_0, t5xxl: q4_k) | 92.37s | 6.08s |
| sd3.5_medium (q8_0, clip_l: f16, clip_g: f16, t5xxl: q4_k) | 18.19s | 5.37s |
| sd3.5_large (q4_0, clip_l: q8_0, clip_g: f16, t5xxl: q4_k) | 104.6 | 5.95 |
| SDXL (q8_0, vae: f16) | 2.70s | 2.72s |
| SD 1.5 (f16) | 1.42s | 1.42s |

I think this makes this PR worth merging.

Green-Sky:
But how do the other performance metrics change. Also does it all work?

stduhpf:
> Also does it all work?

I think so. Surprisingly, results aren't always exactly the same but it's very close nonetheless. I'm not sure what's up with that (maybe it has something to do with text encoders, or maybe it's just #439 since I'm using Vulkan). I have tried various txt2img models so far, they all worked. Didn't try img2img, photomaker, or controlnets yet though. 

> how do the other performance metrics change.

Diffusion/sampling and vae performance are within margin of error. Prompt encoding is significantly faster, when mixing quantizations.

stduhpf:
Ah, controlnets are not working, I'll see if I can fix.

stduhpf:
I think I got pretty much everything working at least as well as it used to with this refactoring now.  If anyone notices something I might have missed, lmk.

thxCode:
it introduces a tensor types mapper to guide the ggml type of each tensor, but i saw the conversion has filtered out these fixed tensors, is this an overkill implementation? https://github.com/leejet/stable-diffusion.cpp/blob/master/model.cpp#L1873-L1902

the root cause is using one `Conditioner weight type` to replace the weight types of various clip_l/clip_g/t5xxl, why do we still keep this wrong information `Conditioner weight type:     fp16`(clip_l: fp16, clip_g: fp16, t5xxl: q8_0)?

can we just refactor the `get_conditioner_wtype` of the model loader, make it able to extract the ggml type of each conditioner according to the sd_version, and then use the correct ggml type to initialize the different conditioners? looks like changing is less but achieving the same goals.

stduhpf:
> can we just refactor the get_conditioner_wtype of the model loader, make it able to extract the ggml type of each conditioner according to the sd_version, and then use the correct ggml type to initialize the different conditioners? looks like changing is less but achieving the same goals.

That's a fair point. It would indeed probably improve the loading times the same way without refactoring the whole thing. But the point of this PR was initially to refactor the model loading logic, the improvement in loading time for conditionning models is just a nice side effect.

My original motivation for doing this refactor was to be able to better support models with mixed quantization types (like those made with https://github.com/city96/ComfyUI-GGUF/blob/main/tools/convert.py or with https://github.com/leejet/stable-diffusion.cpp/pull/447). Now it also make it possible to implement https://github.com/leejet/stable-diffusion.cpp/pull/490 using the keys of the same tensor types map.

leejet:
I think the loading time optimization is just because you use the quantized t5 model, but get_conditioner_wtype doesn't recognize the quantized type correctly.

PR #457, opened 2024-11-11
Refactor: Replace manual memory management with smart pointers

ericcurtin:
Replaced all `malloc`/`free` calls with `std::unique_ptr` to leverage RAII for memory management. Used custom deleters where needed to handle specific free functions, such as `stbi_image_free` and `free_sd_ctx`. Simplified resource cleanup by removing explicit `free` calls, reducing the risk of memory leaks and improving code readability. Adjusted function calls to align with smart pointer usage, ensuring compatibility and preventing raw pointer access

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -119,7 +119,7 @@ add_subdirectory(thirdparty)
 
 target_link_libraries(${SD_LIB} PUBLIC ggml zip)
 target_include_directories(${SD_LIB} PUBLIC . thirdparty)
-target_compile_features(${SD_LIB} PUBLIC cxx_std_11)
+target_compile_features(${SD_LIB} PUBLIC cxx_std_14)
 
 
 if (SD_BUILD_EXAMPLES)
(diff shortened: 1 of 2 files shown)


MrGeRoI:
So true. Im keep getting memory leaks bc it allocates memory for images but im getting crash when i trying to free them

wbruna:
This has good intentions, but the PR has many unrelated and inappropriate changes, like this one:

[image](https://github.com/leejet/stable-diffusion.cpp/pull/457/commits/f7717260d3146cd1136f994193b8c6d659965706#diff-1d95451d72f080678e08278a6d33203dbc4d9502f688ab507f9ac806c8a78bddL738)

And things that look a lot like something an LLM would mess up (for instance, it's removing the whole img2vid call).

I realize this was submitted a long time ago, but if you can clean it up I'll review it again.

leejet:
Thanks for the PR!

I agree that using RAII/smart pointers is generally a good direction. However, this PR currently contains many unrelated and potentially risky changes beyond memory-management refactoring, so I don’t think it’s safe to merge in its current form.

Also, the codebase has gone through several rounds of refactoring since this PR was opened, so there are now significant conflicts with the current code. Some RAII patterns have already been introduced in the codebase as well.

Because of that, I’m going to close this PR for now. If you still see areas where RAII or memory management can be improved, please feel free to open a new PR based on the latest code, with a smaller and more focused change.

Thanks again for the contribution!

PR #460, opened 2024-11-13
Adds support for loading F8 e5m2 weights

LostRuins:
Adds support for loading F8 e5m2 weights, which is an alternative of f8 e4m3. Added in the similar manner as https://github.com/leejet/stable-diffusion.cpp/pull/359 by @Green-Sky  converting to f16.

Tested and seems to work.

### DIFF
--- a/model.cpp
+++ b/model.cpp
@@ -614,6 +614,48 @@ uint16_t f8_e4m3_to_f16(uint8_t f8) {
     return ggml_fp32_to_fp16(*reinterpret_cast<const float*>(&result));
 }
 
+
+uint16_t f8_e5m2_to_f16(uint8_t fp8) {
+    uint8_t sign = (fp8 >> 7) & 0x1;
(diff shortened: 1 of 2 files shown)


model.cpp
```diff
@@ -614,6 +614,48 @@ uint16_t f8_e4m3_to_f16(uint8_t f8) {
...
+        fp16_exponent = 0x1F;
+        fp16_mantissa = 0;
+    } else {
+        fp16_exponent = (uint16_t)true_exponent;
+        fp16_mantissa = mantissa << 8;
+    }
+
+    return fp16_sign | (fp16_exponent << 10) | fp16_mantissa;
+}
+
```
stduhpf, line 658: I might be mistaken, but can't this whole thing be replaced with just `return (uint16_t)fp8<<8;`, since fp8_e5m2 is basically truncated fp16?

stduhpf: (or rather `return static_cast<uint16_t>(fp8) << 8;`)

PR #461, opened 2024-11-14
Ability to split model components to different backend devices

softcookiepp:
The ability to specify separate backend device indexes for CLIP, Unet, and VAE was implemented. This works for all backends that accept a device index as an initialization parameter (CUDA, Vulkan, and SYCL.) The aim of this was to support splitting the parts of the model across multiple GPUs.

### DIFF
--- a/examples/cli/main.cpp
+++ b/examples/cli/main.cpp
@@ -5,6 +5,7 @@
 #include <random>
 #include <string>
 #include <vector>
+#include <stdexcept>
 
 // #include "preprocessing.hpp"
 #include "flux.hpp"
@@ -119,6 +120,10 @@ struct SDParams {
     bool canny_preprocess         = false;
     bool color                    = false;
     int upscale_repeats           = 1;
+    
+    int model_backend_index = -1;
+    int clip_backend_index = -1;
+    int vae_backend_index = -1;
 };
 
 void print_params(SDParams params) {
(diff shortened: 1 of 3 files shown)


Green-Sky (changes requested):

stable-diffusion.cpp
```diff
@@ -166,16 +170,22 @@ class StableDiffusionGGML {
...
-        for (int device = 0; device < ggml_backend_vk_get_device_count(); ++device) {
-            backend = ggml_backend_vk_init(device);
-        }
-        if (!backend) {
-            LOG_WARN("Failed to initialize Vulkan backend");
-        }
+        if (model_backend_index == -1) {
+			// default behavior, last device selected
+			for (int device = 0; device < ggml_backend_vk_get_device_count(); ++device) {
+				backend = ggml_backend_vk_init(device);
```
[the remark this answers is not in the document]
softcookiepp: I copied this from the original code verbatim in order to ensure that behavior prior to my changes was preserved. Though if this is somehow an issue, I will fix it.

softcookiepp:
Everything should be fixed now

examples/cli/main.cpp
```diff
@@ -5,6 +5,7 @@
 #include <random>
 #include <string>
 #include <vector>
+#include <stdexcept>
```
Green-Sky, line 8: There are no exceptions used.

Green-Sky (changes requested):

examples/cli/main.cpp
softcookiepp: Done

softcookiepp: Anything else that needs changing?

PR #468, opened 2024-11-19
Allow building ggml as a separate shared lib

cmdr2:
Hi, this PR allows the creation of two separate libs (e.g. ggml.dll and stable-diffusion.dll) via a build option (`SD_BUILD_SHARED_GGML_LIB`). By default, it continues to make a single shared lib (like it does today).

`cmake -B build -DSD_BUILD_SHARED_LIBS=1 -DSD_BUILD_SHARED_GGML_LIB=1`

This option is useful when distributing updates to programs that use sd.cpp internally. For e.g. when targeting CUDA, with this PR, new versions of `stable-diffusion.cpp` can be distributed with a 4 MB DLL update, instead of re-distributing 347 MB (that includes ggml with compiled ggml CUDA kernels).

ggml doesn't need to change frequently, so there isn't much point wasting bandwidth redistributing the entire static-ly built ggml (with cuda), each time `stable-diffusion.dll` needs to be updated.

I've tested this on Windows, Mac and Linux (Ubuntu).

Please let me know if this PR needs anything else, thanks!


ItzCrazyKns:
+1

wbruna:
LGTM; tested on a few local builds.

leejet:

PR #473, opened 2024-11-20
update ggml to 2024-10-03

Green-Sky:


stable-diffusion.cpp
```diff
@@ -47,6 +48,8 @@ const char* sampling_methods_str[] = {
     "LCM",
 };
 
+static_assert((int)SD_TYPE_COUNT == (int)GGML_TYPE_COUNT);
```
Green-Sky, line 51: sadly a c++17 feature...

Green-Sky:
Looks like newer ggml failes to compile on older rocm, because of a compiler bug.

We need to solve this by updating the hip sdk (like llama.cpp).

Will ignore for now.

Green-Sky:
Ok, changes are upsteam now

Green-Sky:
Including the failing rocm build :)

PR #484, opened 2024-11-26
vae tiling improvements: encoding support and adaptative overlap 

### LINKED ISSUE #588: Over-exposed parts of generated images at some resolutions with VAE (or TAE) tiling, and upscaler
I noticed this problem some time ago, But I forgot to make an issue about it. It's very noticable with Tiled TAESD and with upscaler, not as much with tiled VAE.

It happens with tiled VAE when the resolution isn't a multiple of 128 in both dimensions, with tiled TAEsd when the resolution isn't a multiple of 256 in both dimensions, and with the upscaler when the resolution isn't a multiple of 512 in both dimensions. (so 4x the tile size?)

Example: 

- SD1.x, 512x576 resolution : both tiled autoencoders and upscaler are broken

| VAE | Tiled VAE | TAESD | Tiled TAESD | 
| --- | --- | --- | --- | 
| VAE | VAE Tiling | TAE | Tiled TAE |

| VAE + upscaler | 
| --- |
| RealESRGAN_x4 |

-  SD1.x, 512x640 resolution  : only tiled taesd and upscaler are broken

| VAE | Tiled VAE | TAESD | Tiled TAESD | 
| --- | --- | --- | --- | 
| Image | Image | Image | Image |

| VAE + upscaler | 
| --- |
| Image |

-  SD1.x, 512x768 resolution : only upscaler is broken 

| VAE | Tiled VAE | TAESD | Tiled TAESD | 
| --- | --- | --- | --- | 
| Image | Image | Image  | Image | 

| VAE + upscaler | 
| --- |
| Image |

(I swear it's using the exact same prompt as the other examples)

-  SD1.x, 512x512 resolution: control, everything is fine

| VAE | Tiled VAE | TAESD | Tiled TAESD | 
| --- | --- | --- | --- | 
| Image | Image | Image | Image |

| VAE + upscaler | 
| --- |
| Image |

stduhpf:
- Added a mechanism to automatically fix the tile overlaps in cases of mismatch between image size and tile size. 
- Added a way to change tile size for VAE decoding with the `SD_TILE_SIZE` environment variable. (tile size for vae encoding will be about 1.3x bigger, for the same memory requirement)

With these changes, vae encoding can now be tiled, and a small (unreported?) issue with the upscaler (or tae) over-brightening some parts of the image is now fixed. This could also be a step toward tiled diffusion support. 

Fixes https://github.com/leejet/stable-diffusion.cpp/issues/588,  https://github.com/leejet/stable-diffusion.cpp/issues/564 and  https://github.com/leejet/stable-diffusion.cpp/issues/666

### DIFF
--- a/stable-diffusion.cpp
+++ b/stable-diffusion.cpp
@@ -349,9 +349,9 @@ public:
                                                                      offload_params_to_cpu,
                                                                      model_loader.tensor_storages_types);
                 diffusion_model  = std::make_shared<MMDiTModel>(backend,
-                                                               offload_params_to_cpu,
-                                                               sd_ctx_params->diffusion_flash_attn,
-                                                               model_loader.tensor_storages_types);
+                                                                offload_params_to_cpu,
+                                                                sd_ctx_params->diffusion_flash_attn,
+                                                                model_loader.tensor_storages_types);
             } else if (sd_version_is_flux(version)) {
                 bool is_chroma = false;
                 for (auto pair : model_loader.tensor_storages_types) {
@@ -396,11 +396,11 @@ public:
                                                                     1,
                                                                     true);
                 diffusion_model  = std::make_shared<WanModel>(backend,
-                                                             offload_params_to_cpu,
-                                                             model_loader.tensor_storages_types,
-                                                             "model.diffusion_model",
-                                                             version,
-                                                             sd_ctx_params->diffusion_flash_attn);
+                                                              offload_params_to_cpu,
+                                                              model_loader.tensor_storages_types,
+                                                              "model.diffusion_model",
+                                                              version,
+                                                              sd_ctx_params->diffusion_flash_attn);
                 if (strlen(SAFE_STR(sd_ctx_params->high_noise_diffusion_model_path)) > 0) {
                     high_noise_diffusion_model = std::make_shared<WanModel>(backend,
                                                                             offload_params_to_cpu,
@@ -1306,7 +1306,15 @@ public:
         ggml_tensor* result = NULL;
         if (!use_tiny_autoencoder) {
             process_vae_input_tensor(x);
-            first_stage_model->compute(n_threads, x, false, &result, work_ctx);
+            if (vae_tiling && !decode_video) {
+                // split latent in 32x32 tiles and compute in several steps
+                auto on_tiling = [&](ggml_tensor* in, ggml_tensor* out, bool init) {
+                    first_stage_model->compute(n_threads, in, true, &out, NULL);
+                };
+                sd_tiling(x, result, 8, 32, 0.5f, on_tiling, false);
+            } else {
+                first_stage_model->compute(n_threads, x, false, &result, work_ctx);
+            }
             first_stage_model->free_compute_buffer();
         } else {
             tae_first_stage->compute(n_threads, x, false, &result, work_ctx);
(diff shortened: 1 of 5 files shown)


Green-Sky:
>  * Added a mechanism to automatically fix the tile overlaps in cases of mismatch between image size and tile size.
> 
> 
> With these changes, vae encoding can now be tiled, and a small (unreported?) issue with the upscaler over-brightening some parts of the image is now fixed. This could also be a step toward tiled diffusion support.

Nice! It would also be interesting to see larger tile sizes. Depending on VAE, larger tiles would still fit into the memory budget and would speed things up.

stduhpf:
> The last_x used in the loop at [line 568](https://github.com/leejet/stable-diffusion.cpp/blob/d46ed5e184b97c2018dc2e8105925bdb8775e02c/ggml_extend.hpp#L568) (or line 629 in this PR) in ggml_extend.hpp is only initialized outside the loop.
Locally, it seems that this PR with initializing last_x = false; before the for (int x = 0; ~ loop would improve the generation results.

I'm sorry, I don't quite understand what you mean. `last_x` is already set to `false` when the loop starts.

> However, the brightness issue (issue https://github.com/leejet/stable-diffusion.cpp/issues/588) seems to remain.

You mean this PR doesn't fix it for you? If so this is strange, I can't reproduce the issue with this PR...

| master | PR |
| --- | --- | 
| tae_tiling_576 | tae_tiling_576 |

As you can see the image made with the master branch is much brighter than the one with this PR.
 (ignore the slight variation of the composition, it's just a little unrelated bug of the vulkan backend that happens at non standard resolutions)

stduhpf:
> I think I have misled you, but there is no doubt that this PR is very good, as you have demonstrated in your examples. What I meant to say is that in your right cat PR image, for example, the position around the neck is slightly whiter than without vae-tiling; if it's a technical problem with the GGML backend or tiling (I usually use the vulkan backend), I'll find a better way.

The result with tiled VAE is always going to be slightly lower quality than the full VAE, because itt's lacking information about the whole image. When using TAE, the results are almost completely identical witth the PR, while on master, there is the "overexposure" issue on the tiled TAE.

wbruna:
@leejet , just to let you know: this PR has been included in Koboldcpp since a few releases ago (1.94 - 1.96.2) to fix #588, with no issues reported so far, so it should be safe to merge.

wbruna:
@stduhpf , would you please rebase this onto master? There are a few conflicts due to the VAE changes for Wan support.

stduhpf:
I haven't tested it, I know it builds fine, but I'm not sure if it actually works (I can't test it right now)

wbruna:
> I haven't tested it, I know it builds fine, but I'm not sure if it actually works (I can't test it right now)

No problem, I'll test it here.

stable-diffusion.cpp
```diff
@@ -1406,17 +1512,23 @@ class StableDiffusionGGML {
...
         }
-
         int64_t t0 = ggml_time_ms();
         if (!use_tiny_autoencoder) {
+            float tile_overlap = 0.5f;
+            int tile_size_x    = 32;
+            int tile_size_y    = 32;
+    
+            get_vae_tile_overlap(tile_overlap);
+            get_vae_tile_sizes(tile_size_x, tile_size_y, tile_overlap, x->ne[0] / 8, x->ne[1] / 8);
```
wbruna: This is on the decoding path, so it should be:

```c++
get_vae_tile_sizes(tile_size_x, tile_size_y, tile_overlap, x->ne[0], x->ne[1]);
```

wbruna (changes requested):
I've only noticed a single, very familiar issue 😀

stduhpf:
Encode is also broken, fixing it

wbruna (approved):
There is some unrelated whitespace changes, but apart from that it looks good to me. Working fine with image to image in SDXL, SD3 and Chroma, and image edit with Kontext.

leejet:
Thank you for your contribution. Please change the environment variables to command-line arguments. Once this change is completed, I will merge the PR.

wbruna:
@leejet , a quick question: the VAE tiling doesn't really change the loaded models, and its usefulness depends on the chosen image resolution (we even have a workaround on Koboldcpp to enable or disable it at inference time). Could we move it, together with the new tiling size parameters, to the sd_img_gen_params_t struct (possibly a new sd_tiling_params_t struct)? (I can do this change, either separately or on top of this PR)

leejet:
> @leejet , a quick question: the VAE tiling doesn't really change the loaded models, and its usefulness depends on the chosen image resolution (we even have a workaround on Koboldcpp to enable or disable it at inference time). Could we move it, together with the new tiling size parameters, to the sd_img_gen_params_t struct (possibly a new sd_tiling_params_t struct)? (I can do this change, either separately or on top of this PR)

This is what I expect.

stduhpf:
> I can do this change, either separately or on top of this PR

@wbruna do you want to do these changes yourself or should I go ahead and start working on it?

wbruna:
> @wbruna do you want to do these changes yourself or should I go ahead and start working on it?

I won't be able to until Friday, so feel free to do it. From what I looked, the hardest part will be choosing how to implement the tile size parameters: that env var syntax kind of does too much 😃

stable-diffusion.h
```diff
@@ -113,6 +113,16 @@ enum sd_log_level_t {
     SD_LOG_ERROR
 };
 
+typedef struct {
+    bool enabled;
+    int tile_size_x;
+    int tile_size_y;
+    float target_overlap;
+    bool relative;
```
wbruna: Perhaps this boolean could be dropped, and we apply the same logic as the command-line: if  `rel_size_ ` is bigger than 0.0, it overrides the `tile_size_` ?

stable-diffusion.cpp
```diff
@@ -1426,24 +1487,32 @@ class StableDiffusionGGML {
                                         C,
                                         x->ne[3]);
         }
-
         int64_t t0 = ggml_time_ms();
         if (!use_tiny_autoencoder) {
+            float tile_overlap = vae_tiling_params.target_overlap;
+            int tile_size_x    = vae_tiling_params.tile_size_x;
+            int tile_size_y    = vae_tiling_params.tile_size_y;
```
wbruna: A relative size of 0.0 applies a 4x4 tile (doesn't look good, but works); but a zeroed-out tile size crashes. Maybe an invalid value could just mean the default?
```c++
int tile_size_x = vae_tiling_params.tile_size_x < 4 ? 32 : vae_tiling_params.tile_size_x;
int tile_size_y = vae_tiling_params.tile_size_y < 4 ? 32 : vae_tiling_params.tile_size_y;
```
Then we could also simplify the default values for the param struct, by using zeroed-out fields.

wbruna (changes requested):
Looking good, and working fine on a few tests. The only issue I notice was a crash on an invalid tile size (more detailed comments below).

stable-diffusion.cpp
wbruna: Same issue with the tile_overlap above.

stduhpf:
Thanks for the review! I probably won't be able to work on these changes before tomorrow though.

leejet:
I think this PR can be merged now. Thank you all for your contributions.

PR #486, opened 2024-11-28
some engineering refactors

thxCode:
This PR is going to refactor/introduce the following changes.
- .gitignore: ignore cmake-* and .idea dirs.
- Align the ggml backend (s/CUBLAS/CUDA/g, s/HIPBLAS/HIP/g) and support CANN, MUSA.
- Extract public utils for reusing sampler_method/scheduler/rng.
- Support specifying main-gpu at the API level.
- ~Migrate sd_type_t to ggml_type~.


stable-diffusion.h
```diff
@@ -47,6 +58,22 @@ enum sample_method_t {
...
+    "euler",
+    "heun",
+    "dpm2",
+    "dpm++2s_a",
+    "dpm++2m",
+    "dpm++2mv2",
+    "ipndm",
+    "ipndm_v",
+    "lcm",
+};
```
Green-Sky, line 72: I don't think this should be in the header.

stable-diffusion.h
```diff
@@ -57,48 +84,17 @@ enum schedule_t {
...
-    SD_TYPE_TQ1_0    = 34,
-    SD_TYPE_TQ2_0    = 35,
-    SD_TYPE_COUNT,
+static const char* schedulers_argument_str[] = {
+    "default",
+    "discrete",
+    "karras",
+    "exponential",
+    "ays",
+    "gits",
```
Green-Sky, line 93: same

upscaler.cpp
```diff
@@ -14,29 +14,56 @@ struct UpscalerGGML {
...
     }
 
-    bool load_from_file(const std::string& esrgan_path) {
-#ifdef SD_USE_CUBLAS
+    bool load_from_file(const std::string& esrgan_path, int main_gpu) {
+#ifdef SD_USE_CUDA
+#ifdef SD_USE_HIP
+        LOG_DEBUG("Using HIP backend");
+#else
+#ifdef SD_USE_MUSA
```
Green-Sky, line 22: I think we should change this to `#elif defined(SD_USE_MUSA)`, remove some pp scoping and make it more readable.

Green-Sky (commented):
Generally agree with this stuff. :)

upscaler.cpp
thxCode: updated

leejet:
stable-diffusion.h is a standalone header file and should not depend on ggml.h. When sd.cpp is provided as a third-party library, user only need to use stable-diffusion.h and the corresponding library files.

thxCode:
> stable-diffusion.h is a standalone header file and should not depend on ggml.h. When sd.cpp is provided as a third-party library, user only need to use stable-diffusion.h and the corresponding library files.

rebase master and revert the change, PTAL.

Issue #431, opened 2024-10-05
Request : Q6 GGUF quantization


Danamir:
Is there an easy way to add Q6_K quantization (or eventually Q5_K) to the converter ? I find this particular quantization to be a good compromise between Q8_0 and Q4_K on my system. 

Cheers,

PR #489, opened 2024-11-28
cli: Remove type restrictions

stduhpf:
Allows to use any GGML type instead of only legacy quants and the small selection of k-quants. 

Some quant types are not compatible with some backends, but having more options is good I think.

(also, it solves https://github.com/leejet/stable-diffusion.cpp/issues/431)

### DIFF
--- a/examples/cli/main.cpp
+++ b/examples/cli/main.cpp
@@ -352,11 +352,14 @@ void parse_args(int argc, const char** argv, SDParams& params) {
             for (size_t i = 0; i < SD_TYPE_COUNT; i++) {
                 auto trait = ggml_get_type_traits((ggml_type)i);
                 std::string name(trait->type_name);
-                if (trait->type_size) {
+                if (name == "f32" || trait->to_float && trait->type_size) {
                     if (i)
                         valid_types += ", ";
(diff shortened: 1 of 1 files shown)


examples/cli/main.cpp
```diff
@@ -346,30 +346,25 @@ void parse_args(int argc, const char** argv, SDParams& params) {
...
-                        type.c_str());
+            std::string type        = argv[i];
+            bool found              = false;
+            std::string valid_types = "";
+            for (size_t i = 0; i < SD_TYPE_COUNT; i++) {
+                auto trait = ggml_get_type_traits((ggml_type)i);
+                std::string name(trait->type_name);
+                if (i)
+                    valid_types += ", ";
+                valid_types += name;
```
Green-Sky, line 357: the type enum is non contiguous, so we check for something here, maybe empty name?

stduhpf: Good catch

stduhpf:
I'm really imporessed by "iq4_nl" type. It runs faster than q4_k on my machine, and the quality seems on par with q5_1 (for sd3.5 large models at least).

leejet:

PR #490, opened 2024-11-29
Refactor: Flexible model architecture for dit models (Flux & SD3)

stduhpf:
(Built on top of https://github.com/leejet/stable-diffusion.cpp/pull/455)

### Motivations: 

I stumbled upon this: https://huggingface.co/TencentARC/flux-mini, and thought it would be nice to run it on sdcpp. 

The number of variants Flux and MMDiT (SD3.x) models supported is starting to get a bit overwhelming (3 each so far), and if people keep making these kinds of distillations or self merges, it would be impossible to support them all individually.

With this PR, the number of layers for each kind of block and the presence of some optional features is inferred from the tensor names in the model file when initializing the model runner, making it a lot more flexible.

### New models supported with these chages (examples):

- Flux Mini 3.2B (sucks at text): https://huggingface.co/TencentARC/flux-mini,
output

- Flux Heavy 17B: https://huggingface.co/city96/Flux.1-Heavy-17B
output

### DIFF
--- a/clip.hpp
+++ b/clip.hpp
@@ -545,9 +545,12 @@ protected:
     int64_t vocab_size;
     int64_t num_positions;
 
-    void init_params(struct ggml_context* ctx, ggml_type wtype) {
-        params["token_embedding.weight"]    = ggml_new_tensor_2d(ctx, wtype, embed_dim, vocab_size);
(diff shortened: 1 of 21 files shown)


leejet:

Issue #512, opened 2024-12-05
Build failing with Metal


grorge123:
When I update to the newest commit [9578fdc](https://github.com/leejet/stable-diffusion.cpp/commit/9578fdcc4632dc3de5565f28e2fb16b7c18f8d48), I encounter the build issue for metal. But it works for no metal or cuda.

- Compile command
```
cmake .. -DSD_METAL=ON
cmake --build . --config Release
```

- Cmake error message
```
stable-diffusion.cpp/./ggml_extend.hpp:1143:13: error: use of undeclared identifier 'ggml_backend_metal_set_n_cb'
            ggml_backend_metal_set_n_cb(backend, n_threads);

stable-diffusion.cpp/stable-diffusion.cpp:164:9: error: use of undeclared identifier 'ggml_backend_metal_log_set_callback'
        ggml_backend_metal_log_set_callback(ggml_log_callback_default, nullptr);

```

stduhpf:
I can confirm. I think it could be related to the fact that `ggml-metal.m` was moved from `ggml/src/` to `ggml/src/ggml-metal/`
(https://github.com/ggerganov/ggml/commit/813a12ce9edc1f908f1af69aacd54182005d9d4c, https://github.com/ggerganov/ggml/commit/a131a8200aac68500a44b5068705548db4bd2617)

stduhpf:
I can build by replacing references to `ggml_backend_metal_log_set_callback()` with `ggml_log_set()` and commenting out the call to `ggml_backend_metal_set_n_cb()`. I couldn't find what to replace it with.

PR #513, opened 2024-12-05
Fix metal build

stduhpf:
Some ggml API changes were causing metal builds to fail.

Fixes https://github.com/leejet/stable-diffusion.cpp/issues/512

### DIFF
--- a/stable-diffusion.cpp
+++ b/stable-diffusion.cpp
@@ -161,7 +161,7 @@ public:
 #endif
 #ifdef SD_USE_METAL
         LOG_DEBUG("Using Metal backend");
-        ggml_backend_metal_log_set_callback(ggml_log_callback_default, nullptr);
+        ggml_log_set(ggml_log_callback_default, nullptr);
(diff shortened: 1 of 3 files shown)


stable-diffusion.cpp
```diff
@@ -161,7 +161,7 @@ class StableDiffusionGGML {
 #endif
 #ifdef SD_USE_METAL
         LOG_DEBUG("Using Metal backend");
-        ggml_backend_metal_log_set_callback(ggml_log_callback_default, nullptr);
+        ggml_log_set(ggml_log_callback_default, nullptr);
```
Green-Sky, line 164: I don't think this should be inside of the `ifdef`.

stduhpf: you're probably right. I just replaced 1 to 1 to make sure I'm not breaking other backends as a side effect.

dreampuf:
Could we pass the PR? I have the same problem for MacOS build.

leejet:

PR #516, opened 2024-12-08
Use pretty-progress for tensor loading

stduhpf:
Wainting for tensors to load can sometime take a lot of time, especially when quantizing/converting the model. This gives the user some feedback to show that something is actually happening.


leejet:

Issue #483, opened 2024-11-26
stable diffusion cpp and hyper.sd sd3


michelgirault:
hi, I'm trying to run hypersd sd3 8 steps (see bellow) without any success
https://huggingface.co/ByteDance/Hyper-SD/resolve/main/Hyper-SD3-8steps-CFG-lora.safetensors

i was wondering if someone was successful? i understand you have to fuse the lora with the base model

thanks

stduhpf:
It's just a lora, in theory it should work like any other Lora, but here, it's using the _A/_B convention (instead of _up/_down), which is not supported yet. (see #370).
Maybe you can tray merging the lora into the model with another tool and then use the merge instead of the base model?

michelgirault:
thanks for the feedback, i will try like that

Issue #370, opened 2024-08-26
Some Flux loras can't be applied


stduhpf:
Some Flux loras are working, but others refuse to load properly (even though they work fine on comfyui). Around 50 % of the loras I've tested where failing.

There is no crash, but I get lots of `[WARN ] lora.hpp:176  - unused lora tensor transformer.xxx`, and it ends with `[WARN ] lora.hpp:186  - Only (0 / N) LoRA tensors have been applied`.

I'm not 100% sure what's going on, but a pattern I think I've noticed is that the Loras that can load all have 912 tensors, while all the ones who fail all have some different number of tensors?

<details>
<summary>Example of lora-related logs</summary>

```
[WARN ] stable-diffusion.cpp:617  - In quantized models when applying LoRA, the images have poor quality.
[INFO ] stable-diffusion.cpp:635  - Attempting to apply 1 LoRAs
[INFO ] model.cpp:789  - load ..\ComfyUI\models\loras\ana_de_armas_flux_lora_v1_000002000.safetensors using safetensors format
[DEBUG] model.cpp:857  - init from '..\ComfyUI\models\loras\ana_de_armas_flux_lora_v1_000002000.safetensors'
[INFO ] lora.hpp:33   - loading LoRA from '..\ComfyUI\models\loras\ana_de_armas_flux_lora_v1_000002000.safetensors'
[DEBUG] model.cpp:1526 - loading tensors from ..\ComfyUI\models\loras\ana_de_armas_flux_lora_v1_000002000.safetensors
[DEBUG] ggml_extend.hpp:1029 - lora params backend buffer size =  327.75 MB(RAM) (988 tensors)
[DEBUG] model.cpp:1526 - loading tensors from ..\ComfyUI\models\loras\ana_de_armas_flux_lora_v1_000002000.safetensors
[DEBUG] lora.hpp:69   - finished loaded lora
[WARN ] lora.hpp:176  - unused lora tensor transformer.single_transformer_blocks.0.attn.to_k.lora_A.weight
[WARN ] lora.hpp:176  - unused lora tensor transformer.single_transformer_blocks.0.attn.to_k.lora_B.weight
[...] (This goes on for all the tensors in the model)
[WARN ] lora.hpp:176  - unused lora tensor transformer.transformer_blocks.9.norm1_context.linear.lora_A.weight
[WARN ] lora.hpp:176  - unused lora tensor transformer.transformer_blocks.9.norm1_context.linear.lora_B.weight
[...]

PR #520, opened 2024-12-11
Support more LoRA models (Flux and SD3.x)

stduhpf:
Needs more testing.

Seems to work on CPU and Vulkan (unquantized) backends, on my machine at least.  

Please do some testing an let me know what works and what doesn't.

Should fix https://github.com/leejet/stable-diffusion.cpp/issues/370 and fix https://github.com/leejet/stable-diffusion.cpp/issues/483

### DIFF
--- a/lora.hpp
+++ b/lora.hpp
@@ -6,6 +6,41 @@
 #define LORA_GRAPH_SIZE 10240
 
 struct LoraModel : public GGMLRunner {
+    static enum lora_t {
+        REGULAR      = 0,
+        DIFFUSERS    = 1,
+        DIFFUSERS_2  = 2,
+        DIFFUSERS_3  = 3,
+        TRANSFORMERS = 4,
+        LORA_TYPE_COUNT
+    };
+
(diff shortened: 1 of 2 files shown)


Green-Sky:
Nice, good job @stduhpf , you have been coding like a machine :)

stduhpf:
> Nice, good job @stduhpf , you have been coding like a machine :)

Thanks, I'll probably get bored out of it eventually, but so far so good.

stduhpf:
I was testing SD3 support with https://huggingface.co/ByteDance/Hyper-SD/tree/main?show_file_info=Hyper-SD3-8steps-CFG-lora.safetensors. I can't wrap my head around what's going on with `pos_embed`, the sizes don't match at all the full model... 
Well it seems to work without it, but I would like to understand still.

stduhpf:
Ok I think Lora compatibility just got a lot better, but I still found some loras with very weird naming conventions that I don't quite understand so far.

stduhpf:
I think this broke previously working LoRAs

leejet:

PR #526, opened 2024-12-18
fix(metal): remove deprecated functions

pminev:
Some functions seems to be deprecated from ggml, but are still used and there are compiling errors.

In the CI, it works since the option `SD_METAL` is not passed to the cmake in [the command](https://github.com/leejet/stable-diffusion.cpp/blob/9578fdcc4632dc3de5565f28e2fb16b7c18f8d48/.github/workflows/build.yml#L110C1-L117C43), so the metal build is not tested.

### DIFF
--- a/ggml_extend.hpp
+++ b/ggml_extend.hpp
@@ -1138,11 +1138,6 @@ public:
             ggml_backend_cpu_set_n_threads(backend, n_threads);
         }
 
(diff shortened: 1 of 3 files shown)


leejet:
This should be fixed in https://github.com/leejet/stable-diffusion.cpp/pull/513 .

PR #511, opened 2024-12-04
 support Inpaint models

stduhpf:
Adds support for sd1.x inpaint, sd2.x inpaint, sdxl inpaint, and Flux Fill models, as well as "inpainting" (masked img2img) with normal diffusion models.

### How to use
- Download weights: (example: https://huggingface.co/webui/stable-diffusion-inpainting/blob/main/sd-v1-5-inpainting.safetensors)
- run in `img2img` mode with the `--mask` argument to specify the path to the masking image
- I recommend setting `strength` to `1`

#### Using Inpaint model: 

`sd.exe -M img2img --model ..\models\sd-v1-5-inpainting.safetensors  -p "a lovely dog" --cfg-scale 16 --sampling-method euler -t 24 --color --steps 30 --vae-tiling --strength 1 -i '.\inpaint-base.png' --mask '.\inpaint-mask.png'`

| Input image | mask | result |
| --- | --- | --- |
| image | image | output |

#### Using normal model (not as consistent): 

`sd.exe -M img2img --model ..\models\unet\sd3.5_medium-q6_k.gguf --clip_g ..\models\clip\clip_g.q8_0.gguf --clip_l ..\models\clip\clip_l.q8_0.gguf --t5xxl ..\models\clip\t5xxl_q4_k.gguf  -p 'a lovely dog sitting patiently on a park bench' --cfg-scale 4.5 --sampling-method euler --steps 30 -t 24 -W 512 -H 512 --seed 0 --color --vae-tiling --strength 1 -i '.\New folder\inpaint-base.png' --mask '.\New folder\inpaint-mask.png' `

| Input image | mask | result |
| --- | --- | --- |
| image | image | output |

### Caveats

- Implementing DDIM scheduler would probably help get better results, but this is already pretty good with the default one.
- Mask needs to be the same size as the image, resizing doesn't work.
- Since is based on img2img, the original image is always added to the random noise distribution, even at strength 1. This can be useful, for example to guide the generation by sketching what you want on the image, but it could also lead the model to regenerate what you want to remove from the image. If you want to avoid that, pre-fill the masked parts with 50% gray in the input image.
- Parts of the image outside the inpainting area can be slightly modified, especially with heavily quantized models, which tend to add artefacts all over the image.

Fixes #105

PR #537, opened 2024-12-30
Fix (non inpaint) sd2.x crash when calling is_using_v_parameterization_for_sd2()

stduhpf:
My mistake.

SD2 models are unusable right now (except the inpainting ones) because I messed it up with #511


Green-Sky (approved):

leejet:

PR #571, opened 2025-01-17
fix amd rocm build

null-define:
fix AMD ROCm build, change flag for ROCm from GGML_HIPBLAS to GGML_HIP according to https://github.com/ggerganov/ggml/blob/f196832b59684030515962711db4b19d848c8b68/CMakeLists.txt#L155


leejet:

PR #575, opened 2025-01-20
Lora: fix rank 1 loras

stduhpf:
Fixes #573
This assumes LoRA tensors will always have a pair number of dims, which I believe should always be the case for supported models?

Another idea I had to fix it was to check the rank using the penultimate dim of the lora_up tensor, if this implementation cause issues, I will try that instead.

PR #577, opened 2025-01-22
Lora: Partial LyCORIS support (tucker decomposition for LoCon + LoHa + LoKr)

stduhpf:
This should increase LoRA compatibility.

Technical details: https://github.com/KohakuBlueleaf/LyCORIS/blob/main/docs/Algo-Details.md#lora-for-convolution-with-tucker-decomposition

I'm not 100% sure I'm doing everything right but it looks like it's working fine... 

Implements the changes from https://github.com/leejet/stable-diffusion.cpp/pull/575 too


leejet:

PR #583, opened 2025-02-02
Support JPEG compression

vmobilis:
1) implement jpeg comment block to save parameters of generation;
2) save output with jpeg compression when extension is .jpg.

JPEG quality is simply set to 90, to not complicate command options.

And thanks for the program! 🙂

### DIFF
--- a/thirdparty/stb_image_write.h
+++ b/thirdparty/stb_image_write.h
@@ -1412,7 +1412,7 @@ static int stbiw__jpg_processDU(stbi__write_context *s, int *bitBuf, int *bitCnt
    return DU[0];
 }
 
-static int stbi_write_jpg_core(stbi__write_context *s, int width, int height, int comp, const void* data, int quality) {
+static int stbi_write_jpg_core(stbi__write_context *s, int width, int height, int comp, const void* data, int quality, const char* parameters) {
    // Constants that don't pollute global namespace
    static const unsigned char std_dc_luminance_nrcodes[] = {0,0,1,5,1,1,1,1,1,1,0,0,0,0,0,0,0};
    static const unsigned char std_dc_luminance_values[] = {0,1,2,3,4,5,6,7,8,9,10,11};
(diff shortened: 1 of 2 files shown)


thirdparty/stb_image_write.h
```diff
@@ -1521,6 +1521,18 @@ static int stbi_write_jpg_core(stbi__write_context *s, int width, int height, in
...
+
+      // comment block with parameters of generation
+      if(parameters != NULL) {
+         stbiw__putc(s, 0xFF /* comnent */ );
+         stbiw__putc(s, 0xFE /* marker  */ );
+         size_t param_length = std::min(2 + strlen("parameters") + 1 + strlen(parameters) + 1, (size_t) 0xFFFF);
+         stbiw__putc(s, param_length >> 8); // no need to mask, length < 65536
+         stbiw__putc(s, param_length & 0xFF);
+         s->func(s->context, (void*)"parameters", strlen("parameters") + 1); // std::string is zero-terminated
+         s->func(s->context, (void*)parameters, param_length - 2 - strlen("parameters") - 1);
```
idostyle, line 1533: Might be missing a zero-termination for parameters with length above 65522 (0xFFFF - strlen("parameters") - 1 - 1)

stduhpf:
Nice. I'm surprised stb_image doesn't support metadata by default.

thirdparty/stb_image_write.h
vmobilis: @idostyle, thanks for review.
JPEG comment block:
comment_marker { 0xFF, 0xFE /* comment marker \*/ }
comment_body {
&nbsp; uint16be comment_body.length(),
&nbsp; "comment" /* std::string terminates with 0 for compatibility with C strings \*/
}

Comment length can be 2..65535 bytes, according to https://www.w3.org/Graphics/JPEG/itu-t81.pdf, "Table B.8 – Comment segment parameter sizes and values".

         size_t param_length = std::min(2 + strlen("parameters") + 1 + strlen(parameters) + 1, (size_t) 0xFFFF);
// maker_length + len('parameters') + 0 + params.length() + <b>0</b>,
// truncated to 65535 bytes, so comment can be 65535 - 2 - 10 = 65523 bytes.

         s->func(s->context, (void*)parameters, param_length - 2 - strlen("parameters") - 1);
// 65535 - 2 /* marker*/  - 10 - 1 /* 'parameters' + 0 */

Second 0, terminating params, is not substracted.

vmobilis: @idostyle, I finally got it, you're right, despite that length of comment block is limited, it should always be terminated with 0, for safety. Fixed.

leejet:

PR #590, opened 2025-02-07
lora: unapply current loras properly

### LINKED ISSUE #410: apply_loras does not unapply old lora weights
This only matters when the same sd_ctx is used for multiple prompts - Loras that have been applied in a previous prompt but don't appear in the current prompt are not unapplied.

Steps to reproduce:
- Create an sd_ctx
- Generate an image using a prompt with a lora
- Generate an image using a prompt without the lora in the previous step
Expected results:
- The lora is no longer applied to the second image
Actual results:
- The lora is still applied to the second image

Why this happens:
curr_lora_state contains the loras that are currently applied and their weights. However, inside apply_loras, we only handle loras in the current prompt, and we don't consider any other loras that are in curr_lora_state and might need to be removed.

Potential fix:
In apply_loras(), loop over curr_lora_state and unapply any loras that do not appear in the lora_state (that is, add them to lora_state_diff with negative multiplier)

stduhpf:
Fixes https://github.com/leejet/stable-diffusion.cpp/issues/410

### DIFF
--- a/stable-diffusion.cpp
+++ b/stable-diffusion.cpp
@@ -673,19 +673,20 @@ public:
         for (auto& kv : lora_state) {
             const std::string& lora_name = kv.first;
             float multiplier             = kv.second;
-
-            if (curr_lora_state.find(lora_name) != curr_lora_state.end()) {
-                float curr_multiplier = curr_lora_state[lora_name];
-                float multiplier_diff = multiplier - curr_multiplier;
-                if (multiplier_diff != 0.f) {
-                    lora_state_diff[lora_name] = multiplier_diff;
-                }
-            } else {
-                lora_state_diff[lora_name] = multiplier;
-            }
+            lora_state_diff[lora_name] += multiplier;
         }
-
-        LOG_INFO("Attempting to apply %lu LoRAs", lora_state.size());
+        for (auto& kv : curr_lora_state) {
+            const std::string& lora_name = kv.first;
+            float curr_multiplier        = kv.second;
+            lora_state_diff[lora_name] -= curr_multiplier;
+        }
+        
+        size_t rm = lora_state_diff.size() - lora_state.size();
+        if (rm != 0) {
+            LOG_INFO("Attempting to apply %lu LoRAs (removing %lu applied LoRAs)", lora_state.size(), rm);
+        } else {
+            LOG_INFO("Attempting to apply %lu LoRAs", lora_state.size());
+        }
 
         for (auto& kv : lora_state_diff) {
             apply_lora(kv.first, kv.second);


fszontagh:
I just tested in GUI without checkpoint  reload:

1. run:
Prompt (without lora): `a cute cat glass statue`
txt2img_1739009207_473208252_512x768

2. run:
Prompt (with lora): `a cute cat glass statue <lora:glass_statue_v1:1>`
txt2img_1739009215_473208252_512x768

3. run:
Prompt (without lora): `a cute cat glass statue`
txt2img_1739009226_473208252_512x768

No other parameters are changed between the runs, only the prompt. 
The first and last output have some little difference.

The trigger word was `glass statue`, the lora is [from here](https://civitai.com/models/542525/glass-statue)

The parameters:
```
model: catCitronAnimeTreasure_v10
parser: stable-diffusion.cpp
cfg scale: 7.00
operations: txt2img
scheduler: default
app: StableDiffusionGUI 0.2.5 0d0fc16
size: 512x768
sampler: Euler a
backend: stable-diffusion.cpp (ver.: d46ed5e using cuda)
seed: 473208252
steps: 30
schedule type: Automatic
```

stduhpf:
Yes there is always going to be some very slight difference due to precision issues. With floats, `(a+b)-b` might not always be exactly the same as `a`, but I believe it should be close enough... I wouldn't have noticed the images are different if you didn't point it out. 

The only alternative would be to completely reload the model every time the LoRAs are changed, but this would increase the time per generation a bit.  (Maybe this could be an optional feature?)

Or have the LoRA weights loaded separately and added in the forward pass during inference but this will require a through refactor of every model, and it will end up making the models run slower with higher memory requirements when using LoRAs.

fszontagh:
> but I believe it should be close enough... 

Yeah, i think this is acceptable.

> The only alternative would be to completely reload the model every time the LoRAs are changed, but this would increase the time per generation a bit. (Maybe this could be an optional feature?)

I think this need to be implemented on the "user side", not in the lib itself. Or it can be faster if the lib itself can reload a model? Or maybe with something "differential" model reload? 


But i am now so happy to this fix as is.

> Or have the LoRA weights loaded separately and added in the forward pass during inference but this will require a through refactor of every model, and it will end up making the models run slower with higher memory requirements when using LoRAs.

Other implementations how handle this situation?

stduhpf:
> Other implementations how handle this situation?

 I'm not sure. I know this is how it's implemented in LoRA trainers (because they need gradients for the LoRA and for the base models separately), but for inference, I don't know for sure.

leejet:

PR #601, opened 2025-02-22
Fix embeddings with quantized models

### LINKED ISSUE #600: Vulkan embeddings and lora issue
I've just noticed that image generation using vulkan when an embedding is used generates a black image (only specifying --type f32 it to generate an image).

Also it's impossible to run a generation using a lora as I get this error:
```
stable-diffusion/stable-diffusion.cpp/ggml/src/ggml.c:3013: GGML_ASSERT(ggml_nelements(a) == ggml_nelements(b)) failed
ptrace: Operation not permitted.
No stack.
The program is not being run.
``` 

The issues seems to have originated from commit 7ce63e7.

stduhpf:
fixes #600

### DIFF
--- a/clip.hpp
+++ b/clip.hpp
@@ -546,7 +546,7 @@ protected:
     int64_t num_positions;
 
     void init_params(struct ggml_context* ctx, std::map<std::string, enum ggml_type>& tensor_types, const std::string prefix = "") {
-        enum ggml_type token_wtype    = (tensor_types.find(prefix + "token_embedding.weight") != tensor_types.end()) ? tensor_types[prefix + "token_embedding.weight"] : GGML_TYPE_F32;
+        enum ggml_type token_wtype    = GGML_TYPE_F32;  //(tensor_types.find(prefix + "token_embedding.weight") != tensor_types.end()) ? tensor_types[prefix + "token_embedding.weight"] : GGML_TYPE_F32;
         enum ggml_type position_wtype = GGML_TYPE_F32;  //(tensor_types.find(prefix + "position_embedding.weight") != tensor_types.end()) ? tensor_types[prefix + "position_embedding.weight"] : GGML_TYPE_F32;
 
         params["token_embedding.weight"]    = ggml_new_tensor_2d(ctx, token_wtype, embed_dim, vocab_size);


vmobilis:
@stduhpf, will that fix https://github.com/ggml-org/ggml/pull/1118 be useful after this patch, or it becomes needless?

vmobilis:
@stduhpf, on second thought, disabling quantization for conditioner has disadvantage: CLIP will consume more memory – depending on quantization, significantly more.
So maybe this should be at least optional.

But, if it won't complicate things much, @stduhpf, you brought good idea.
Considering that stable-diffusion.cpp is able to quantize parts of model on the fly, setting different quantization separately for conditioner and diffuser will produce 24² = 576 variants for **same prompt and sampler**.
For example:
<table>
<tr><td>clip f32, diffuser f32</td><td>clip f32, diffuser q4_1</td><td>clip q4_1, diffuser q4_1</td></tr>
<tr><td></td><td></td><td></td></tr>
</table>

stduhpf:
> @stduhpf, on second thought, disabling quantization for conditioner has disadvantage: CLIP will consume more memory – depending on quantization, significantly more. So maybe this should be at least optional.

Well with this PR It's not the whole clip model being forced to f32, "just" the `token_embedding.weight` tensor, wich is about 30% of clip-l total params, but only 10% of clip-g. 

There would be about a hundred megabytes of compute buffer to save with proper support for quantized concat(), compared to forcing f32. But then, the custom embeddings would get quantized too, which might cause significant quality loss... (maybe this could explain why your result is oversaturated?)

vmobilis:
@stduhpf, maybe it's because I'm building for ARM? If you want, here is the version I'm currently building from, will it make black images too?

I didn't sync SD_TYPE_COUNT and GGML_TYPE_COUNT, therefore weight type must always be specified. It's with your patch, and to temporarily enable quantization, there is an option "--keep-quantization".

Build options are in file "build_a55.sh", it makes clean build. But note the CPU options, they're tuned for ARM.
[sdcpp.tar.xz.zip](https://github.com/user-attachments/files/18951124/sdcpp.tar.xz.zip)

stduhpf:
I was getting normal-looking images with your version of the code, but that was just because of this line:  `bool keep_quant = false;` (clip.hpp:542). Setting it to `true`, I got black images again. 
I don't have a decent ARM system to test this on, and I'm not compiling with blas either. Maybe that could be the reason.

vmobilis:
@stduhpf, setting keep_quant to false turns on the patch. I see then, it doesn't work, maybe it's just my luck that it works for me... A friend even tried to run it on 3 Gb phone, it barely runs SD1.5 model with q4_K quantization, closing Google services due to lack of memory. 🙃

I tried without Blas, it works too (and I'm building on [Termux](https://github.com/termux/termux-app), any phone with Android 7+ and 6+ Gb of RAM will be sufficient).
I did not try with Vulkan because it crashes with non-f32 quantization ealier, when applying LoRA.
Anyway, without working concat() arbitrary quantization won't work.

Thanks for your work!

leejet:

Issue #606, opened 2025-02-28
img2img is broken


vmobilis:
Reference image:
<img src="https://github.com/user-attachments/assets/28b9251d-95f3-4ca8-93d6-4db855a057c3" height=100>
Command line:
`
./sd --model "dreamshaper_8.safetensors" -o "img2img.png" -p "green apple" --mode img2img -i "horse.jpg" --strength 0.4 --type f16
`
<table>
<tr><td>2eac844bbd30c0f0c4f0073f79941276a4bcecb8</td><td>cc92a6a1b3d819670711a573c7211731050d8c60</td><td>8f4ab9add3788b15e928ee8e3d7d0568423abf1f</td></tr>
<tr><td><img src="https://github.com/user-attachments/assets/0c3942e6-77a8-47a2-ac26-c1b26b2c876d"></td><td><img src="https://github.com/user-attachments/assets/f101134f-6905-4fdd-80f5-9ee4a11cbb25"></td><td><img src="https://github.com/user-attachments/assets/74ea12e4-fbb1-44e8-a4ce-ad43df7c5202"></td></tr>
</table>

Result is almost or identical to reference.
F32 and quantized types work same.
Regression is introduced in 8f4ab9add3788b15e928ee8e3d7d0568423abf1f commit.

Model version is Stable Diffusion **1.5**.
https://civitai.com/models/4384?modelVersionId=128713

stduhpf:
I can't reproduce this issue with commit 19d876ee300a055629926ff836489901f734f2b7

![Image](https://github.com/user-attachments/assets/502adfd8-a830-415e-95d8-57a8e66a936a)

8f4ab9add3788b15e928ee8e3d7d0568423abf1f also works for me:

![Image](https://github.com/user-attachments/assets/c892292f-c277-40eb-aa69-ce68868ada0e)

I don't know how this could be hapenning on your end, all I can think is that maybe it runs in inpaint mode with an empty mask? But this shouldn't be possible, if the mask is not specified, it's supposed to be initialized to fill the whole frame.

vmobilis:
@stduhpf, I found the reason: compiler filled mask with zeroes.
If I printed few values from [arr](https://github.com/leejet/stable-diffusion.cpp/blob/19d876ee300a055629926ff836489901f734f2b7/examples/cli/main.cpp#L938), the mask was filled with 0xFF, but otherwise the buffer seemed to be created without fill.
[...]

PR #610, opened 2025-02-28
Img2img mask can be filled with zeroes

vmobilis:
When using -O3 optimization, a mask can be filled with zeroes.
It looks (I didn't use debugger) like buffer with 0xFFs is ignored by some compilers and filled with 0x00. In my case, Clang does it.
This completely masks img2img result.

Also, this commit omits mask creation for txt2img and reduces memory usage.

Fixes https://github.com/leejet/stable-diffusion.cpp/issues/606.

### DIFF
--- a/examples/cli/main.cpp
+++ b/examples/cli/main.cpp
@@ -931,18 +931,6 @@ int main(int argc, const char* argv[]) {
         }
     }
 
-    if (params.mask_path != "") {
-        int c             = 0;
-        mask_image_buffer = stbi_load(params.mask_path.c_str(), &params.width, &params.height, &c, 1);
-    } else {
-        std::vector<uint8_t> arr(params.width * params.height, 255);
-        mask_image_buffer = arr.data();
-    }
-    sd_image_t mask_image = {(uint32_t)params.width,
-                             (uint32_t)params.height,
-                             1,
-                             mask_image_buffer};
-
     sd_image_t* results;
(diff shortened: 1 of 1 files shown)


vmobilis:
@leejet, @stduhpf, could you please help, to not make needless commits, do I need to include util.h into main.cpp if I want to use LOG_DEBUG()?

I built successfully locally, but all GitHub Windows builds seem to fail.

stduhpf:
It would work to include utils.h in the main example, but It would probably be best to either call `sd_log_cb()` directly, or to just use `printf()` instead of the logger.

leejet:
This is tricky and not very intuitive, or we could use a for loop to assign the value 255 to arr's elements.

leejet:
There is a dangling pointer issue here, arr memory is released after going out of scope, but it is still referenced by mask_image_buffer. I fixed this issue in master, please try it if it can solve your problem.

vmobilis:
@stduhpf, using sd_log_cb() directly would require copying log_printf() functionality, I'll try to use util.h.
But I haven't figured out how yet, linker of all Windows builds can not find log_printf() body, which should be compiled in libstable-diffusion.a.
Maybe I need to unmangle its name.

@leejet, will LOG_DEBUG() with notice to not remove it be more convenient?

vmobilis:
@leejet, yes, your commit fixed the issue, thank you very much!

PR #618, opened 2025-03-03
Clang complains about needless braces

vmobilis:
Because GITS_NOISE array is of
`const std::vector<std::vector<float>>*`

not
`std::vector< const std::vector<std::vector<float>>* >`

And thank you for the program.


Green-Sky (approved):

leejet:

PR #619, opened 2025-03-03
Silence warnings about enums mismatch

vmobilis:
Clang warns that ggml_type != sd_type_t.

And thank you for the program.


Green-Sky (approved):

leejet:
I believe the current master branch code should not have this issue. Anyway, thank you for your contribution.

PR #621, opened 2025-03-05
conditionner: suport sdxl embedddings

stduhpf:
SDXL embeddings can have custom vectors for both text encoders, not just clip_l.


JindrichSima:
**It works amazingly!**

I tested it on a pony model. It didn't work before the fix, now it works. I hope it will be added to the main branch. 
The apg branch was interesting too, I wonder if it will eventually be added too.

Thank you for your work.

leejet:

PR #632, opened 2025-03-20
fix: move processing implementation to a separate unit

pminev:


leejet:
Could you please explain why it needs to be moved to a separate unit? Is there any specific requirement for this?

PR #634, opened 2025-03-23
Loras can now be found in subdirectories of lora model folder.

ImKyra:
Made to not change the execution flow in case of the lora being directly in the model folder so it stays optimized if the correct path is provided.

/!\ Couldn't test unix version yet /!\


ImKyra:
Tested on kubuntu 22.04, works fine too

PR #643, opened 2025-03-31
harden for large files

Green-Sky:
`int`/`size_t` can be 32bit, which is bad for offsets into files larger than 2/4gig.


leejet:

PR #646, opened 2025-04-03
GGMLRunner: avoid multiple calls to get_graph

idostyle:
Not setting up the compute graph twice might result in a minor (/negligible?) performance improvement.

Previously it worked like the following in GGMLRunner#compute:
1. calls alloc_compute_buffer
1.1 calls reset_compute_ctx
1.2 calls get_graph
2. calls reset_compute_ctx
3. calls get_graph

Tested on cpu and vulkan locally.

### DIFF
--- a/ggml_extend.hpp
+++ b/ggml_extend.hpp
@@ -1105,12 +1105,10 @@ protected:
         }
     }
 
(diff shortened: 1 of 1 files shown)


wbruna:
A similar change was effectively done in 90e87bc846f17059771efb8aaa31e9ef0cab6f78 .

PR #651, opened 2025-04-07
feat: Add SYCL docker file

richiejp:
This updates GGML to avoid a linker error with oneapi and mkl_core as described here https://github.com/mudler/LocalAI/issues/5132

Also I added a Dockerfile for SYCL to test this. It compiles, but I'm not completely sure it works because I haven't gotten good results from image generation yet. However I don't have a good baseline to go from.

### DIFF
--- a/Dockerfile.sycl
+++ b/Dockerfile.sycl
@@ -1,0 +1,19 @@
+ARG SYCL_VERSION=2025.1.0-0
+
+FROM intel/oneapi-basekit:${SYCL_VERSION}-devel-ubuntu24.04 AS build
+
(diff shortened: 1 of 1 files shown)


richiejp:
OK, it seems to work
goblet

richiejp:
I backtracked one commit to avoid a compile error with CUDA and HIP: https://github.com/mudler/LocalAI/actions/runs/14430111877/job/40463894930

It's now on  a commit which is synced with whisper.cpp so I guess that is safer :shrug:

richiejp:
The sync commit was included in another PR, now this just adds the docker file.

leejet:

PR #652, opened 2025-04-09
[RFC] Support DistillT5

idostyle:
Add support for https://huggingface.co/LifuWang/DistillT5 https://github.com/LifuWang-66/DistillT5

Can be tested with https://huggingface.co/Eviation/DistillT5 e.g. https://huggingface.co/Eviation/DistillT5/blob/main/DistillT5-F32.safetensors which removes the additional wrapping.

- [x] Probably needs to be integrated, and at best automatically detected, alongside T5 XXL
- [ ] Check if there is a way to use the unmodified LifuWang DistillT5 model with the additional encoder wrapping on the tensor names
image

I would appreciate any ideas on how best to incorporate this in a less intrusive way.

### DIFF
--- a/t5.hpp
+++ b/t5.hpp
@@ -357,7 +357,7 @@ public:
 
         BuildTrie(&pieces);
     }
-    ~T5UniGramTokenizer(){};
+    ~T5UniGramTokenizer() {};
 
     std::string Normalize(const std::string& input) const {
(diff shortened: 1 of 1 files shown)


stduhpf:
Nice, just as I was thinking of adding it.

PR #654, opened 2025-04-11
model: support T5xxl with llama.cpp naming convention

stduhpf:
Most available T5 gguf quants are using llama.cpp naming convention.

For example: https://huggingface.co/city96/t5-v1_1-xxl-encoder-gguf/tree/main


leejet:

PR #680, opened 2025-05-18
Add OpenCL backend support

rmatif:
This PR adds support for the [OpenCL](https://github.com/ggml-org/llama.cpp/pull/10693) backend. Currently, this support is limited to Adreno GPUs. With the current ggml submodule, only Adreno 7xx series GPUs are supported. However, a previous PR that synced ggml will also enable support for Adreno 6xx series GPUs.

I'll update [Local-Diffusion](https://github.com/rmatif/Local-Diffusion) to use the latest ggml-opencl for those who don't want to bother with compiling and using adb


rmatif:
> Just out of curiosity, will this also work with desktop GPUs on x64 architecture in the future? Any possibility?

It may run on some Intel GPUs, but it's not optimized for them. For desktop GPUs, there are already plenty of well-optimized backends available

rmatif:
Should be fully working from https://github.com/ggml-org/ggml/commit/6bc3202d8bd8b98e9a506492656cdbca7a0fa3c1

leejet:

Issue #669, opened 2025-05-03
F64 dtype


valkarias:
Models with F64 type tensors do not work, specifically this SD 1.5 model: https://civitai.com/models/7371/rev-animated?modelVersionId=425083

![Image](https://github.com/user-attachments/assets/4c7d6c2b-1b1e-4833-a62c-2bdf7ff9904e)

-- even after adding a --type argument, it still fails 

![Image](https://github.com/user-attachments/assets/85ba658a-28c1-426a-880d-fcc4d113a532)

![Image](https://github.com/user-attachments/assets/80807beb-5ab1-4670-bc78-951e29aa8e36)

further research only led me to this other issue: https://github.com/leejet/stable-diffusion.cpp/issues/153
the offered program by the owner sadly did not work and i didnt want to fiddle with it any further

DanaNikifor:
[And where do you see fp64?](https://photos.app.goo.gl/BebAS4miCEn97JCq7)

DanaNikifor:
It seems like everyone is trying to reduce the model and keep the quality, but here it is))

valkarias:
> It seems like everyone is trying to reduce the model and keep the quality, but here it is))

have you looked into the related issue link, it would give more context. im not reducing the model size or manipulating it in any way. it errors simply if i use it with stable-diffusion.cpp showing errors of it not recognizing certain tensors

vmobilis:
@stduhpf, this can be fixed, maybe easily, since GGML / SD supports I64 type:
https://github.com/ggml-org/ggml/blob/ff9052988b76e137bcf92bb335733933ca196ac0/include/ggml.h#L379

The function
https://github.com/leejet/stable-diffusion.cpp/blob/10c6501bd05a697e014f1bee3a84e5664290c489/model.cpp#L961
should be appended with:
```
    } else if (dtype == "I64") {
        ttype = GGML_TYPE_I64;
```

I tested on SD 1.1 (https://huggingface.co/CompVis/stable-diffusion-v1-1) model, its [FP16 text encoder](https://huggingface.co/CompVis/stable-diffusion-v1-1/blob/main/text_encoder/model.fp16.safetensors) also has i64 tensor which breaks generation.

Also, its unet tensors have "unet.*" naming, therefore line https://github.com/leejet/stable-diffusion.cpp/blob/10c6501bd05a697e014f1bee3a84e5664290c489/model.cpp#L1618
should be changed to:
```
[...]

PR #681, opened 2025-05-18
Fix loading diffusers model (+support F64/I64 types)

### LINKED ISSUE #153: unsupported dtype 'F64'
Hi! Thanks for the great tool!

When I run sd.cpp with some not pruned SD models it drops the error.
```
[ERROR] model.cpp:773  - unsupported dtype 'F64'
```
It seems to me a simple thing to fix, either by supporting double float internally or converting during load into single float.
I want to hear you thoughts on this situation and how to solve it.

With best regards, AG

stduhpf:
Loading diffusers models wasn't working, this fixes it for me,. I also added support for SDXL diffusers model (the second text encoder wasn't being handled).

During testing, I came across models with 64bit types, so I added support for those too since they are supported by GGML. (Fixes https://github.com/leejet/stable-diffusion.cpp/issues/153, https://github.com/leejet/stable-diffusion.cpp/issues/669)

### DIFF
--- a/model.cpp
+++ b/model.cpp
@@ -1107,7 +1107,7 @@ bool ModelLoader::init_from_safetensors_file(const std::string& file_path, const
         }
 
         std::string new_name = prefix + name;
-        new_name             = convert_tensor_name(new_name);
+        new_name             = prefix == "unet." ? convert_tensor_name(new_name) : new_name;
 
         TensorStorage tensor_storage(new_name, type, ne, n_dims, file_index, ST_HEADER_SIZE_LEN + header_size_ + begin);
         tensor_storage.reverse_ne();
@@ -1568,7 +1568,7 @@ SDVersion ModelLoader::get_sd_version() {
                     }
                 }
             }
-            if (tensor_storage.name.find("conditioner.embedders.1") != std::string::npos || tensor_storage.name.find("cond_stage_model.1") != std::string::npos) {
+            if (tensor_storage.name.find("conditioner.embedders.1") != std::string::npos || tensor_storage.name.find("cond_stage_model.1") != std::string::npos || tensor_storage.name.find("te.1") != std::string::npos) {
                 has_multiple_encoders = true;
                 if (is_unet) {
                     is_xl = true;
--- a/model.cpp
+++ b/model.cpp
@@ -1106,10 +1106,8 @@ bool ModelLoader::init_from_safetensors_file(const std::string& file_path, const
             n_dims = 1;
         }
 
-        std::string new_name = prefix + name;
-        new_name             = prefix == "unet." ? convert_tensor_name(new_name) : new_name;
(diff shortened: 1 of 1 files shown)


Green-Sky (approved):

model.cpp
```diff
@@ -1071,7 +1106,10 @@ bool ModelLoader::init_from_safetensors_file(const std::string& file_path, const
             n_dims = 1;
         }
 
-        TensorStorage tensor_storage(prefix + name, type, ne, n_dims, file_index, ST_HEADER_SIZE_LEN + header_size_ + begin);
+        std::string new_name = prefix + name;
+        new_name             = convert_tensor_name(new_name);
+
+        TensorStorage tensor_storage(new_name, type, ne, n_dims, file_index, ST_HEADER_SIZE_LEN + header_size_ + begin);
```
stduhpf: This is breaking some LoRAs.

leejet:

wbruna:
@stduhpf , I'm getting an 'f64 unsupported' error trying the model mentioned on #153 ( https://civitai.com/models/7371/rev-animated?modelVersionId=425083 , fp32 file):
```
./sd --model ./revAnimated_v2Rebirth.safetensors -p flower 
ggml_vulkan: 0 = AMD Radeon RX 7600 XT (RADV NAVI33) (radv) | uma: 0 | fp16: 1 | warp size: 64 | shared memory: 65536 | int dot: 1 | matrix cores: KHR_coopmat
[INFO ] stable-diffusion.cpp:210  - loading model from './revAnimated_v2Rebirth.safetensors'
[INFO ] model.cpp:998  - load ./revAnimated_v2Rebirth.safetensors using safetensors format
[INFO ] stable-diffusion.cpp:259  - Version: SD 1.x 
[INFO ] stable-diffusion.cpp:292  - Weight type:                 f32
[INFO ] stable-diffusion.cpp:293  - Conditioner weight type:     f32
[INFO ] stable-diffusion.cpp:294  - Diffusion model weight type: f32
[INFO ] stable-diffusion.cpp:295  - VAE weight type:             f16
  |====================>                             | 459/1130 - 0.00it/sterminate called after throwing an instance of 'std::runtime_error'
  what():  type f64 unsupported for integer quantization: no dequantization available
Abortado (imagem do núcleo gravada)
```

This is on master-dafc32d .

stduhpf:
Yeah, F64 is still not properly supported. I didn't realize there isn't a built in "dequantization" (or rather quantization in that case) function for F64 to F32 in GGML, so I just implemented it with https://github.com/leejet/stable-diffusion.cpp/pull/726. This fix the crash during loading time, but I can't get inference to run with either Vulkan or CPU backends when using the models with F64 weights. Forcing it to use F32 works with the new new changes I made ( `--type f32`).

PR #683, opened 2025-05-19
Add CosXL support

### LINKED ISSUE #630: Support for CosXL/CosXL Edit
Would be great if someone could implement **cosxl** and **cosxl edit**!  

At first glance, this would require implementing the **Cosine-Continuous EDM VPred schedule**, currently, the master branch produces garbage images with the existing scheduler when trying to inference with cosxl

**cosxl edit** might require additional work, but I believe it's worth it since it enables **zero-shot image editing while maintaining consistency**.  

**Example Image to motivate**

Example Image to motivate

stduhpf:
Fixes https://github.com/leejet/stable-diffusion.cpp/issues/630

| ComfyUI output | This PR's output (same settings) |
| --- | --- |
| image | output |

### DIFF
--- a/denoiser.hpp
+++ b/denoiser.hpp
@@ -346,6 +346,38 @@ struct CompVisVDenoiser : public CompVisDenoiser {
     }
 };
 
+struct CompVisEDMVDenoiser : public CompVisVDenoiser {
+    float sigmas[TIMESTEPS];
+    float min_sigma                         = 0.002;
+    float max_sigma                         = 120.0;
+    float sigma_data                        = 1;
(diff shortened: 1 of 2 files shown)


stduhpf:
Hmm it looks like the `--schedule` argument has no effect, maybe this can be related to those artifacts...

stduhpf:
Looks like adding a simple negative prompt helps a lot to mitigate the overasturation issue, but it still doesn't look quite right:

output

(ComfyUI doesn't need a negative prompt to generate good images with cosXL, so something fishy is going on here)

stduhpf (commented):
Setting `force_zero_embeddings` to false seems to also work for some reason.

leejet:

PR #687, opened 2025-05-24
fix: allow changing clip_skip back to its default value

wbruna:
CLIPTextModel ignored setting clip_skip back to -1, instead retaining whatever value was previously defined.

This is not relevant to command-line sd, since it's currently not possible to change clip_skip between generations, but could affect any frontend that reuses model instances for multiple images.


LostRuins:
I don't think this fix is correct. It's breaking some models now.

PR #692, opened 2025-05-26
Support for Flux Controls + Flex.2

stduhpf:
###  Flux Controls:

https://huggingface.co/black-forest-labs/FLUX.1-Canny-dev
https://huggingface.co/black-forest-labs/FLUX.1-Depth-dev

Flux models with built-in controlnet-like conditionning suport  (depth or canny depending on the model)

### Flex.2 (preview)

https://huggingface.co/ostris/Flex.2-preview

Pruned and de-distilled opensource Flux model with support for inpainting and controlnet-like conditionning (line, pose and depth all supported by a single model).

### Usage

To enable control image conditionning, use the `--control-image` argument (same as when using a control net). You can also use `--control-strength` to change how much effect the control image should have.


iwr-redmond:
Will this PR make it easier to support SDXL control LoRAs in the future?

https://huggingface.co/stabilityai/control-lora

leejet:

Issue #686, opened 2025-05-23
Allow @stduhpf and @Green-Sky to approve PRs


iwr-redmond:
@stduhpf is a prolific coder with many PRs added to this project. It would be helpful to provide @stduhpf with write access to this repository so that development of stable-diffusion.cpp can continue when @leejet is unavailable.

Green-Sky:
Yes. llama.cpp does that too. I claim that was one of the reasons why it took off and scaled so well in the early days.

ggerganov:
In the settings of the Github repo, you can create a rule that requires a number of approvals to merge a PR. For example:

![Image](https://github.com/user-attachments/assets/f470154d-3a0e-421e-9d57-ab8010a90234)

This is what we use in the llama.cpp repo and it works OK IMO.

iwr-redmond:
Thanks for your feedback, everyone. I've updated the issue title to reflect the discussion.

iwr-redmond:
Copying the comments from #696 for easy reference:

@kgigitdev:

> I'm sure that nobody would take offence if a temporary (friendly, and prominently attributed) fork were created to contain suitably-approved merged and conflict-resolved branches from all the developers who have pending PRs, as well as lots of useful work that is currently being duplicated across lots of forks.

@Green-Sky:

> If we ended up doing this, I would ask @ggerganov to host that project at the ggml org. But let's wait a while longer and see what @leejet ends up doing.

SkutteOleg:
Hopefully optional PRs remain optional, and the library does not get bloated. Though current open PRs seem to be alright.

iwr-redmond:
It's been over three months since @leejet last contributed according to Github's tracker, a much longer period than previous hiatuses. Is it time to migrate to the [bleeding edge code](https://github.com/stduhpf/stable-diffusion.cpp) maintained by @stduhpf?

Green-Sky:
> It's been over three months since [@leejet](https://github.com/leejet) last contributed according to Github's tracker, a much longer period than previous hiatuses. Is it time to migrate to the [bleeding edge code](https://github.com/stduhpf/stable-diffusion.cpp) maintained by [@stduhpf](https://github.com/stduhpf)?

I think that is already naturally happening.
But don't treat https://github.com/stduhpf/stable-diffusion.cpp/tree/bleedingedge as stable 😄

iwr-redmond:
> I think that is already naturally happening.

[...]

PR #675, opened 2025-05-09
Add Timestep shift, SGM Uniform and Simple scheduler and support for NitroFusion

rmatif:
This PR adds support for the timestep-shift technique required for inference with [NitroFusion](https://arxiv.org/abs/2412.02030) models and [Diff-Instruct*](https://arxiv.org/abs/2410.20898) and [other one-step models](https://huggingface.co/XDG-XHS). It also adds support for two schedulers: SGM Uniform and Simple, because the existing scheduler, for a mysterious reason, fails at step 2 calculation and produces an output similar to step 1.

NitroFusion is one of the best models for single-step inference, making it useful for inference on compute-constrained devices like mobile phones or CPUs.

Example command:

`./bin/sd -m nitrosd-realism_f16.gguf -v -p "cute cat" --cfg-scale 1 --steps 1 --timestep-shift 250 -H 1024 -W 1024 --seed 2024 --schedule sgm_uniform`

| Step    | NitroSD-Realism (Timestep-shift 250) | NitroSD-Vibrant (Timestep-shift 500) | Diff-Instruct* (Timestep-shift 400) |
|---------|--------------------------------------|--------------------------------------|--------------------------------------|
| Step 1  | 1step | 1step-vibrant | 1step-diff-instructstar |
| Step 2  | 2step | 2step-vibrant | Not well supported |
| Step 3  | 3step | 3step-vibrant | Not well supported |
| Step 4  | 4step | 4step-vibrant | Not well supported |

The recommended timestep-shift values by the authors are 250 for NitroSD-Realism and 500 for NitroSD-Vibrant and 400 for Diff-Instruct*.

I created GGUF versions of NitroFusion that already include the [fixed SDXL VAE](https://huggingface.co/madebyollin/sdxl-vae-fp16-fix), available for download [here](https://huggingface.co/mrfatso/NitroFusion-GGUF).

The authors mentioned it's possible to extract LoRA weights from these models and apply them to other checkpoints. I’ll try to do that in the future.

EDIT : Just add [Diff-Instruct* GGUF](https://huggingface.co/mrfatso/Diff-InstructStar-GGUF)
References: 

Timestep-shift implementation: [node.py](https://github.com/ChenDarYen/ComfyUI-TimestepShiftModel/blob/main/node.py)
[...]

PR #696, opened 2025-05-28
Chroma support (pruned Flux model)

### LINKED ISSUE #690: Support other fine tuned flux models and gguf versions
Please add support for fine tuned flux models I'm trying to use chromav31unlocked.gguf but it's not working

stduhpf:
https://huggingface.co/lodestones/Chroma

Chroma is a Flux model with modulation layers pruned off, which makes it fit in a lower memory footprint. Unlike Flux, it doesn't use Clip-L, only t5-xxl.

### Usage
 
```
.\build\bin\Release\sd.exe --diffusion-model .models\diffusion-models\chroma-unlocked-v33-Q5_0.gguf --t5xxl .\models\t5\t5xxl_q4_k.gguf --vae .\models\vae\flux\ae.f16.gguf -p "A Cute cat holding a sign that says: \`"Stable diffusion.cpp Now supports Chroma!\`"" --cfg-scale 4 --sampling-method euler --vae-tiling -W 1024 -H 1024
```
output

### Advanced usage

The following environment variables can be set to change the behavior:
- SD_CHROMA_USE_T5_MASK (default to "OFF")
- SD_CHROMA_USE_DIT_MASK (default to "ON")
- SD_CHROMA_MASK_PAD_OVERRIDE (default to 1)
- SD_CHROMA_ENABLE_GUIDANCE (default to "OFF", setting it to "ON" without  `--guidance 0` arg seems to break inference)

(closes https://github.com/leejet/stable-diffusion.cpp/issues/690)


stduhpf:
Huh, It seems to kind of work (slowly) on CPU backend... Like the image is not good, but at least it kind of looks like the prompt if you squint your eyes. Maybe there's an issue with the Vulkan implementation of GGML, or there's somthing I'm doing that breaks when using GPU?

 Can someone test with Cuda?

 prompt: 
 ```
'Extreme close-up photograph of a single tiger eye, direct frontal view. The iris is very detailed and the pupil resembling a dark void. The word "Chroma" is across the lower portion of the image in large white stylized letters, with brush strokes 
resembling those made with Japanese calligraphy. Each strand of the thick fur is highly detailed and distinguishable. Natural lighting to capture authentic eye shine and depth.'
```
(`--cfg-scale 1 --sampling-method euler --vae-tiling --steps 16 --guidance 0`)
output
(same settings produce a black image on Vulkan)

**Comment by stduhpf** [comment]
Ok, running the Vulkan build with preview on, I can say there's something very wrong going on. Sometimes (very rarely, I got this like twice in a hundred tests) the output looks correct after the first step, then it turns to noise, then a full black image (Probably NaN/inf). About half of the time it looks like noise from the first step already and then turns black after a few steps. The rest of the time it starts off with a black image and stays like that. It's extremely inconsistent.

**Comment by Green-Sky** [comment]
Ran it on cuda.
output
(different prompt, but I dont think thats important :smile: )

**Comment by rmatif** [comment]
@stduhpf Ran it on CUDA too and I got this, it's inconsistent too, I ran it 10 times and I got same thing each time 

output_chroma

**Comment by Green-Sky** [comment]
It slightly varies.

trun2
trun1
test_run

**Comment by Green-Sky** [comment]
I tried compiling it with ubsan and asan, but
```
[DEBUG] stable-diffusion.cpp:165  - Using CUDA backend
ggml_cuda_init: failed to initialize CUDA: out of memory
ggml_backend_cuda_init: invalid device 0
[DEBUG] stable-diffusion.cpp:188  - Using CPU backend
```
and thats the first thing that happens. looks like an upstream issue. also we should update ggml <.<

**Comment by Green-Sky** [comment]
Output with ubsan (no asan)

trun3

since this looks like the vulkan output, I am guessing the issue is one and the same.

**Comment by stduhpf** [comment]
Ok it's very important to keep the distilled guidance scale to 0. For some reason the model still accepts it as an input, but it completely breaks apart if it's not zero (I double checked it with ComfyUI, it's not an issue with my code). Maybe I should just force it to zero for chroma to keep things simple?

**Comment by Green-Sky** [comment]
> Ok it's very important to keep the distilled guidance scale to 0. For some reason the model still accepts it as an input, but it completely breaks apart if it's not zero (I double checked it with ComfyUI, it's not an issue with my code). Maybe I should just force it to zero for chroma to keep things simple?

Go for it. Down the line we should put recommended/forced values into the gguf file.

**Comment by Green-Sky** [comment]
chroma1

The q4_k is hurting it somewhat, as expected.

`--sampling-method euler --steps 16 --guidance 0 --cfg-scale 4 --diffusion-fa -W 1024 -H 1024`

```
[DEBUG] ggml_extend.hpp:1174 - flux params backend buffer size =  4824.80 MB(VRAM) (643 tensors)
[DEBUG] ggml_extend.hpp:1126 - flux compute buffer size: 854.00 MB(VRAM)
[INFO ] stable-diffusion.cpp:1628 - txt2img completed in 208.26s
```

**Comment by stduhpf** [comment]
> The q4_k is hurting it somewhat, as expected.

Keeping the `distilled_guidance_layer`weights to high precision seems to help a lot. (for example: https://huggingface.co/silveroxides/Chroma-GGUF, it's using BF16, so comaptibility with gguf backends might not work, I had to update GGUF to get it to wor on vulkan)

**Comment by Green-Sky** [comment]
> Which begs the question: should we condsider making the convert tool "smarter", like it is in llama.cpp, with different quant types depending on the role of each tensor?

We should.

I find this model with current sd.cpp quantization incredibly hard to prompt too, but that is probably the token padding / masking.

chroma_apls-q5_k

This is supposed to be q5_k quality, normal flux looks way better, even flux light looks better.

**Comment by kgigitdev** [comment]
Hi @stduhpf and @Green-Sky ,

Apologies for the question, but is this branch intended to work solely with vanilla master and master's version of ggml? I've been maintaining a little script for personal use that merges specific branches from here and there to get a more up-to-date build, and no matter what I do I can't get this branch to work, either on its own or with other branches.

Here's the extract from my script that shows the current state of what I have generally been merging in (with a few of them commented out, as you can see, but I've included them anyway since they do work in some other combinations):

```sh
BRANCHES="zhouwg/sync_with_latest_ggml"
BRANCHES="${BRANCHES} wbruna/fix_sdxl_lora"
BRANCHES="${BRANCHES} stduhpf/sdxl-embd"
BRANCHES="${BRANCHES} stduhpf/tiled-vae-encode"
BRANCHES="${BRANCHES} stduhpf/imatrix"
BRANCHES="${BRANCHES} stduhpf/lcppT5"
BRANCHES="${BRANCHES} stduhpf/unchained"
BRANCHES="${BRANCHES} stduhpf/dt"
# BRANCHES="${BRANCHES} stduhpf/diffusers"
BRANCHES="${BRANCHES} stduhpf/override-te"
BRANCHES="${BRANCHES} stduhpf/concat-controls"
# BRANCHES="${BRANCHES} stduhpf/ip2p"
BRANCHES="${BRANCHES} ImKyra/master"
# BRANCHES="${BRANCHES} Green-Sky/large_file_hardening"
# BRANCHES="${BRANCHES} rmatif/sigmas"
```

Now, I already suspected that this branch would probably not work directly in my script, since you said, "I had to update GGML to get bf16 to work on vulkan", and that probably conflicts with `zhouwg/sync_with_latest_ggml`. But in general, it has been harder and harder to get a working build with all the latest features and fixes in place.

Now, Chroma being a Flux derivative, I should also mention that for a while now I've not been able to get Flux to work either; the problems are generally of the form:

A gajillion of these:

```sh
[ERROR] model.cpp:1938 - tensor 'first_stage_model.decoder.conv_in.bias' not in model file
```

or a gajillion of these (with the `--diffusion-model` argument):

```sh
[INFO ] model.cpp:1897 - unknown tensor 'model.diffusion_model.model.diffusion_model.double_blocks.0.img_attn.norm.key_norm.scale | f8_e4m3 | 1 [128, 1, 1, 1, 1]' in model file
```

(the doubled prefix probably indicating that something has automatically prefixed `model.diffusion_model.` where one already existed).

followed by:

```sh
[ERROR] stable-diffusion.cpp:441  - load tensors from model loader failed
```

From looking at `ModelLoader::get_sd_version()`, it looks like there's a ton of heuristics to determine the model type based on the tensor names, but that's all probably doomed to fail anyway if the tensor loading has previously failed. Or, I'm missing a branch somewhere that adds new entries to `enum SDVersion`.

And that's the point where I give up, since I don't know enough about the expected naming of the tensors.

Which also brings us to the elephant in the room that you're all too polite to talk about :-) : vis-a-vis #686 , given that I'm probably not the only one with the above problems, I'm sure that nobody would take offence if a temporary (friendly, and prominently attributed) fork were created to contain suitably-approved merged and conflict-resolved branches from all the developers who have pending PRs, as well as lots of useful work that is currently being duplicated across lots of forks.

Green-Sky:
@kgigitdev I feel you. For my use case I depend on the webserver api.
Can you open a separate issue for this? It seems to be gernally flux, like you said.

> Now, I already suspected that this branch would probably not work directly in my script, since you said, "I had to update GGML to get bf16 to work on vulkan"

This does not seem to be in this pr.


> Which also brings us to the elephant in the room that you're all too polite to talk about :-) : vis-a-vis https://github.com/leejet/stable-diffusion.cpp/issues/686 , given that I'm probably not the only one with the above problems, I'm sure that nobody would take offence if a temporary (friendly, and prominently attributed) fork were created to contain suitably-approved merged and conflict-resolved branches from all the developers who have pending PRs, as well as lots of useful work that is currently being duplicated across lots of forks.

If we ended up doing this, I would ask @ggerganov to host that project at the ggml org.
But let's wait a while longer and see what @leejet ends up doing.

kgigitdev:
Hi @Green-Sky ,

Thanks for your swift answer.

> If we ended up doing this, I would ask @ggerganov to host that project at the ggml org.

Gosh, yes, I hadn't even thought of that option. I think I had subconsciously assumed that stable-diffusion.cpp was only ever on the periphery of the llama.cpp people, because llama == serious work whereas stable diffusion == frippery and frivolity.

stduhpf:
> A gajillion of these:
> 
> ```shell
> [ERROR] model.cpp:1938 - tensor 'first_stage_model.decoder.conv_in.bias' not in model file
> ```

That probably means the VAE is missing, or that the tensors from the VAE can't be found because their names are not the ones expected.

> or a gajillion of these (with the `--diffusion-model` argument):
> 
> ```shell
> [INFO ] model.cpp:1897 - unknown tensor 'model.diffusion_model.model.diffusion_model.double_blocks.0.img_attn.norm.key_norm.scale | f8_e4m3 | 1 [128, 1, 1, 1, 1]' in model file
> ```
> 
> (the doubled prefix probably indicating that something has automatically prefixed `model.diffusion_model.` where one already existed).

Yes, the doubled prefix means the tensor names are already prefixed in the model file you're using, which means you should use `--model` instead of `--diffusion-model`. This often happens with versions of Flux with built-in vae, in which case that would also fix the other issue.

stduhpf:
| With attention masking | Without | 
| --- | --- |
| output copy 6 | output copy 7 |

Somehow the results I get with the attention masking look a bit worse than what I had before without it. Maybe once I implement the mask modification to attend some padding tokens, it will fix itself? For now if you just want to generate better pictures, use an earlier commit. 
The compute buffer size is also significantly increased so I can no longer generate 1024x1024 with Vulkan (It reaches the allocation limit)

Edit: I fixed the compute buffer size issue, and after trying different prompts I'm not sure which is the best actually. Maybe using masking is not so bad after all.

wbruna:
> Which begs the question: should we consider making the convert tool "smarter", like it is in llama.cpp, with different quant types depending on the role of each tensor?

FWIW, I've been working on an option for sd.cpp to choose the quant by tensor pattern, a la llama.cpp's overridetensors: e128cfa4eb64 . The conversion itself already works; could be useful for testing.

stduhpf:
Comparison of masking off, vs on with different unmasked padding levels:


| | t5 masking | No t5 masking |
| - | - | - |
| No Padding | t5-M0 | not5-M0 |
| 1 Pad (default) | t5-M1 | not5-M1 |
| 2 Pad | t5-M2 | not5-M2 |
| 4 Pad | t5-M4 | not5-M4 |
| 8 Pad | t5-M8 | not5-M8 |
| No Dit Masking | t5-MI | not5-MI |

| | t5 masking | No t5 masking |
| - | - | - |
| No Padding | t5-M0 | not5-M0 |
| 1 Pad (default) | t5-M1 | not5-M1 |
| 2 Pad | t5-M2 | not5-M2 |
| 4 Pad | t5-M4 | not5-M4 |
| 8 Pad | t5-M8 | not5-M8 |
| No Dit Masking | t5-MI | not5-MI |

| | t5 masking | No t5 masking |
| - | - | - |
| No Padding | t5-M0 | not5-M0 |
| 1 Pad (default) | t5-M1 | not5-M1 |
| 2 Pad | t5-M2 | not5-M2 |
| 4 Pad | t5-M4 | not5-M4 | 
| 8 Pad | t5-M8 | not5-M8 |
| No Dit Masking | t5-MI | not5-MI |

Flux compute buffer size (512x768): 1197.13 MB (DiT Masking on) vs 813.13 MB (DiT masking Disabled) ~1.5x increase
t5 compute buffer size: 84.25 MB (T5 Masking on) vs 68.25 MB (T5 Masking off) 

Sampling speed (512x768):  13.48s/it vs 13.70s/it

Not sure enabling masking is worth it. Masking pad tokens for T5, but not for the DiT seems to break things (though I kinda like how trippy it looks sometimes). 

I don't know why they didn't use the same unmasked padding for t5 and the DiT in the original implementation, this feels odd. I might try to experiment around with including the same pad tokens for both models.

LostRuins:
They all look... more or less just as good lol. I can't really tell tbh.

stduhpf:
322fad2df2f630215b6d338cadcc5618110ebade fixes the huge compute buffer size difference. 

Maybe we should use inplace ops whenever it's possible, there is quite a bit of memory usage to spare by doing so (in another PR?).

Green-Sky:
> [322fad2](https://github.com/leejet/stable-diffusion.cpp/commit/322fad2df2f630215b6d338cadcc5618110ebade) fixes the huge compute buffer size difference.

Nice, I did not get around testing this, but looks like `--diffusion-fa` would have circumvented the issue (maybe, depends where).

> Maybe we should use inplace ops whenever it's possible, there is quite a bit of memory usage to spare by doing so (in another PR?).

Yes. Let's differentiate from stuff like comfy.

Green-Sky:
I just tried it and it asserts:
```
[DEBUG] t5.hpp:398  - token length: 512
ggml_gallocr_reserve_n: reallocating CPU buffer from size 0.00 MiB to 84.25 MiB
[DEBUG] ggml_extend.hpp:1126 - t5 compute buffer size: 84.25 MB(RAM)
[DEBUG] conditioner.hpp:1353 - computing condition graph completed, taking 9259 ms
ggml_gallocr_reserve_n: reallocating CPU buffer from size 0.00 MiB to 84.25 MiB
[DEBUG] ggml_extend.hpp:1126 - t5 compute buffer size: 84.25 MB(RAM)
sd: /build/gblmj4jz1z0xx3ls4s5xv1wd7s79hdj7-source/ggml/src/ggml-cpu/ggml-cpu.c:10104: ggml_compute_forward_soft_max_f32: Assertion `sum > 0.0' failed.
sd: /build/gblmj4jz1z0xx3ls4s5xv1wd7s79hdj7-source/ggml/src/ggml-cpu/ggml-cpu.c:10104: ggml_compute_forward_soft_max_f32: Assertion `sum > 0.0' failed.
sd: /build/gblmj4jz1z0xx3ls4s5xv1wd7s79hdj7-source/ggml/src/ggml-cpu/ggml-cpu.c:10104: ggml_compute_forward_soft_max_f32: Assertion `sum > 0.0' failed.
```

edit: Im at 322fad2df2f630215b6d338cadcc5618110ebade , cuda

stduhpf:
> I just tried it and it asserts:
> 
> ```
> [DEBUG] t5.hpp:398  - token length: 512
> ggml_gallocr_reserve_n: reallocating CPU buffer from size 0.00 MiB to 84.25 MiB
> [DEBUG] ggml_extend.hpp:1126 - t5 compute buffer size: 84.25 MB(RAM)
> [DEBUG] conditioner.hpp:1353 - computing condition graph completed, taking 9259 ms
> ggml_gallocr_reserve_n: reallocating CPU buffer from size 0.00 MiB to 84.25 MiB
> [DEBUG] ggml_extend.hpp:1126 - t5 compute buffer size: 84.25 MB(RAM)
> sd: /build/gblmj4jz1z0xx3ls4s5xv1wd7s79hdj7-source/ggml/src/ggml-cpu/ggml-cpu.c:10104: ggml_compute_forward_soft_max_f32: Assertion `sum > 0.0' failed.
> sd: /build/gblmj4jz1z0xx3ls4s5xv1wd7s79hdj7-source/ggml/src/ggml-cpu/ggml-cpu.c:10104: ggml_compute_forward_soft_max_f32: Assertion `sum > 0.0' failed.
> sd: /build/gblmj4jz1z0xx3ls4s5xv1wd7s79hdj7-source/ggml/src/ggml-cpu/ggml-cpu.c:10104: ggml_compute_forward_soft_max_f32: Assertion `sum > 0.0' failed.
> ```
> 
> edit: Im at [322fad2](https://github.com/leejet/stable-diffusion.cpp/commit/322fad2df2f630215b6d338cadcc5618110ebade) , cuda

That's with `--diffusion-fa`, right?

Green-Sky:
> > I just tried it and it asserts:
> > ```
> > [DEBUG] t5.hpp:398  - token length: 512
> > ggml_gallocr_reserve_n: reallocating CPU buffer from size 0.00 MiB to 84.25 MiB
> > [DEBUG] ggml_extend.hpp:1126 - t5 compute buffer size: 84.25 MB(RAM)
> > [DEBUG] conditioner.hpp:1353 - computing condition graph completed, taking 9259 ms
> > ggml_gallocr_reserve_n: reallocating CPU buffer from size 0.00 MiB to 84.25 MiB
> > [DEBUG] ggml_extend.hpp:1126 - t5 compute buffer size: 84.25 MB(RAM)
> > sd: /build/gblmj4jz1z0xx3ls4s5xv1wd7s79hdj7-source/ggml/src/ggml-cpu/ggml-cpu.c:10104: ggml_compute_forward_soft_max_f32: Assertion `sum > 0.0' failed.
> > sd: /build/gblmj4jz1z0xx3ls4s5xv1wd7s79hdj7-source/ggml/src/ggml-cpu/ggml-cpu.c:10104: ggml_compute_forward_soft_max_f32: Assertion `sum > 0.0' failed.
> > sd: /build/gblmj4jz1z0xx3ls4s5xv1wd7s79hdj7-source/ggml/src/ggml-cpu/ggml-cpu.c:10104: ggml_compute_forward_soft_max_f32: Assertion `sum > 0.0' failed.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > edit: Im at [322fad2](https://github.com/leejet/stable-diffusion.cpp/commit/322fad2df2f630215b6d338cadcc5618110ebade) , cuda
> 
> That's with `--diffusion-fa`, right?

Yes, without it triggers the ASSERT just once (not sure why it glitches like that...)
So it does not matter.

Green-Sky:
836fd723264392e5b9cfc16490533e262113f5a8 is the first bad commit

edit: also on cpu (the error kind of indicated that)

Green-Sky:
If I build in release, it gets further but still dies:
```
[DEBUG] ggml_extend.hpp:1126 - t5 compute buffer size: 84.25 MB(RAM)
[DEBUG] conditioner.hpp:1353 - computing condition graph completed, taking 9854 ms
[DEBUG] ggml_extend.hpp:1126 - t5 compute buffer size: 84.25 MB(RAM)
[DEBUG] conditioner.hpp:1353 - computing condition graph completed, taking 19615 ms
[INFO ] stable-diffusion.cpp:1390 - get_learned_condition completed, taking 39012 ms
[INFO ] stable-diffusion.cpp:1413 - sampling using Euler method
[INFO ] stable-diffusion.cpp:1450 - generating image: 1/1 - seed 42
[DEBUG] stable-diffusion.cpp:819  - Sample
[DEBUG] flux.hpp:1099 - Forcing guidance to 0 for chroma model (SD_CHROMA_ENABLE_GUIDANCE env variable to "ON" to enable)
ggml_backend_cuda_buffer_type_alloc_buffer: allocating 1907.00 MiB on device 0: cudaMalloc failed: out of memory
ggml_gallocr_reserve_n: failed to allocate CUDA0 buffer of size 1999638144
[ERROR] ggml_extend.hpp:1119 - flux: failed to allocate the compute buffer

[DEBUG] flux.hpp:1099 - Forcing guidance to 0 for chroma model (SD_CHROMA_ENABLE_GUIDANCE env variable to "ON" to enable)
Segmentation fault (core dumped)
```

and with diffusion-fa :
```
[DEBUG] ggml_extend.hpp:1126 - t5 compute buffer size: 84.25 MB(RAM)
[DEBUG] conditioner.hpp:1353 - computing condition graph completed, taking 9781 ms
[DEBUG] ggml_extend.hpp:1126 - t5 compute buffer size: 84.25 MB(RAM)
[DEBUG] conditioner.hpp:1353 - computing condition graph completed, taking 19578 ms
[INFO ] stable-diffusion.cpp:1390 - get_learned_condition completed, taking 38554 ms
[INFO ] stable-diffusion.cpp:1413 - sampling using Euler method
[INFO ] stable-diffusion.cpp:1450 - generating image: 1/1 - seed 42
[DEBUG] stable-diffusion.cpp:819  - Sample
[DEBUG] flux.hpp:1099 - Forcing guidance to 0 for chroma model (SD_CHROMA_ENABLE_GUIDANCE env variable to "ON" to enable)
/build/jxshlkj34r3s9yfmhsihki3q9mzvmpm3-source/ggml/src/ggml.c:4326: GGML_ASSERT(mask->ne[1] >= GGML_PAD(q->ne[1], GGML_KQ_MASK_PAD) && "the Flash-Attention kernel requires the mask to be padded to GGML_KQ_MASK_PAD and at least n_queries big") failed
ptrace: Operation not permitted.
No stack.
The program is not being run.
Aborted (core dumped)
```

so flash attention also needs some love too.

Green-Sky:
stack:
```
#5  0x0000000000524ab5 in ggml_compute_forward_soft_max_f32 (params=params@entry=0x7ffd462987e0, dst=dst@entry=0xb23f110)
    at ggml/src/ggml-cpu/ggml-cpu.c:10104
#6  0x0000000000524b86 in ggml_compute_forward_soft_max (params=params@entry=0x7ffd462987e0, dst=dst@entry=0xb23f110)
    at ggml/src/ggml-cpu/ggml-cpu.c:10127
#7  0x0000000000532efc in ggml_compute_forward (params=params@entry=0x7ffd462987e0, tensor=0xb23f110)
    at ggml/src/ggml-cpu/ggml-cpu.c:14100
#8  0x000000000053329e in ggml_graph_compute_thread (data=0x8c75580) at ggml/src/ggml-cpu/ggml-cpu.c:15064
#9  0x0000000000533344 in ggml_graph_compute._omp_fn.0 () at ggml/src/ggml-cpu/ggml-cpu.c:15339
#10 0x00007ffff7f9174e in gomp_thread_start () from ...
```

`nc` is `256`, `*dp@nc` is `{-nan(0x400000) <repeats 256 times>}`, `*wp@nc` is `{-inf <repeats 256 times>}`

btw, again, I am running chroma v33 q4_k.
`t5xxl_q8_0.gguf`

stduhpf:
> If I build in release, it gets further but still dies:
> 
> ```
> [DEBUG] ggml_extend.hpp:1126 - t5 compute buffer size: 84.25 MB(RAM)
> [DEBUG] conditioner.hpp:1353 - computing condition graph completed, taking 9854 ms
> [DEBUG] ggml_extend.hpp:1126 - t5 compute buffer size: 84.25 MB(RAM)
> [DEBUG] conditioner.hpp:1353 - computing condition graph completed, taking 19615 ms
> [INFO ] stable-diffusion.cpp:1390 - get_learned_condition completed, taking 39012 ms
> [INFO ] stable-diffusion.cpp:1413 - sampling using Euler method
> [INFO ] stable-diffusion.cpp:1450 - generating image: 1/1 - seed 42
> [DEBUG] stable-diffusion.cpp:819  - Sample
> [DEBUG] flux.hpp:1099 - Forcing guidance to 0 for chroma model (SD_CHROMA_ENABLE_GUIDANCE env variable to "ON" to enable)
> ggml_backend_cuda_buffer_type_alloc_buffer: allocating 1907.00 MiB on device 0: cudaMalloc failed: out of memory
> ggml_gallocr_reserve_n: failed to allocate CUDA0 buffer of size 1999638144
> [ERROR] ggml_extend.hpp:1119 - flux: failed to allocate the compute buffer
> 
> [DEBUG] flux.hpp:1099 - Forcing guidance to 0 for chroma model (SD_CHROMA_ENABLE_GUIDANCE env variable to "ON" to enable)
> Segmentation fault (core dumped)
> ```

What if you use small resolutions without fa? Does it still crash?

Green-Sky:
> What if you use small resolutions without fa? Does it still crash?

512x512 is the same. It's independent of fa. (also t5 f16 same issue)

Did you try compiling for debug with asserts enabled?
tip: you can enable "debug" optimizations (-Og for gcc/clang), which enable most of the useful optimizations without hindering debugability too much.


edit: @stduhpf are you running with an updated ggml perhaps?

stduhpf:
> edit: @stduhpf are you running with an updated ggml perhaps?

I was using updated GGML during development (for vulkan bf16 support), bud I did check for crashes on release cpu build with the ggml version used by the master branch. (And it seemed to work on my machine)
Maybe the crashes are related to the debug?

Green-Sky:
> > edit: @stduhpf are you running with an updated ggml perhaps?
> 
> I was using updated GGML during development (for vulkan bf16 support), bud I did check for crashes on release cpu build with the ggml version used by the master branch. (And it seemed to work on my machine) Maybe the crashes are related to the debug?

Probably not, in release it crashes a little later (see prev comments).

stduhpf:
Well, it works on my machine... Both CPU and Vulkan builds work fine (as long as there is no ggml b16 tensors in the file).

I can confirm flash attention doesn't seem work , though it worked at 64x64 for some reason.

Green-Sky:
Full sd info:
```
Option:
    n_threads:         8
    mode:              txt2img
    model_path:
    wtype:             unspecified
    clip_l_path:
    clip_g_path:
    t5xxl_path:        models/flux-extra/t5xxl_fp16.safetensors
    diffusion_model_path:   models/chroma-unlocked-v33-q4_k.gguf
    vae_path:          models/flux-extra/ae-f16.gguf
    taesd_path:
    esrgan_path:
    controlnet_path:
    embeddings_path:
    stacked_id_embeddings_path:
    input_id_images_path:
    style ratio:       20.00
    normalize input image :  false
    output_path:       output.png
    init_img:
    mask_img:
    control_image:
    clip on cpu:       false
    controlnet cpu:    false
    vae decoder on cpu:false
    diffusion flash attention:false
    strength(control): 0.90
    prompt:            Photograph of the Alps. Vivid summer afternoon. Everything is perfectly in focus.
    negative_prompt:   low quality, ugly, unfinished, out of focus, blury, text, sketch, cartoony, bad anatomy, amateurish
    min_cfg:           1.00
    cfg_scale:         4.50
    slg_scale:         0.00
    guidance:          0.00
    eta:               0.00
    clip_skip:         -1
    width:             64
    height:            64
    sample_method:     euler
    schedule:          default
    sample_steps:      16
    strength(img2img): 0.75
    rng:               cuda
    seed:              42
    batch_count:       1
    vae_tiling:        true
    upscale_repeats:   1
System Info:
    SSE3 = 1
    AVX = 1
    AVX2 = 1
    AVX512 = 0
    AVX512_VBMI = 0
    AVX512_VNNI = 0
    FMA = 1
    NEON = 0
    ARM_FMA = 0
    F16C = 1
    FP16_VA = 0
    WASM_SIMD = 0
    VSX = 0
```

I did quantize chroma myself with sd.cpp previously, not sure if that matters.

stduhpf:
OK, just tried a debug build, I can confirm it crashes.

stduhpf:
Something wrong is going on in ggml (https://github.com/ggml-org/ggml/blob/ff9052988b76e137bcf92bb335733933ca196ac0/src/ggml-cpu/ggml-cpu.c#L2297) with the softmax when all the tokens are masked . Since the masked values are all set to `-INFINITY`, the max is also `-INFINITY`, and that breaks the computation of the sum, which triggers the assert. (also happens on updated ggml)

This is an edge case that could probably be handled explcitly, but it should not happen in a normal situation.

I'm not sure why all the tokens are masked here though.

stduhpf:
@Green-Sky I think I fixed the T5 assert issue. 

No Idea what kind of mask `ggml_flash_attn_ext()` expects though. I tried padding it with zeroes, or repeating it along the second dimension to avoid triggering the assert, but then the model generates black images.

stduhpf:
Sadly I think this also fixed the trippy visuals when only DIT masking is disabled.

Green-Sky:
> @Green-Sky I think I fixed the T5 assert issue.

Yeah, works now :)

> No Idea what kind of mask `ggml_flash_attn_ext()` expects though. I tried padding it with zeroes, or repeating it along the second dimension to avoid triggering the assert, but then the model generates black images.

I will look into that tomorrow.



> Sadly I think this also fixed the trippy visuals when only DIT masking is disabled.

heh

---

Btw, when I generate 64x64 the vae crashes..., but I don't think its this pr.

Green-Sky:
@stduhpf got it working, I did the same (wrong) thing first with padding as you did, but then I realized it just needed a transpose (and cast to f16)

opened pr https://github.com/stduhpf/stable-diffusion.cpp/pull/2

Green-Sky:
> This is a scheduled Windows Server 2019 brownout. The Windows Server 2019 image will be removed on 2025-06-30. For more details

"windows-latest" is actually "windows-2019"

Green-Sky (approved):
There are still a bunch of TODOs with hardcoded values, are you going to resolve them?

Also, note for everyone, flash attention with masking is broken everywhere except cuda, I suspect an upstream issue. (its not really tested)

stduhpf:
> There are still a bunch of TODOs with hardcoded values, are you going to resolve them?

Some todos were outdated, I removed them. As for the hardcoded values (it's the expected number of double and single blocks), I think it's fine to keep them hardcoded as long as there is no block pruning (like Flux Lite) or upscaled models based on Chroma.

Green-Sky:
@stduhpf where is the mask set? I can't find the `-INFINITY` related to that.

I had the idea, that maybe the mask is wrongly positive `INFINITY`, instead of negative `INFINITY` (cpu backend test for `-INFINITY` in mask)

stduhpf:
@Green-Sky in t5.hpp, line 438 (i used `-HUGE_VALF`, which is an alias for `-(float)INFINITY`

Green-Sky:
`chroma_modify_mask_to_attend_padding()` gets called multiple times, ending up unmasking more and more.

eg two step sampling:
```
[INFO ] stable-diffusion.cpp:1413 - sampling using Euler method
[INFO ] stable-diffusion.cpp:1450 - generating image: 1/1 - seed 42
[DEBUG] stable-diffusion.cpp:819  - Sample
[DEBUG] flux.hpp:1108 - Forcing guidance to 0 for chroma model (SD_CHROMA_ENABLE_GUIDANCE env variable to "ON" to enable)
[DEBUG] flux.hpp:725  - PAD: 1
[DEBUG] flux.hpp:734  - MASKED: 492
[DEBUG] ggml_extend.hpp:1138 - flux compute buffer size: 324.34 MB(RAM)
[DEBUG] flux.hpp:1108 - Forcing guidance to 0 for chroma model (SD_CHROMA_ENABLE_GUIDANCE env variable to "ON" to enable)
[DEBUG] flux.hpp:725  - PAD: 1
[DEBUG] flux.hpp:734  - MASKED: 491
  |=========================>                        | 1/2 - 124.44s/it[DEBUG] flux.hpp:1108 - Forcing guidance to 0 for chroma model (SD_CHROMA_ENABLE_GUIDANCE env variable to "ON" to enable)
[DEBUG] flux.hpp:725  - PAD: 1
[DEBUG] flux.hpp:734  - MASKED: 490
  |==================================================| 2/2 - 124.99s/it
[INFO ] stable-diffusion.cpp:1489 - sampling completed, taking 249.50s
```

stduhpf:
> `chroma_modify_mask_to_attend_padding()` gets called multiple times, ending up unmasking more and more.

Nice catch!

Amin456789:
@stduhpf could please please release this in ur repo for now
i really wanna try it. i need avx2 compiled  exe
thank u

Amin456789:
@LostRuins please add this to ur great app too

LostRuins:
Indeed I will, once it's merged here. stduhpf is doing very fine work.

stduhpf:
@Amin456789 Done.

LostRuins:
Well it works, sort of. 
I dunno if it's because I am using lousy settings or a poor quant.

cat holding sign that says EVIL CAT

Prompt is `cat holding sign that says EVIL CAT`
This is chroma v35 at q4_0. Euler at 20 steps, cfg 5.0. 768x768
The cat's face looks *really* mashed.

Edit: Also, I had to double this to get it to work at higher resolutions
https://github.com/leejet/stable-diffusion.cpp/blob/10c6501bd05a697e014f1bee3a84e5664290c489/stable-diffusion.cpp#L1554

Green-Sky:
@LostRuins try adding a simple negative prompt.

stduhpf:
> @LostRuins try adding a simple negative prompt.

Ah maybe my fix to avoid including more and more padding at each diffusion step messed up the empty negative prompts. I can't check it today.

LostRuins:
Hmm it is possible. I use the exact same seed and settings, but with the negative prompt `ugly` 
`parse 'ugly' to [['ugly', 1], ]`

this is the result
777800-cat holding sign that says EVIL CAT ### ugly

LostRuins:
I will be happy to test any fix you have on the same prompt and seed again.

stduhpf:
> I will be happy to test any fix you have on the same prompt and seed again.

You could try with the penultimate commit (https://github.com/leejet/stable-diffusion.cpp/pull/696/commits/efaa137bc0fb6ac098ced5600dfe2b17e754eb4d). The padding is technically incorrect with this one, but it doesn't  seem to matter in practice.

LostRuins:
777800-cat holding sign that says EVIL CAT

777800-cat holding sign that says EVIL CAT ### ugly


**@stduhpf @Green-Sky Significant improvement for the case with no negative prompt** 

I used the same seeds for both cases. Top one has no negative prompt, bottom one does (same as earlier test). 

Oddly, the one with negative prompt has changed too, it seems a bit worse now? Could just be variance though.

stduhpf:
> Oddly, the one with negative prompt has changed too, it seems a bit worse now? Could just be variance though.

It could be because of the incorrect padding.

LostRuins:
I see. I will save my settings and try again to compare after it's fixed then. Will provide a good benchmark. Thanks!

stduhpf:
@LostRuins Did you quantize v35 to q4_0 and convert to GGUF yourself or did you use someone else's quantized model? 

(I'm trying to reproduce your issue, on my machine, on the lastest commit, it looks like it using [silveroxides](https://huggingface.co/silveroxides)' v29 q4_0 , with the other settings same as you, without a negative prompt)
output
That doesn't look as broken to me. So maybe it's a problem with the model? I need to try with v35, but I'd like to know which one to download.

LostRuins:
I used a preconverted model. You can try it here: https://huggingface.co/silveroxides/Chroma-GGUF/blob/main/chroma-unlocked-v35/chroma-unlocked-v35-Q4_0.gguf

T5 from `t5xxl_fp8_e4m3fn.safetensors` from Flux Dev
No Clip-L was used.
`ae.safetensors` VAE from Flux Dev
No LoRAs were used.

Generation params: 
Prompt: `cat holding sign that says "EVIL CAT"`
Seed: `777800`
Steps: `20`
Sampler: `Euler`
Width: `768`
Height: `768`
CFG Scale: `5`
Clip-Skip: `0` (which becomes default 2 by sd.cpp logic)
NO negative prompt (important)

The image and both cases are reproducible by me on the same settings.

LostRuins:
But actually even for your example it does look at bit mashed too. Look at the ears, compare them if you revert the newest commit on the same seed. 

image

Compare your cat ears with mine above

stduhpf:
`sd.exe --diffusion-model models/chroma-unlocked-v35-Q4_0.gguf --t5xxl models/t5xxl_q4_k.gguf --vae models/ae.f16.gguf -p "cat holding sign that says EVIL CAT" --cfg-scale 5 --sampling-method euler -W 768 -H 768 -s 777800`
output

I'm using t5xxl_q4_k.gguf though, maybe that could be the cause of the difference? Or maybe you're using flash attention?

stduhpf:
`sd.exe --diffusion-model models/chroma-unlocked-v35-Q4_0.gguf --t5xxl models/t5xxl_fp8_e4m3fn.safetensors --vae models/ae.f16.gguf -p "cat holding sign that says EVIL CAT" --cfg-scale 5 --sampling-method euler -W 768 -H 768 -s 777800`
output

Ok that's a bigger difference than I expected from just chosing a different T5 quant. But it's still not looking too broken on my end...

LostRuins:
um, the paws in your last image are completely mashed. This is not flux quality heheh... Unless chroma itself is broken, but even Sd1.5 can generate better paws that that 😂

image

Can you try the last option, but first revert your https://github.com/stduhpf/stable-diffusion.cpp/commit/e22c57cc1ee6b5aef01a936ade953e9b202aac20 , keep the same seed and everything else the same, and compare the result? How does it look?

stduhpf:
> Can you try the last option, but first revert your https://github.com/stduhpf/stable-diffusion.cpp/commit/e22c57cc1ee6b5aef01a936ade953e9b202aac20 , keep the same seed and everything else the same, and compare the result? How does it look?

output

> um, the paws in your last image are completely mashed. This is not flux quality heheh... Unless chroma itself is broken, but even Sd1.5 can generate better paws that that 😂
image

This could just be an effect of the quantization to be honest. q4_0 is not the best for quality.

stduhpf:
With `chroma-unlocked-v35-Q48_0.gguf` (and t5xxl fp8)
output

LostRuins:
Haha well I dunno... maybe someone else can chip in. Could indeed just be the model or quant maybe?

stduhpf:
@LostRuins  Are you using flash attention or not? 

Because after looking quickly into it, it seems like flash attention only works with either causal attention masks or no mask at all, and in the case of Chroma, the mask is weirdly shaped.

LostRuins:
That requires `SD_USE_FLASH_ATTENTION` manually defined right? If so, I am not using it

stduhpf:
Then I'm clueless.

LostRuins:
Alright, well we'll go with whatever solution you think works best. I'm just rather surprised at how wonky the chroma outputs are. Maybe it is just the model. But I think you'd agree that these are a big step down from Flux subjectively speaking.

LostRuins:
image


Anyway I believe there's still something wrong, both with/without the final commit in this PR, but I can't put my finger on it. Sometimes the object in the image is very clear and looks fine, other times there are just weird artifacts in some way or another. I have no idea if it's just the model or something else is wrong. Sometimes the exact same prompt gives good results for one seed and bad results with another.

The prompt for the above was "cat" with various seeds and no negative prompt at 512x512px

stduhpf:
> Anyway I believe there's still something wrong, both with/without the final commit in this PR, but I can't put my finger on it. Sometimes the object in the image is very clear and looks fine, other times there are just weird artifacts in some way or another. I have no idea if it's just the model or something else is wrong. Sometimes the exact same prompt gives good results for one seed and bad results with another.
> 
> The prompt for the above was "cat" with various seeds and no negative prompt at 512x512px

| ComfyUI | This PR |
| --- | --- |
| image | output |

As far as I'm concerned, this is an issue with the model itself. It doesn't seem to handle short prompts and/or low resolutions very well.

LostRuins:
Ah i see. I don't use comfy so I can't compare. In that case, cool. Should be good to go then

LostRuins (approved):

rmatif:
I was going to say the same thing. Generally, when it struggles like that, it's most likely because it doesn't handle low resolution well, you can experience that with other models too

stduhpf:
Still 512x512, with an "ai upscaled" prompt (i used a LLM to come up with a prompt for a "cat" picture):

> A majestic Maine Coon cat with thick, flowing brown tabby fur and piercing golden eyes, sitting regally on an ornate velvet cushion by a sunlit window. The warm afternoon light casts soft highlights on its fur, creating a luxurious texture. The background features a cozy Victorian-style room with bookshelves and a steaming teacup. Rendered in ultra-realistic 8K detail with subtle painterly touches, cinematic depth of field, and a dreamy atmosphere.

image

1024x1024 with simple "cat" prompt:
output

stduhpf:
Both changes:
output

stduhpf:
It seems to also be related to the attention masking. Looks like not being able to attend to enough tokens kind of breaks T5.

| With masking (default) | T5 masking only (disabled for DiT)  | DiT masking only (disabled for t5) | No masking at all |
| --- | --- | --- | --- |
| image | image | dit_mask | no_mask |

stduhpf:
Hmm actually I'm staring to think T5 might not be supposed to use masking by default. I thought it was, but it looks like the ComfyUI implementation doesn't use it. The model is still broken with short prompts and low resolutions when t5 masking is disabled, but not quite as much (just like the Comfyui implementation).

LostRuins:
Will this change affect regular flux too? Or just chroma

stduhpf:
Just Chroma. Regular Flux doesn't use any kind of attention masking

LostRuins:
777800-cat holding sign that says EVIL CAT

Applied your patch. Same mashed evil cat seed as https://github.com/leejet/stable-diffusion.cpp/pull/696#issuecomment-2953872885 , looks slightly less mashed now. But this cat still looks like it needs to visit a vet.

Edit: I spoke too soon. Once I added "ugly" as a negative prompt I got this monstrosity.

777800-cat holding sign that says EVIL CAT ### ugly

stduhpf:
@LostRuins I'm confused. These look like this on my end ( 768x768, cfg scale 5, euler sampling, seed 777800, chroma-unlocked-v35-Q4_0.gguf (silveroxides), t5xxl_fp8_e4m3fn.safetensors)

output copy
output

It's far from perfect either, but it's clearly not as broken. Could it be something weird happening on koboldcpp end? Have you tried running it from sdcpp cli?

LostRuins:
I would say they are in the same ballpark. Your second image is basically what I am seeing too. 

image

Yeah I do run sd.cpp locally outside of kcpp, same results. Also other models are unaffected. I don't really modify the backend/graph so those parts should be the same.

stduhpf:
It's strange that the text on the sign is almost perfectly identical, but the cat's faces are completely different.

LostRuins:
Yeah. So strange. I am using the same HEAD commit as this, so perhaps there are some variances with either the quant/t5xxl i am using, or perhaps hardware related precision errors? Strangely all of these are related to chroma.

Here's compared with **Flux Dev 1**. 
- **Same T5XXL**
- **Same VAE**
- **Same prompt**
- **Same seed, sampler settings, step count and prompt, with no negative prompt.**
- Added Flux Dev 1 Clip-L

Flux Dev 1 Result:

image

Btw I did notify the creator of Chroma as feedback, and got this reply...

```
That has nothing to do with the model and everything to do with the implementation in the PR.
If issues this severe were standard then don't you think that you would have seen more examples from other places of this?
btw if t5 in stable-diffusion.cpp previously did not require specific total chunk length or attention mask then adding attention mask is unnecessary
attention mask is only needed if one uses transformers default which pads the t5 model to 512.
by simply not padding it to 512 you effectively remove that part
also empty negatives produces artifacts since the model is dedistilled schnell from start and bfl did not use attention mask. So you either use negative to a reasonable token count or you pad negative to a set minimum if negative is empty.
Also single token prompting on a model with unpadded t5 encoder and no CLIP is obviously out of scope.
```

image
https://huggingface.co/silveroxides/Chroma-GGUF/discussions/8 they're on huggingface.

LostRuins:
@stduhpf I'm planning to integrate your chroma PR into koboldcpp soon, there's been quite a lot of back-and-forth with these different approaches. What do you think are the current best configs from what we've tried?

stduhpf:
> @stduhpf I'm planning to integrate your chroma PR into koboldcpp soon, there's been quite a lot of back-and-forth with these different approaches. What do you think are the current best configs from what we've tried?

I *think* the current default one (no t5 masking, Dit masking with 1 padding included) should be the best but I'm not very confident about it anymore.

wbruna:
@LostRuins , may I suggest targeting the bleedingedge release for Koboldcpp? Apart from Chroma support, it also includes many pending fixes for sd.cpp (like that VAE tile initialization bug), and both Koboldcpp and sd-server would benefit from any fixes needed for running with a persistent model in memory.

LostRuins:
Yes I have merged the VAE tile fix, what else is there? I don't see any `bleedingedge` branch and everything else is mostly up to date

wbruna:
> Yes I have merged the VAE tile fix, what else is there?

From the top of my head there is also leejet/stable-diffusion.cpp#484 and leejet/stable-diffusion.cpp#681 (apart from stuff that you probably fixed directly on Koboldcpp already, like leejet/stable-diffusion.cpp#658).

> I don't see any `bleedingedge` branch and everything else is mostly up to date

https://github.com/stduhpf/stable-diffusion.cpp/commits/bleedingedge/ , with several tagged releases.

LostRuins:
681 isnt needed for kobo because kobo doesn't load ckpt or diffusers models, only gguf and safetensors. 
484 seems to already be working? Tiled VAE is working fine. 

I'd rather not rebase off a significantly different codebase as it'll require a bunch of testing to ensure everything works and I might have to deal with possible regressions. Instead if there are any critical or important issues I can patch those in manually.

wbruna:
> 681 isnt needed for kobo because kobo doesn't load ckpt or diffusers models, only gguf and safetensors. 484 seems to already be working? Tiled VAE is working fine.

I gave an example on LostRuins/koboldcpp#1603, to avoid more unrelated discussions on this PR.

> I'd rather not rebase off a significantly different codebase as it'll require a bunch of testing to ensure everything works and I might have to deal with possible regressions. Instead if there are any critical or important issues I can patch those in manually.

Fair enough.

LostRuins:
Alright I added the tiling fix

flux.hpp
```diff
@@ -270,36 +279,64 @@ namespace Flux {
             blocks["txt_mlp.2"] = std::shared_ptr<GGMLBlock>(new Linear(mlp_hidden_dim, hidden_size));
         }
 
+        std::vector<ModulationOut> get_distil_img_mod(struct ggml_context* ctx, struct ggml_tensor* vec) {
+            // TODO: not hardcoded?
+            const int single_blocks_count = 38;
+            const int double_blocks_count = 19;
```
stduhpf, line 285: Looks like I'll have to implement this todo:
https://huggingface.co/lodestones/flux-ultra-lite

stduhpf: tensor 'model.diffusion_model.distilled_guidance_layer.in_proj.weight' has wrong shape in model file: got [32, 5120, 1, 1], expected [64, 5120, 1, 1]
 
 That's odd. Is it using a different VAE?

stduhpf: Hmm looks like it's not quite the same architecture as chroma. There a similarities, but somehow the "time_in" "vector_in" and "guidance_in" are back, as well as the modulation for double blocks only. And the distilled_guidance_layer doesn't have the same input shape...

Green-Sky: Gotta wait for more info then.

stduhpf: My current guess is that it's using the double blocks from flux lite and the single blocks from Chroma.

LostRuins: Nice. Hoping your earlier commits get merged by leejet soon.

stduhpf: Welp I only got it to generate static latent noise. 
We will definitely have to wait for more info to make it run properly.
output

leejet:
Thank you for your contribution!

Green-Sky:
https://github.com/ggml-org/llama.cpp/pull/14735
maybe this is related to the fa bugs we encountered. might try later.

LostRuins:
I saw **chroma v50 annealed** got released, so I gave that a spin with more "cat"

image

then i booted up flux

image

**meow meow meow meow :(** 
(btw any updates about that negative prompt padding silveroxides mentioned?)

stduhpf:
> (btw any updates about that negative prompt padding silveroxides mentioned?)

Not really, I didn't quite understand what they meant by that. Results seem to somewhat match what we get with similar settings in comfyUI, so I think we're good.

wbruna:
By the way, are any of you able to use Chroma with flash attention? Normal renders work fine, but `--diffusion-fa` only makes fully black images (chroma-unlocked-v46-flash-Q4_0.gguf with Vulkan).

Green-Sky:
> By the way, are any of you able to use Chroma with flash attention? Normal renders work fine, but `--diffusion-fa` only makes fully black images (chroma-unlocked-v46-flash-Q4_0.gguf with Vulkan).

Yes, it is up there in the thread somewhere. Somehow when we do flash attention with the mask (which you can turn off for worse results), something goes wrong.

ggml_extend.hpp
```diff
@@ -864,6 +864,18 @@ __STATIC_INLINE__ struct ggml_tensor* ggml_nn_attention_ext(struct ggml_context*
         v = ggml_reshape_3d(ctx, v, d_head, L_k, n_head * N);  // [N * n_head, L_k, d_head]
         v = ggml_cast(ctx, v, GGML_TYPE_F16);
 
+        if (mask != nullptr) {
+            mask = ggml_transpose(ctx, mask);
+
+            if (mask->ne[1] < GGML_PAD(q->ne[1], GGML_KQ_MASK_PAD)) {
+                LOG_DEBUG("mask dims %ld, %ld, %ld, %ld\n", mask->ne[0], mask->ne[1], mask->ne[2], mask->ne[3]);
+                LOG_DEBUG("needs padding, padding from %ld to %ld\n", mask->ne[1], GGML_PAD(q->ne[1], GGML_KQ_MASK_PAD));
+                mask = ggml_pad(ctx, mask, 0, GGML_PAD(q->ne[1], GGML_KQ_MASK_PAD) - mask->ne[1], 0, 0);
```
Green-Sky, line 873: Actually, now that I think of it. We should not pad with zeros, but with infs (or was it -inf).

stduhpf: Ah yes, -inf. (`-HUGE_VALF` if I remember correctly)

Green-Sky:
Test run with v50 (final) flash-heun (recommends 10steps, cfg 1).

prompt1: `photo of a lovely cat, sitting on the veranda. It is a late summer day.`
prompt2: `A Cute cat holding a sign that says: "Stable diffusion.cpp Now supports Chroma!"`

| settings | img1 | img2 |
|--------|--------|--------|
| q4_k heun+simple 16steps | output | output |
| q4_k heun+simple 10steps | output | output |
| q4_k heun+discrete 10steps | output | output |
| q4_k euler+discrete 10steps | output | output |
| q4_k euler+simple 10steps | output | output |

`~5.5s/it @ 768x768 on cuda`

**The simple schedule is important.** I tested this previously against flux, and it made no real difference, but not here.

I **think** the quantization is more detrimental to this model than other flux models. You can somewhat recover the artifacts with more steps like usual.
I will check how others are quantizing the models and see how that impacts quality.

Also this is a short prompt. The quality increases with prompt length with this model.

Tests done with mostly defaults otherwise, no flash attention.

Simple schedule from https://github.com/leejet/stable-diffusion.cpp/pull/675

ggml_extend.hpp
Green-Sky: Not that it matters too much, since we are not hitting this code path anyway.

LostRuins:
EDIT: Simple schedule does seem marginally better. Maybe it should be enforced?
it works... it's just... feels like SD1.5 with added text.

stduhpf:
> it works... it's just... feels like SD1.5 with added text.

I think that's the point? It's a undistilled, general purpose base model that could be fine-tuned further for better results just like sd1.5.

Green-Sky:
> EDIT: Simple schedule does seem marginally better. Maybe it should be enforced? it works... it's just... feels like SD1.5 with added text.

We don't even default to `euler` for flux models... But I agree. We need a better organisaiton for defaults, since there are so many different models and architectures now. And we perform detection in code anyway.

> > it works... it's just... feels like SD1.5 with added text.
> 
> I think that's the point? It's a undistilled, general purpose base model that could be fine-tuned further for better results just like sd1.5.

Yep, the focus was to dedistill and learn missing stuff, mostly nsfw I guess.
And you can force "nice" images with adding "aesthetic 10, aesthetic 9, aesthetic 8", eg.

edit: also ran it `sd --diffusion-model /run/media/green/d20c801b-7aae-4042-85ec-bf2153257be8/green/workspace/stable-diffusion.cpp/models/chroma-unlocked-v50-flash-heun-q4_k.gguf --cfg-scale 1 -t 8 --t5xxl models/flux-extra/t5xxl_fp16.safetensors --vae models/flux-extra/ae-f16.gguf --steps 12 --sampling-method heun --schedule simple -W 768 -H 768 --vae-tiling -p 'A Cute cat holding a sign that says: "Stable diffusion.cpp Now supports Chroma!". aesthetic 10, aesthetic 9, aesthetic 8' -v`

output

FeiTianXi:
When using the Chroma model to generate images for character consistency, the custom Collector (Advanced) prompts an error! AttributeError: 'Chroma' object has no attribute 'time_in'.
The Flux model's raw images can be used normally!

PR #697, opened 2025-05-30
fix: allow resetting clip_skip to its default value

wbruna:
CLIPTextModel currently ignores attempts to set clip_skip back to -1, retaining the previously set value instead. While this is not an issue to the sd command (which does not support changing clip_skip between generations), it affects frontends that reuse model instances for multiple images.

Since each model version's default clip_skip value is defined by its respective Conditioner class, it needs to be applied every time they get a different clip_skip value, so move that logic from their constructors into their set_clip_skip methods.

Related: LostRuins/koboldcpp#1546

(the original PR #687 was closed because I submitted it on the master branch by mistake)

### DIFF
--- a/clip.hpp
+++ b/clip.hpp
@@ -678,8 +678,8 @@ public:
     bool with_final_ln        = true;
 
     CLIPTextModel(CLIPVersion version = OPENAI_CLIP_VIT_L_14,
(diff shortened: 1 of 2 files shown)


leejet:

PR #702, opened 2025-06-12
Add validation for Control Net image dimensions

Vsevo:


stduhpf:
I'm thinking it may be better to just resize the Control Net images to the desired dimensions, like is done for the input images.

Vsevo:
Good point! I actually thought about doing it that way first, but went with minimal changes instead. I'll make the update.

leejet:
This commit https://github.com/leejet/stable-diffusion.cpp/pull/778/commits/33ff442c1daa6cef0c1b2dd958466e0153a023c4 unified the processing of image loading, so there is no need to check the dimensions of the controlnet images any longer. Anyway, thank you for your contribution.

PR #703, opened 2025-06-17
Tiling: force zero-initialize output

### LINKED ISSUE #1603: VAE tiling may cause image interference between generations
With the default VAE tiling, some image generations can seemingly be corrupted by previously generated images.

A test with plain text2img, model cyberrealisticPony_semiRealV30, DMD LoRA at 1.0, LCM 8 steps, CFG 1, Seed 1, 1024x640, VAE tiling on, in sequence (no batch, just clicking Generate repeatedly):

* prompt: "car", 5 generations:

Image

* prompt: "forest", 1 generation:

Image

* prompt "forest", 2 generations:

Image

* prompt "forest", 1 generation:

Image

These were generated on rocm build a7706be , but I see the same behavior with Vulkan on the 1.93.2 build.

I can't reproduce it when disabling VAE tiling, so this could be related to leejet/stable-diffusion.cpp#588 (but I get a warning "Requested buffer size (4362076160) exceeds device memory allocation limit (4294967292)!" when disabling VAE tiling, so I don't know if I can really trust this test).

stduhpf:
Should fix https://github.com/LostRuins/koboldcpp/issues/1603

### DIFF
--- a/ggml_extend.hpp
+++ b/ggml_extend.hpp
@@ -597,6 +597,8 @@ typedef std::function<void(ggml_tensor*, ggml_tensor*, bool)> on_tile_process;
 
 // Tiling
 __STATIC_INLINE__ void sd_tiling(ggml_tensor* input, ggml_tensor* output, const int scale, const int tile_size, const float tile_overlap_factor, on_tile_process on_processing) {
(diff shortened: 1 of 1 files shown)


leejet:

PR #716, opened 2025-07-03
cleanup: do not force VAE type to f32 on SDXL

wbruna:
This seems to be a leftover from the initial SDXL support: it's not enough to avoid NaN issues, and it's not not needed for the fixed sdxl-vae-fp16-fix .


leejet:

PR #724, opened 2025-07-07
feat: overriding quant types for specific tensors on model conversion

wbruna:
Allows specifying weight quantization types using regex patterns, with a syntax similar to the llama.cpp tensor overrides.


leejet:

PR #725, opened 2025-07-07
added --size and --aspect-ratio argument

kuesji:


stduhpf:
You should probably add a way to ensure the dimensions are multiples of 64 (with the aspect ratio).

kuesji:
i thought leaving size check to that part of code makes more sense. 

https://github.com/leejet/stable-diffusion.cpp/blob/dafc32d0dd0922272f970c93f961834af460f013/examples/cli/main.cpp#L716C1-L724C6

leejet:
Thank you for your contribution, but I don't think these two arguments are helpful.

wbruna:
Since the user isn't specifying one of the dimensions, a calculation that approximates better the missing dimension for a multiple of 64 could be more useful, especially for preserving image ratio on img2img runs.

I did something similar on Koboldcpp, in LostRuins/koboldcpp#1586 : in that case, it adapts the ratio to a desired image area, but getting a single dimension from the other + ratio wouldn't be much different.

PR #726, opened 2025-07-08
Fix conversion between F64 and F32

stduhpf:
Partially adresses https://github.com/leejet/stable-diffusion.cpp/pull/681#issuecomment-3049713263

Quantization/dequantization between F64 and F32 wasn't being handled properly, this was causing some issues.

F64 support in most GGML backends isn't really complete right now, so models using F64 weights can still make the program crash during inference. At least it's possible to convert F64 models to F32 with these changes.


leejet:
Thank you for your contribution. But I think it would be better to convert f64 to f32 and i64 to i32 directly when loading from file. You can check out my latest commit.

PR #738, opened 2025-07-23
Add sd.cpp-webui as an available frontend

daniandtheweb:
[sd.cpp-webui](https://github.com/daniandtheweb/sd.cpp-webui) is a frontend for stable-diffusion.cpp that I've been developing for a while.
It started as a simple way to test changes to the backend as it's mostly a command generator that calls directly the sd binary but that allows to easily swap between versions just by pasting in the main sd.cpp-webui folder a different binary version, no reload required.

Right now this frontend has most of the features of the main program (I'm working on adding the missing ones) and has a built in gallery with a metadata reader to easily access the image information of the images generated with this program (there's also a fallback that tries to read the metadata of images originally created on ComfyUI so that prompts can be easily reused).

I'd love to get some feedback on this and if anyone has any improvement to suggest I'd be happy to hear it.


leejet:
Great job!

hartmark:
This is quite cool wrapper, but I think it would be nice if sd-binary could stay resident in memory and cache the models so any following subsequent image renderings would go faster.

PR #741, opened 2025-07-24
Add LocalAI to README's UIs

mudler:
LocalAI integrates stable-diffusion.cpp as well!

It's been long time actually that we have integrated, but we didn't saw a section here in the README - congrats for the work, this is an amazing project and wish for the best!


leejet:
Great!

PR #743, opened 2025-07-27
add missing models and parameters to image metadata

wbruna:
This updates and supersedes #316 . I adjusted the original PR to make use of the new API functions, but preserved the original author, since it was just a minor tweak.

I also included diffusion and text encoder models into the metadata, following the auto111 format. But it's not clear how multiple text encoders should be included: all examples I've found used a simple 'TE:' prefix.


leejet:

PR #745, opened 2025-07-30
SD1.x models with a Tiny U-Net

akleine:
Add support for running  SD1.x models with a Tiny U-Net   in sd.cpp.  

These models are only 1 GB sized and need less running time, so there are good new perspectives for low power devices like some Android phones or Raspberry PI computers.


akleine:
Meanwhile I have found a much better way to solve this problem.
Will create another PR later.

PR #749, opened 2025-08-01
feat: support build against system installed GGML library

Seas0:
Just like in `llama.cpp` and `whisper.cpp`


Green-Sky (approved):

leejet:

wbruna:
Wouldn't this cause a `sizeof (struct ggml_tensor)` mismatch between the system ggml and sd.cpp, since sd.cpp overrides `GGML_MAX_NAME` default value?

@leejet @Green-Sky

Green-Sky:
Oh yea. We need `static_assert(GGML_MAX_NAME >= 128)` or similar.

PR #755, opened 2025-08-06
Support for ref images RoPE index (omni-kontext)

stduhpf:
Support for models like [omni-kontext](https://github.com/Saquib764/omini-kontext)

## How to use:

Download [omni-kontext character LoRA](https://huggingface.co/saquiboye/omini-kontext-character/tree/main) and add it to your lora model directory.

example command:
`sd.exe --diffusion-model models\flux1-kontext-dev-q5_1.gguf --clip_l models\clip\clip_l\clip_l.safetensors --t5xxl models\clip\t5\t5xxl_fp16.safetensors --vae models\vae\flux\ae.f16.gguf -W 1024 -H 512 -p "Add the character to the scene  <lora:character_5000:1>" -r 'scene_02.png' -r 'boy_reference_512.png,2' --cfg-scale 1 --sampling-method euler --color --lora-model-dir models\loras\flux\kontext\`

note the `,2` after the path to the second image. That's the desired RoPE index of the character's image (defaults to index 1).

## Results:

| Background | Character |
| --- | --- |
|   |  |

| | Without LoRA | With omni-kontext character LoRA |
| --- | --- | --- |
| master (index 1 for both images, using 2D RoPE offsets) | no_lora_no_indices | lora_no_indices |
| PR (index 2 for character) | no_lora_indices | lora_indices |


leejet:
I think it's not necessary to manually assign a number to each ref image. Instead, we can use a single parameter, such as --increase-ref-index, to control whether to increment the index.

stduhpf:
I like having the option to set it manually, but you're right, a `--increase-ref-index` would be more convenient.

leejet:

PR #760, opened 2025-08-07
make weight override more robust against ggml changes

wbruna:
Recently, GGML_TYPE_COUNT got bumped for the new GGML_TYPE_MXFP4 quant, getting it out-of-sync with SD_TYPE_COUNT.
    
To make it easier to build stable-diffusion.cpp against different ggml versions, adjust the type conversions to consider both GGML_TYPE_COUNT and SD_TYPE_COUNT as limits.

(this got committed on top of #716 by mistake; feel free to drop the first commit)


leejet:

PR #766, opened 2025-08-14
fix: avoid segfault for pix2pix models with no reference images

wbruna:


stduhpf:
What about setting the reference as an empty latent image if not provided?

wbruna:
@stduhpf , something like this?

Issue #761, opened 2025-08-08
VRAM exhaustion during SDXL inference on 8GB AMD GPU (Vulkan backend) causes fragmented VAE processing


magictext:
When running SDXL models using Vulkan backend on an AMD Radeon RX 6600 (8GB VRAM), the UNET computation exhausts nearly all available VRAM. This forces subsequent VAE decoding to operate in fragmented slices.

Optimization Suggestions:

**Model Swapping with Caching**
During batch image generation:

- Cache intermediate latent tensors after UNET computation

- Completely unload the UNET model from VRAM

- Load VAE model once and process all cached latents collectively

**Asynchronous CPU Offloading**
Implement pipelining where:

- GPU processes UNET for image *N+1*

- CPU simultaneously decodes VAE for image N (using cached latents)

These approaches could significantly reduce VRAM pressure and potentially improve throughput on GPUs with limited memory.

stduhpf:
Are you running sd.cpp as a library or from the cli? Because with the cli it should by default free all the unet-related buffers before the VAE decoding stage.

magictext:
YES 
I use the sd.cpp from cli
if I generate image without ---vae-tiling it will
```
[DEBUG] ggml_extend.hpp:1173 - clip compute buffer size: 1.40 MB(VRAM)
[DEBUG] ggml_extend.hpp:1173 - clip compute buffer size: 2.33 MB(VRAM)
[DEBUG] ggml_extend.hpp:1173 - clip compute buffer size: 2.33 MB(VRAM)
[DEBUG] conditioner.hpp:486  - computing condition graph completed, taking 179 ms
[DEBUG] ggml_extend.hpp:1173 - clip compute buffer size: 1.40 MB(VRAM)
[DEBUG] ggml_extend.hpp:1173 - clip compute buffer size: 2.33 MB(VRAM)
[DEBUG] conditioner.hpp:486  - computing condition graph completed, taking 267 ms
[DEBUG] conditioner.hpp:358  - parse '' to [['', 1], ]
[DEBUG] clip.hpp:311  - token length: 77
[DEBUG] ggml_extend.hpp:1173 - clip compute buffer size: 1.40 MB(VRAM)
[DEBUG] ggml_extend.hpp:1173 - clip compute buffer size: 2.33 MB(VRAM)
[DEBUG] ggml_extend.hpp:1173 - clip compute buffer size: 2.33 MB(VRAM)
[...]

PR #767, opened 2025-08-14
cli: show usage on unknown arg

stduhpf:
Makes troubleshooting easier. (ex: https://github.com/leejet/stable-diffusion.cpp/issues/761#issuecomment-3168224912)


leejet:

PR #768, opened 2025-08-14
feat: reduce CLIP memory usage with no embeddings

wbruna:
The CLIP weights need to be converted to f32 for textual inversions (fbd42b6fc16d14fbd362993fa1d083740a05f113, for #601), but that increases the amount of allocated VRAM even when embeddings aren't being used.

On a typical SDXL render on Vulkan, this change reduces peak VRAM usage around 190MB.

I'm including a refactor of the clip_skip handling, since this change would end up conflicting with it. Please let me know if I shall submit it separately instead.

### DIFF
--- a/clip.hpp
+++ b/clip.hpp
@@ -548,9 +548,15 @@ protected:
     int64_t embed_dim;
     int64_t vocab_size;
     int64_t num_positions;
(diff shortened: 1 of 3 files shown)


leejet:

PR #736, opened 2025-07-21
flash attention: correct head dim check and L_k padding

Green-Sky:
Gains are minimal, if any.
SD1 now has some usage of flash attention.

Some numbers where found by @bssrdf in #386 .

From what I can tell, the numbers should translate to vulkan and probably rocm pretty well.

## sd1

| model | backend | resolution | before sampling time | after sampling time | before comp buff | after comp buff
|--------|--------|--------|--------|--------|--------|--------|
| CyberRealistic_V9_FP16, 30steps, 5cfg | cuda | 512x768 | 20.37s | **19.57s** | 1220.07 MB | **1218.29 MB** |
| CyberRealistic_V9-q8_0, 30steps, 5cfg | cuda | 512x768 | 20.99s | **20.01s** | 1220.07 MB | **1218.29 MB** |
| CyberRealistic_V9_FP16, 30steps, 5cfg | cuda | 768x1024 | 69.43s | **66.23s** | 4736.59 MB | **4734.80 MB** |
| CyberRealistic_V9-q8_0, 30steps, 5cfg | cuda | 768x1024 | 69.66s | **66.77s** | 4736.59 MB | **4734.80 MB** |

<details><summary>fattn log</summary>

```
[DEBUG] ggml_extend.hpp:845  - attention_ext L_q:6144 L_k:6144 n_head:8 C:320 d_head:40 N:1
[DEBUG] ggml_extend.hpp:845  - attention_ext L_q:6144 L_k:77 n_head:8 C:320 d_head:40 N:1
[DEBUG] ggml_extend.hpp:845  - attention_ext L_q:6144 L_k:6144 n_head:8 C:320 d_head:40 N:1
[DEBUG] ggml_extend.hpp:845  - attention_ext L_q:6144 L_k:77 n_head:8 C:320 d_head:40 N:1
[DEBUG] ggml_extend.hpp:845  - attention_ext L_q:1536 L_k:1536 n_head:8 C:640 d_head:80 N:1
[DEBUG] ggml_extend.hpp:883  -  uses flash attention
[DEBUG] ggml_extend.hpp:845  - attention_ext L_q:1536 L_k:77 n_head:8 C:640 d_head:80 N:1
[DEBUG] ggml_extend.hpp:883  -  uses flash attention
[DEBUG] ggml_extend.hpp:885  -  padding k and v dim1 by 179
[DEBUG] ggml_extend.hpp:845  - attention_ext L_q:1536 L_k:1536 n_head:8 C:640 d_head:80 N:1
[...]

PR #774, opened 2025-08-23
revert the L_k padding, introduced as part of #736

### LINKED ISSUE #756: img2img 'weakened' by diffusion-fa
Since ab835f7 , img2img generations need higher strength values to apply changes to the input images.

This is an SDXL test on Vulkan, with the prompt `./sd -m ./cyberrealisticPony_semiRealV35a.safetensors -s 1 -p 'silver car' -i ./base.png `, leaving the strength at the default (0.75) for the first gens:

| base image | 1896b28 | 1896b28<br/>--diffusion-fa | ab835f7 | ab835f7<br/>--diffusion-fa | ab835f7<br/>--diffusion-fa<br/>--strength 0.85 |
| :---: | :---: | :---: | :---: | :---: | :---: |
| base | Image | Image | Image | Image | Image |

I first noticed this issue on current master 5900ef6 . All binaries built with current ggml 7dee1d6 .

@Green-Sky

Green-Sky:
fixes #756

### DIFF
--- a/ggml_extend.hpp
+++ b/ggml_extend.hpp
@@ -875,9 +875,10 @@ __STATIC_INLINE__ struct ggml_tensor* ggml_nn_attention_ext(struct ggml_context*
                                                 d_head == 112 ||
                                                 d_head == 128 ||
                                                 d_head == 256);
-#if 0
+#if 1
     can_use_flash_attn      = can_use_flash_attn && L_k % 256 == 0;
 #else
+	// this causes issues, see https://github.com/leejet/stable-diffusion.cpp/issues/756
     if (can_use_flash_attn && L_k % 256 != 0) {
         // TODO(Green-Sky): might be worth just padding by default
         if (L_k == 77 || L_k == 4208 || L_k == 3952) {


leejet:
For the WAN model, flash attention is almost essential; otherwise, the speed will be very slow. So in this PR https://github.com/leejet/stable-diffusion.cpp/pull/778, I removed many restrictions of the flash attention. It is basically controlled entirely by parameters. As for the possible side effects, this is for the user to weigh and decide.

wbruna:
> As for the possible side effects, this is for the user to weigh and decide.

@leejet , so we'll need to add an extra runtime flag and command-line parameter to disable the kv_pad? Fine by me, but isn't that a little bit cluttered for the interface? (note that FA seems to work fine for image generation otherwise).

I could implement that flag, but it'll necessarily conflict with the Wan PR; and since right now it just crashes for image generation, I won't be able to test it. Wouldn't it be better to apply this PR for now, and I'll figure out how to not break it again inside the Wan branch?

Green-Sky:
No, not before. We should think of a solution **on top of** the wan pr, since it touches that code.

leejet:
I'm not sure if this will be of any help.

```diff
diff --git a/ggml_extend.hpp b/ggml_extend.hpp
index 110bbbc..f583869 100644
--- a/ggml_extend.hpp
+++ b/ggml_extend.hpp
@@ -1013,6 +1013,9 @@ __STATIC_INLINE__ struct ggml_tensor* ggml_nn_attention_ext(struct ggml_context*
     if (flash_attn) {
         // LOG_DEBUG("attention_ext L_q:%d L_k:%d n_head:%d C:%d d_head:%d N:%d", L_q, L_k, n_head, C, d_head, N);
         bool can_use_flash_attn = true;
+        if (L_k < 256) {
+            can_use_flash_attn = false;
+        }
         if (can_use_flash_attn && L_k % 256 != 0) {
             kv_pad = GGML_PAD(L_k, 256) - L_k;
         }
```

wbruna:
That would turn flash attention off for the length-77 tensors that seemed to be the problem, yes.

But... If I may get _way_ over my head here, would something like this make sense instead?

```diff
diff --git a/ggml_extend.hpp b/ggml_extend.hpp
index 110bbbc..d5721d5 100644
--- a/ggml_extend.hpp
+++ b/ggml_extend.hpp
@@ -1031,12 +1031,13 @@ __STATIC_INLINE__ struct ggml_tensor* ggml_nn_attention_ext(struct ggml_context*
     ggml_tensor* kqv = nullptr;
     if (flash_attn) {
         // LOG_DEBUG(" uses flash attention");
         if (kv_pad != 0) {
             // LOG_DEBUG(" padding k and v dim1 by %d", kv_pad);
             k = ggml_pad(ctx, k, 0, kv_pad, 0, 0);
+            scale *= sqrtf((float)(L_k + kv_pad) / (float)L_k);
         }
         k = ggml_cast(ctx, k, GGML_TYPE_F16);
 
         v = ggml_nn_cont(ctx, ggml_permute(ctx, v, 0, 2, 1, 3));  // [N, n_head, L_k, d_head]
         v = ggml_reshape_3d(ctx, v, d_head, L_k, n_head * N);     // [N * n_head, L_k, d_head]
         if (kv_pad != 0) {
```

As I (barely) understand it, this would compensate for the attention mechanism being applied over the larger, padded tensor. It does seem to fix #756 for me, although I can't really say if it could cause other issues.

leejet:
This issue occurred because no mask was applied after kv_pad. I fixed it in this commit https://github.com/leejet/stable-diffusion.cpp/commit/aa5566f0058b39188dec07af256e744df1986b36

PR #776, opened 2025-08-25
Add Nvidia 30 series (cuda arch 86) to build

nekopendev:
This PR adds compute level 8.6 to cuda build.

I tried to add 50 series compute level 12.0, but it seems you need a higher version than cuda 12.6 for it.


leejet:

PR #778, opened 2025-08-29
feat: add wan2.1/2.2 support

leejet:
Feature:
- Wan2.1 T2V 1.3B
- Wan2.1 T2V 14B
- Wan2.1 I2V 14B
- Wan2.2 T2V A14B
- Wan2.2 I2V A14B
- Wan2.2 TI2V 5B

TODO:
- [ ] FLF2V
- [ ] Vace
- [ ] Fun control
- [ ] Reduce the memory usage of WAN VAE

Warning: Currently, only the CUDA and CPU backends support WAN VAE. If you are using another backend, try using `--vae-on-cpu` to run the WAN VAE on the CPU. Although this will be very slow.

## Examples

Since GitHub does not support AVI files, the file I uploaded was converted from AVI to MP4. Some of the flickering is due to the transcoding issue, not the generation problem.

### Wan2.1 T2V 1.3B

```
.\bin\Release\sd.exe -M vid_gen --diffusion-model  ..\..\ComfyUI\models\diffusion_models\wan2.1_t2v_1.3B_fp16.safetensors --vae ..\..\ComfyUI\models\vae\wan_2.1_vae.safetensors --t5xxl ..\..\ComfyUI\models\text_encoders\umt5-xxl-encoder-Q8_0.gguf  -p "a lovely cat" --cfg-scale 6.0 --sampling-method euler -v -n "色调艳丽，过曝，静态，细节模糊不清，字幕，风格，作品，画作，画面，静止，整体发灰，最差质量，低质量，JPEG压缩残留，丑陋的，残缺的，多余的手指，画得不好的手部，画得不好的脸部， 畸形的，毁容的，形态畸形的肢体，手指融合，静止不动的画面，杂乱的背景，三条腿，背景人很多，倒着走" -W 832 -H 480 --diffusion-fa --video-frames 33
```

https://github.com/user-attachments/assets/803b5717-3637-4acd-8541-f57b6c846987

### Wan2.1 T2V 14B

```
[...]

PR #781, opened 2025-08-31
fix sample params initialization

wbruna:
For the wan branch ( #778 ).


Green-Sky:
Memsetting some types is undefined or implementation defined behavior and should be avoided. Since this is kinda c++, wouldn't something like `*sample_params = {};` work better?

wbruna:
It is less of an issue for these parameter structs, though, since they are supposed to be C-compatible. But I fully agree type-safe initializations would be better; I just followed the way the other init functions are doing it.

Edit: replaced the memsets, and rebased onto master.

leejet:

PR #782, opened 2025-08-31
feat: throttle model loading progress updates

wbruna:
Some terminals have slow display latency, so frequent output during model loading can actually slow down the process. This was noticed with Koboldcpp:
https://github.com/LostRuins/koboldcpp/commit/7e1289ade8#diff-7a80f0aa6edc2ee454dc0ef3cd77423a3d7ed7dabe346a923979e58a54440b7b

Also, since tensor loading times can vary a lot, the progress display now shows the average across past iterations instead of just the last one.


leejet (approved):

leejet:

Issue #802, opened 2025-09-07
CUDA: flashattn head dim issues


Green-Sky:
> Removing this d_head check causes sd1 with fa to crash, since cuda does not support 40. 

 _Originally posted by @Green-Sky in [cb1d975](https://github.com/leejet/stable-diffusion.cpp/commit/cb1d975e96bbcd0ec65df9a11331224960cf17fd#r165311448)_

Green-Sky:
Maybe we need to add an extra `ggml_backend_supports_op()` and see if we have to fall back...

leejet:
I think https://github.com/leejet/stable-diffusion.cpp/pull/803 should fix the problem.

Green-Sky:
@JohannesGaessler vulkan gained support for `40` and `160` with https://github.com/ggml-org/llama.cpp/pull/15537 and it might be possible compiling the code paths for those k dims might just work without much code changes.

JohannesGaessler:
It's not possible. As of right now the CUDA FA kernels have a minimum granularity of 16 for the head size, 64 for GPUs without tensor cores.

Green-Sky:
I see, too bad.

PR #803, opened 2025-09-07
add flash attn support check

leejet:
fix https://github.com/leejet/stable-diffusion.cpp/issues/802


Green-Sky:
Will test in a sec.

update1: works with cuda again. (6 images in 523.34s, similar to before)
update2: vulkan still works. (6 images in 205.99s, slower but I also started it heat soaked)
update3: reran vulkan with a little cooled down system (6 images in 183.43s)

Green-Sky:
I more and more wish the whole architecture was more llama.cpp like.

PR #809, opened 2025-09-09
fix: accept NULL in sd_img_gen_params_t::input_id_images_path

wbruna:
It's the value assigned by sd_img_gen_params_init, so generate_image should accept it.


leejet:

PR #812, opened 2025-09-10
chore: remove sd3 flash attention warn

leejet:
It looks like Flash Attention can now be used in SD3.

```bash
 .\bin\Release\sd.exe -m ..\models\sd3_medium_incl_clips_t5xxlfp16.safetensors -p "a lovely cat" --cfg-scale 4.5 --sampling-method euler -v --clip-on-cpu --diffusion-fa
```
output

### DIFF
--- a/stable-diffusion.cpp
+++ b/stable-diffusion.cpp
@@ -344,9 +344,6 @@ public:
                 LOG_INFO("Using flash attention in the diffusion model");
             }
             if (sd_version_is_sd3(version)) {
-                if (sd_ctx_params->diffusion_flash_attn) {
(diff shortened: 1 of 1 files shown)


wbruna:
It works in the sense that it doesn't crash or mess up the image, but I kept the warning because it doesn't do anything either: the `diffusion_flash_attn` parameter isn't even being forwarded to the diffusion model.

Green-Sky:
Right, I think I did not wire up sd3.

leejet:
Check this PR https://github.com/leejet/stable-diffusion.cpp/pull/815

PR #821, opened 2025-09-13
fix: make tiled VAE reuse the compute buffer

wbruna:
The compute buffer was being freed and allocated again between each tile.

Note that both the encode and decode paths already call `free_compute_buffer()` explicitly right after the `compute()` calls.


leejet:

PR #827, opened 2025-09-14
simplify the logic of pm id image loading

leejet:


wbruna:
@leejet , the help text mentions the `normalize_input` flag is part of Photomaker functionality (and in that case it should probably be moved inside the `sd_pm_params_t` struct). But the `generate_image_internal` function doesn't seem to do anything with it?

leejet:
For the original version of Photomaker, normalization is necessary. The `normalize_input` parameter will be removed later.

PR #830, opened 2025-09-14
fix: do not accept LoRA paths outside --lora-model-dir

wbruna:
Attempts to block LoRA names in the prompt from accessing files outside the specified LoRA model directory, which could be a security concern.

Related discussion in: #820


wbruna:
Obsolete; see #1156

PR #833, opened 2025-09-16
fix: move tiling cacl and debug print into the tiling code branch

Green-Sky:
Got confused why it was printing vae tiling sizes after I updated.


Green-Sky:
@stduhpf please take a look at this.

Green-Sky:
force pushed, there where 2 places.

wbruna:
Looks good to me (I was actually about to do that same change 🙂).

BTW, should we apply the tiling size customization for TAESD too? Depending on the image, changing the tile size can sometimes help to avoid tiling artifacts.

Green-Sky:
> Looks good to me (I was actually about to do that same change 🙂).

:smirk: 

> BTW, should we apply the tiling size customization for TAESD too? Depending on the image, changing the tile size can sometimes help to avoid tiling artifacts.

Maybe. I am also not sure anyone is actually using tiling with TAESD.

Tiling and TAESD for Wan (TAEW) is gonna be a more interesting topic for the near future.

leejet:

PR #834, opened 2025-09-16
fix: better progress display for second-order samplers

wbruna:
Second-order samplers call the model twice per step, except for the last iteration. Since the progress is updated only for each second call, the last step is never shown.

The timing information is also misleading, since it's displaying the number of full steps, but measuring only the last half. Comparing to a first-order sampler, the progress shows the same timing for each iteration, and the same number of steps, but takes almost twice as long.

So, change the display to show average time per step, which should give a better idea of the expected time until completion, and update the progress display after all model calls.

### DIFF
--- a/stable-diffusion.cpp
+++ b/stable-diffusion.cpp
@@ -1089,11 +1089,12 @@ public:
         }
         struct ggml_tensor* denoised = ggml_dup_tensor(work_ctx, x);
 
+        int64_t t0 = ggml_time_us();
+
         auto denoise = [&](ggml_tensor* input, float sigma, int step) -> ggml_tensor* {
-            if (step == 1) {
+            if (step == 1 || step == -1) {
                 pretty_progress(0, (int)steps, 0);
             }
-            int64_t t0 = ggml_time_us();
 
             std::vector<float> scaling = denoiser->get_scalings(sigma);
             GGML_ASSERT(scaling.size() == 3);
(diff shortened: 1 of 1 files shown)


wbruna:
Since the time always advances, but we have two updates per step, the displayed average will oscillate, especially on the first iterations. We could instead keep updating for full steps, and explicitly include the last one:

```diff
diff --git a/stable-diffusion.cpp b/stable-diffusion.cpp
index 7b03bee..d21c2eb 100644
--- a/stable-diffusion.cpp
+++ b/stable-diffusion.cpp
@@ -1224,7 +1224,7 @@ public:
                 vec_denoised[i] = latent_result * c_out + vec_input[i] * c_skip;
             }
             int64_t t1 = ggml_time_us();
-            if (step != 0) {
+            if (step > 0 || step == -(int)steps) {
                 int showstep = std::abs(step);
                 pretty_progress(showstep, (int)steps, (t1 - t0) / 1000000.f / showstep);
                 // LOG_INFO("step %d sampling completed taking %.2fs", step, (t1 - t0) * 1.0f / 1000000);
```

stduhpf:
image
I'm not 100% certain it's because of this PR yet, but after updating sd.cpp, I noticed this issue with progress display. (this is with euler sampler) Instead of displaying the speed it seem to just display the time elapsed since the start.

leejet:
I wasn’t able to reproduce your issue. Could you provide the detailed command-line arguments?
```
[DEBUG] ggml_extend.hpp:1582 - Wan2.x-T2V-14B compute buffer size: 202.72 MB(VRAM)
  |==================================================| 14/14 - 1.57it/s
[INFO ] stable-diffusion.cpp:3008 - sampling completed, taking 8.90s
```

stduhpf:
@leejet nevermind, I had wrongly resolved conflicts when merging unrelated changes, and the progress was getting overwritten with the wrong value.

PR #836, opened 2025-09-16
fix workflow

leejet:


.github/workflows/build.yml
```diff
@@ -288,6 +288,11 @@ jobs:
       - windows-latest-cmake
 
     steps:
+      - name: Clone
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
```
Green-Sky, line 294: does that work with `rev-list --count` ?

Green-Sky: ah i am stupid, 0 means all.

This trips me up every time...

Green-Sky: v3 is pretty old though

leejet: Yes, but we just need it to work.

PR #790, opened 2025-09-06
Optimize tensor loading

rmatif:
Following the need here https://github.com/leejet/stable-diffusion.cpp/issues/772 and the discussion here https://github.com/leejet/stable-diffusion.cpp/discussions/789 

It acheives up to x3 faster loading on SDXL model

The PR introduce parallelization on the entire tensor processing and loading pipeline. Tensor preprocessing and deduplication are now distributed across a thread pool, using thread-local maps followed by a final merge to minimize contention. The core loading loop leverages an atomic counter to dispatch tensors to worker threads, each with its own file handle to enable true concurrent I/O on non-zip archives (this operation is not thread-safe on zip files), this will overlaps I/O and cpu works

cc @wbruna

PR #839, opened 2025-09-16
fix: avoid crash on multithreaded model conversion

wbruna:
Since 55c2e05d98752fa7236c18a7352ec421f5e345ab / #790  , model conversion easily crashes, unless `--threads 1` is used.

This mutex _seems_ to fix the issue for me, but I didn't look for other possible race conditions.


wbruna:
I believe this was superseded by 567f9f14f0f61c4ca436f45bd86c47d260586448 .

Issue #837, opened 2025-09-16
clip related crash, segmentation fault


kuesji:
it seems like 48956ffb874f9a2c1306b721a6896cac0a2a3256 introduced a bug and it causes crash on my laptop.
i tried to compile latest release but it was crashing. i was able the trace the issue back to that commit.

let me know if anything else is needed. i'm going to stay at ddc4a18b92dc58b2490d6528f7118a9a166c5d69 for now.

<details> 
<summary> output of execution with -v arg </summary>

> Option: 
    n_threads:                         8
    mode:                              img_gen
    model_path:                        /data/kuesji/archive/models/image_generation/models/sdxl_illustrious:hassakuxl:v2.2.gguf
    wtype:                             unspecified
    clip_l_path:                       
    clip_g_path:                       
    clip_vision_path:                  
    t5xxl_path:                        
    diffusion_model_path:              
    high_noise_diffusion_model_path:   
    vae_path:                          
    taesd_path:                        
    esrgan_path:                       
    control_net_path:                  
    embedding_dir:                     
    stacked_id_embed_dir:              
    input_id_images_path:              
    style ratio:                       20.00
    normalize input image:             false
    output_path:                       output.png
    init_image_path:                   
    end_image_path:                    
    mask_image_path:                   
    control_image_path:                
    ref_images_paths:
    increase_ref_index:                false
    offload_params_to_cpu:             false
    clip_on_cpu:                       false
    control_net_cpu:                   false
    vae_on_cpu:                        true
    diffusion flash attention:         false
    diffusion Conv2d direct:           false
    vae_conv_direct:                   false
    control_strength:                  0.90
    prompt:                            purple cat
    negative_prompt:                   
    clip_skip:                         -1
    width:                             576
    height:                            1024
    sample_params:                     (txt_cfg: 3.50, img_cfg: 3.50, distilled_guidance: 3.50, slg.layer_count: 3, slg.layer_start: 0.01, slg.layer_end: 0.20, slg.scale: 0.00, scheduler: karras, sample_method: euler_a, sample_steps: 8, eta: 0.00)
    high_noise_sample_params:          (txt_cfg: 7.00, img_cfg: 7.00, distilled_guidance: 3.50, slg.layer_count: 3, slg.layer_start: 0.01, slg.layer_end: 0.20, slg.scale: 0.00, scheduler: default, sample_method: euler_a, sample_steps: -1, eta: 0.00)
    moe_boundary:                      0.875
[...]

PR #840, opened 2025-09-17
do not quantize embedding weight

leejet:
Try to fix https://github.com/leejet/stable-diffusion.cpp/issues/837.

### DIFF
--- a/clip.hpp
+++ b/clip.hpp
@@ -553,12 +553,13 @@ protected:
     void init_params(struct ggml_context* ctx, const String2GGMLType& tensor_types = {}, const std::string prefix = "") {
         enum ggml_type token_wtype = GGML_TYPE_F32;
         if (!force_clip_f32) {
-            auto tensor_type = tensor_types.find(prefix + "token_embedding.weight");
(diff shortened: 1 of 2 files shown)


clip.hpp
```diff
@@ -554,11 +554,11 @@ class CLIPEmbeddings : public GGMLBlock {
         enum ggml_type token_wtype = GGML_TYPE_F32;
         if (!force_clip_f32) {
             auto tensor_type = tensor_types.find(prefix + "token_embedding.weight");
-            if (tensor_type != tensor_types.end())
+            if (tensor_type != tensor_types.end() && tensor_type->second == GGML_TYPE_F16) {
```
wbruna: Maybe lower quants could be kept at F16 too (or even Q8)? The main reason for quantization is reducing memory usage, so it'd be surprising if a Q8_0 or Q4_0 quant ended up using more VRAM than F16.

leejet:
I added support for more quant types. The current implementation is a bit hardcoded, but this should work safely for now. We can look into a more elegant solution later.

PR #843, opened 2025-09-19
HipBLAS / ROCm build instruction fix

Stefan-Olt:
Build instructions for ROCm were broken with automatic GPU detection (wrong variable set). Additionally I added some information where this automatic detection won't work, for example with not officially supported cards.

I also added `-DCMAKE_POSITION_INDEPENDENT_CODE=ON` because without it I get linking errors (tried on two different 2 systems), it also shouldn't hurt on systems where it is not needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,13 +125,14 @@ cmake --build . --config Release
 
 ##### Using HipBLAS
 This provides BLAS acceleration using the ROCm cores of your AMD GPU. Make sure to have the ROCm toolkit installed.
+To build for another GPU architecture than installed in your system, set `$GFX_NAME` manually to the desired architecture. This is also necessary if your GPU is not officially supported by ROCm, for example set `$GFX_NAME` to `gfx1030` for consumer RDNA2 cards.
 
 Windows User Refer to [docs/hipBLAS_on_Windows.md](docs%2FhipBLAS_on_Windows.md) for a comprehensive guide.
 
 ```
 export GFX_NAME=$(rocminfo | grep -m 1 -E "gfx[^0]{1}" | sed -e 's/ *Name: *//' | awk '{$1=$1; print}' || echo "rocminfo missing")
 echo $GFX_NAME
-cmake .. -G "Ninja" -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DSD_HIPBLAS=ON -DCMAKE_BUILD_TYPE=Release -DGPU_TARGETS=$GFX_NAME -DCMAKE_BUILD_WITH_INSTALL_RPATH=ON
+cmake .. -G "Ninja" -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DSD_HIPBLAS=ON -DCMAKE_BUILD_TYPE=Release -DAMDGPU_TARGETS=$GFX_NAME -DCMAKE_BUILD_WITH_INSTALL_RPATH=ON -DCMAKE_POSITION_INDEPENDENT_CODE=ON
 cmake --build . --config Release
 ```
 


stduhpf:
Are the linking issues without PIC related to openMP? Because on my end I have to set `GGML_OPENMP` to `OFF` for it to link.

wbruna:
FWIW, the choice of `GPU_TARGETS` versus `AMDGPU_TARGETS`  depends on the ROCm version:

https://rocm.docs.amd.com/en/docs-6.4.0/about/release-notes.html#id26

> The AMDGPU_TARGETS build variable should be replaced with GPU_TARGETS. AMDGPU_TARGETS is deprecated.

wbruna:
> Are the linking issues without PIC related to openMP? Because on my end I have to set `GGML_OPENMP` to `OFF` for it to link.

I also need to build it with PIC (by `-DSD_BUILD_SHARED_LIBS=ON`) to avoid a linking error, but I don't remember seeing openMP mentioned in the build log.

stduhpf:
> FWIW, the choice of `GPU_TARGETS` versus `AMDGPU_TARGETS` depends on the ROCm version:
> 
> https://rocm.docs.amd.com/en/docs-6.4.0/about/release-notes.html#id26
> 
> > The AMDGPU_TARGETS build variable should be replaced with GPU_TARGETS. AMDGPU_TARGETS is deprecated.

So we probably should just set both to the same values?

Stefan-Olt:
llama.cpp only sets `AMDGPU_TARGETS` and I'm using the latest ROCm 7.0.1. But setting both will not hurt

wbruna:
The detection line is kind of broken too:

```sh
$ rocminfo | grep -m 1 -E "gfx[^0]{1}" | sed -e 's/ *Name: *//' | awk '{$1=$1; print}' || echo "rocminfo missing"
gfx1102
$ rocminfo_ | grep -m 1 -E "gfx[^0]{1}" | sed -e 's/ *Name: *//' | awk '{$1=$1; print}' || echo "rocminfo missing"
bash: rocminfo_: command not found
$ 
```

Stefan-Olt:
> The detection line is kind of broken too:
> 
> ```shell
> $ rocminfo | grep -m 1 -E "gfx[^0]{1}" | sed -e 's/ *Name: *//' | awk '{$1=$1; print}' || echo "rocminfo missing"
> gfx1102
> $ rocminfo_ | grep -m 1 -E "gfx[^0]{1}" | sed -e 's/ *Name: *//' | awk '{$1=$1; print}' || echo "rocminfo missing"
> bash: rocminfo_: command not found
> $ 
> ```
From where do you get that underscore?

wbruna:
> From where do you get that underscore?

It's just an invalid command to check if the `echo "rocminfo missing"` really gets executed when the command isn't found (easier than renaming my `rocminfo` binary 🙂).

Stefan-Olt:
Updated instructions to
- Set both variables: AMDGPU_TARGETS and GPU_TARGETS
- Actually check if the rocminfo command exists and print error if not
- Check if the GFX_NAME variable was set to a non-empty value, if not print error, else detected GPU architecture

README.md
```diff
@@ -125,13 +125,14 @@ cmake --build . --config Release
...
+To build for another GPU architecture than installed in your system, set `$GFX_NAME` manually to the desired architecture (replace first command). This is also necessary if your GPU is not officially supported by ROCm, for example you have to set `$GFX_NAME` manually to `gfx1030` for consumer RDNA2 cards.
 
 Windows User Refer to [docs/hipBLAS_on_Windows.md](docs%2FhipBLAS_on_Windows.md) for a comprehensive guide.
 
 ```
-export GFX_NAME=$(rocminfo | grep -m 1 -E "gfx[^0]{1}" | sed -e 's/ *Name: *//' | awk '{$1=$1; print}' || echo "rocminfo missing")
-echo $GFX_NAME
-cmake .. -G "Ninja" -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DSD_HIPBLAS=ON -DCMAKE_BUILD_TYPE=Release -DGPU_TARGETS=$GFX_NAME -DCMAKE_BUILD_WITH_INSTALL_RPATH=ON
+if command -v rocminfo; then export GFX_NAME=$(rocminfo | grep -m 1 -E "gfx[^0]{1}" | sed -e 's/ *Name: *//' | awk '{$1=$1; print}'); else echo "rocminfo missing!"; fi
+if [ -z "${GFX_NAME}" ]; then echo 'Error: Could detect GPU!'; else echo "Building for GPU: ${GFX_NAME}"; fi
```
wbruna: Small typo (Couldn't).

Otherwise, looks good to me. But if you'll change it anyway, that grep+sed+awk can be simplified a bit:

```sh
$ rocminfo | grep -m 1 -E "gfx[^0]{1}" | sed -e 's/ *Name: *//' | awk '{$1=$1; print}'
gfx1102
$ rocminfo | awk '/ *Name: +gfx[1-9]/ {print $2; exit}'
gfx1102
```

Stefan-Olt: Updated, fixed typo and simplified as you proposed

leejet:

Issue #838, opened 2025-09-16
SDLX Black output afer PR 55c2e05


pedroCabrera:
Hi there, i found out after PR 55c2e05 my SDXL output are pure black, it is working fine in 52a97b3
Im runing a RTX4090 with -DSD_CUDA=ON 

The same PR is working fine loading FLUX

**WORKING LOG 52a97b3**
sd.exe -m models\SDXL_TURBO\sd_xl_turbo_1.0.safetensors --vae models\SDXL_TURBO\sdxl-vae-fp16-fix.safetensors -H 1024 -W 1024 -p "a lovely cat" -v
Option:
    n_threads:                         12
    mode:                              img_gen
    model_path:                        models\SDXL_TURBO\sd_xl_turbo_1.0.safetensors
    wtype:                             unspecified
    clip_l_path:
    clip_g_path:
    clip_vision_path:
    t5xxl_path:
    diffusion_model_path:
    high_noise_diffusion_model_path:
    vae_path:                          models\SDXL_TURBO\sdxl-vae-fp16-fix.safetensors
    taesd_path:
    esrgan_path:
    control_net_path:
    embedding_dir:
    stacked_id_embed_dir:
    input_id_images_path:
    style ratio:                       20.00
    normalize input image:             false
    output_path:                       output.png
    init_image_path:
    end_image_path:
    mask_image_path:
    control_image_path:
    ref_images_paths:
    control_video_path:
    increase_ref_index:                false
    offload_params_to_cpu:             false
    clip_on_cpu:                       false
    control_net_cpu:                   false
    vae_on_cpu:                        false
    diffusion flash attention:         false
    diffusion Conv2d direct:           false
    vae_conv_direct:                   false
    control_strength:                  0.90
    prompt:                            a lovely cat
    negative_prompt:
    clip_skip:                         -1
    width:                             1024
    height:                            1024
    sample_params:                     (txt_cfg: 7.00, img_cfg: 7.00, distilled_guidance: 3.50, slg.layer_count: 3, slg.layer_start: 0.01, slg.layer_end: 0.20, slg.scale: 0.00, scheduler: default, sample_method: default, sample_steps: 20, eta: 0.00)
[...]

PR #844, opened 2025-09-19
make tensor loading deterministic

rmatif:
Changes introduced in https://github.com/leejet/stable-diffusion.cpp/pull/790 could fails on some tensors because we deduped by name using an unordered_map, whichever thread flushed last won, so some chunks pointed at the wrong offsets and corrupt the load. This fix keeps the most recent source tensor (highest index) per name, then re-sorts by the original order before reading this makes loads deterministic 

This will fix https://github.com/leejet/stable-diffusion.cpp/issues/838

### DIFF
--- a/model.cpp
+++ b/model.cpp
@@ -1,4 +1,5 @@
 #include <stdarg.h>
+#include <algorithm>
 #include <atomic>
 #include <chrono>
 #include <fstream>
@@ -2006,13 +2007,25 @@ bool ModelLoader::load_tensors(on_new_tensor_cb_t on_new_tensor_cb, int n_thread
             w.join();
         }
 
-        std::unordered_map<std::string, IndexedStorage> latest_map;
+        std::vector<IndexedStorage> deduplicated;
+        deduplicated.reserve(all_results.size());
+        std::unordered_map<std::string, size_t> name_to_pos;
         for (auto& entry : all_results) {
-            latest_map[entry.ts.name] = entry;
+            auto it = name_to_pos.find(entry.ts.name);
+            if (it == name_to_pos.end()) {
+                name_to_pos.emplace(entry.ts.name, deduplicated.size());
+                deduplicated.push_back(entry);
+            } else if (deduplicated[it->second].index < entry.index) {
+                deduplicated[it->second] = entry;
+            }
         }
 
-        processed_tensor_storages.reserve(latest_map.size());
-        for (auto& [name, entry] : latest_map) {
+        std::sort(deduplicated.begin(), deduplicated.end(), [](const IndexedStorage& a, const IndexedStorage& b) {
+            return a.index < b.index;
+        });
+
+        processed_tensor_storages.reserve(deduplicated.size());
+        for (auto& entry : deduplicated) {
             processed_tensor_storages.push_back(entry.ts);
         }
     }


wbruna:
I can't say I'm following the code (this really needs a lot more comments IMHO), but going by your description: if this is happening because of an unstable ordering of the `all_results` vector, wouldn't it be much simpler to replace it with a vector-of-vectors, indexed by thread id?

<details>
<summary>Something like this?</summary>

```diff
diff --git a/model.cpp b/model.cpp
index 2e7127d..78669de 100644
--- a/model.cpp
+++ b/model.cpp
@@ -1968,17 +1968,18 @@ bool ModelLoader::load_tensors(on_new_tensor_cb_t on_new_tensor_cb, int n_thread
         };
 
         std::mutex vec_mutex;
-        std::vector<IndexedStorage> all_results;
+        std::vector<std::vector<IndexedStorage> > all_results;
 
         int n_threads = std::min(num_threads_to_use, (int)tensor_storages.size());
         if (n_threads < 1) {
             n_threads = 1;
         }
         std::vector<std::thread> workers;
+        all_results.resize(n_threads);
 
         for (int i = 0; i < n_threads; ++i) {
             workers.emplace_back([&, thread_id = i]() {
-                std::vector<IndexedStorage> local_results;
+                std::vector<IndexedStorage> & local_results = all_results[i];
                 std::vector<TensorStorage> temp_storages;
 
                 for (size_t j = thread_id; j < tensor_storages.size(); j += n_threads) {
@@ -1995,11 +1996,6 @@ bool ModelLoader::load_tensors(on_new_tensor_cb_t on_new_tensor_cb, int n_thread
                     }
                 }
 
-                if (!local_results.empty()) {
-                    std::lock_guard<std::mutex> lock(vec_mutex);
-                    all_results.insert(all_results.end(),
-                                       local_results.begin(), local_results.end());
-                }
             });
         }
         for (auto& w : workers) {
@@ -2007,8 +2003,10 @@ bool ModelLoader::load_tensors(on_new_tensor_cb_t on_new_tensor_cb, int n_thread
         }
 
         std::unordered_map<std::string, IndexedStorage> latest_map;
-        for (auto& entry : all_results) {
-            latest_map[entry.ts.name] = entry;
+        for (auto& thread_results: all_results) {
+            for (auto& entry : thread_results) {
+                latest_map[entry.ts.name] = entry;
+            }
         }
 
         processed_tensor_storages.reserve(latest_map.size());
```

</details>

Edit: ok, not as-is because each thread isn't processing the entries contiguously. But they could: thread 0 gets the entries `0..(tensor_storages.size()/n_threads)`, thread 1 the next block, and so forth:

```c++
        // how many tensors per thread
        size_t slice = tensor_storages.size() / n_threads;
        // each of the first threads will take one leftover
        size_t leftovers = tensor_storages.size() % n_threads;

        for (int i = 0; i < n_threads; ++i) {
            workers.emplace_back([&, thread_id = i]() {
                std::vector<IndexedStorage> & local_results = all_results[thread_id];
                std::vector<TensorStorage> temp_storages;
                size_t first, last;
                if (thread_id < leftovers) {
                    first = (slice + 1) * thread_id;
                    last = first + slice + 1;
                } else {
                    first = tensor_storages.size() - slice * (n_threads - thread_id);
                    last = first + slice;
                }
                for (size_t j = first; j < last; j++) {
```

(@rmatif , feel free to borrow this if it makes sense; I can't properly test it anyway)

leejet:
@rmatif Could you describe in detail the reason for the changes? Just looking at the code leaves me a bit confused.

leejet:
I think I understand where the problem is. This PR should fix it. Thank you for your contribution.

PR #846, opened 2025-09-20
refactor: simplify weight conversion code

wbruna:


leejet:
The related logic has already been refactored.

PR #852, opened 2025-09-22
Fix dockerfile libgomp1 dependency + improvements

SergeantSerk:
Fixes an issue with stable-diffusion needing `libgomp.so.1` but not present during runtime.

```
/sd: error while loading shared libraries: libgomp.so.1: cannot open shared object file: No such file or directory
```

- Introduces cmake parallel build
- No install recommends for the build stage to prevent any extra packages from being installed in the build stage

Build time (on my machine, no caching) `docker build --no-cache -t sd .`:
- Without improvements: **96.0s**
- With improvements: **59.4s**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,9 +2,7 @@ ARG UBUNTU_VERSION=22.04
 
 FROM ubuntu:$UBUNTU_VERSION AS build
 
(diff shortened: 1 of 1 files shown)


leejet:

PR #854, opened 2025-09-24
multithread tensor loading fixes

wbruna:
While debugging #838  , I noticed that some models were being loaded with the maximum number of threads regardless of the `--threads` parameter, so I adjusted those calls to pass the thread number. For textual inversions, I kept the processing threads at 1, since the thread spawning was giving me higher latency than just processing those smaller files directly.

Edit: removed the fix for #838


leejet:
I think the fix in https://github.com/leejet/stable-diffusion.cpp/pull/844 is a bit clearer, so I chose to merge https://github.com/leejet/stable-diffusion.cpp/pull/844

wbruna:
> I think the fix in #844 is a bit clearer, so I chose to merge #844

Alright. I'll rebase this, keeping only the other fixes, then.

leejet:

PR #855, opened 2025-09-24
Add Support more esrgan models & x2 & x1 models

pedroCabrera:
Hi there, i have modified the ESRGAN code to support more models, i tested with a bunch from https://openmodeldb.info/ and some work out of the box but some of them had to be converted with :

```
import torch
checkpoint = torch.load(r"model.pth",map_location="cpu")
torch.save(checkpoint,r"model_converted.pth",_use_new_zipfile_serialization=True)

```
I have also tested with x2 and x1 models from that web and some of them works  ( 2x-ESRGAN.pth ,  1x-Anti-Aliasing.pth )
It also supports RealESRGAN_x4plus.pth as well as the original RealESRGAN_x4plus_anime_6B.pth

PD: I'm integrating stable-diffusion.cpp into [Nuke](https://www.foundry.com/products/nuke-family/nuke) wich is a vfx software , my repo is here -> [pc_nuke_diffusion](https://github.com/pedroCabrera/pc_nuke_diffusion), still early development but most of the cli options are avaliable there already.

esrgan_test

### DIFF
--- a/esrgan.hpp
+++ b/esrgan.hpp
@@ -137,10 +137,10 @@ public:
         // upsample
         if (scale >= 2) {
             auto conv_up1 = std::dynamic_pointer_cast<Conv2d>(blocks["conv_up1"]);
-            feat = lrelu(ctx, conv_up1->forward(ctx, ggml_upscale(ctx, feat, 2, GGML_SCALE_MODE_NEAREST)));
+            feat          = lrelu(ctx, conv_up1->forward(ctx, ggml_upscale(ctx, feat, 2, GGML_SCALE_MODE_NEAREST)));
             if (scale == 4) {
                 auto conv_up2 = std::dynamic_pointer_cast<Conv2d>(blocks["conv_up2"]);
-                feat = lrelu(ctx, conv_up2->forward(ctx, ggml_upscale(ctx, feat, 2, GGML_SCALE_MODE_NEAREST)));
+                feat          = lrelu(ctx, conv_up2->forward(ctx, ggml_upscale(ctx, feat, 2, GGML_SCALE_MODE_NEAREST)));
             }
         }
         // for all scales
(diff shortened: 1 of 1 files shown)


esrgan.hpp
```diff
@@ -164,34 +177,185 @@ struct ESRGAN : public GGMLRunner {
         return "esrgan";
     }
 
-    bool load_from_file(const std::string& file_path, int n_threads) {
+    bool load_from_file(const std::string& file_path) {
```
fszontagh: This change causes the build to fail with the following error:

`upscaler.cpp:58:45: error: no matching function for call to ‘ESRGAN::load_from_file(const std::string&, int&)’
   58 |         if (!esrgan_upscaler->load_from_file(esrgan_path, n_threads)) {
      |              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~
`

fszontagh:
A simple python script to _convert_ an old type .pth: 
`convert_pth.py old_format.pth new_format_pth.pth`
https://gist.github.com/fszontagh/4b1765d3ec75e1200f7a17ad8969da8b

Original image (512x768)
0ad0e5d9-7853-49d8-a5d9-3a00d333ef85_resized_512x768_refined_upscale_test_current_sd_normal

Upscaled image with the converted [4x_Nickelback_70000G.pth](https://openmodeldb.info/models/4x-Nickelback) (2048x3072)

0ad0e5d9-7853-49d8-a5d9-3a00d333ef85_resized_512x768_refined_upscale_test

This is upscaled with RealESRGAN_x4plus.pth ([which is worked earlier](https://github.com/leejet/stable-diffusion.cpp/blob/master/docs/esrgan.md) ) (2048x3072)
0ad0e5d9-7853-49d8-a5d9-3a00d333ef85_resized_512x768_refined

esrgan.hpp
pedroCabrera: yep, its fixed now, i was working in an older version without the multithread feature, now its fixed, thanks!

wbruna:
#865 can be used to test the upscaler models with external images.

leejet:

PR #861, opened 2025-09-27
canny preprocessor fixes

wbruna:


leejet:
It looks like this just increases the size of `mem_size`?

wbruna:
The struct is also received by value, so freeing the data buffer leaves behind an invalid pointer; so I'm reusing the existing buffer instead..

leejet:
So it’s just to avoid reallocating memory, and it won’t affect the result?

wbruna:
> So it’s just to avoid reallocating memory, and it won’t affect the result?

It's more like 'avoid changing the data pointer', since we can't alter its value for the code layers above (it's a C interface). It shouldn't affect the canny processing result in any way: the input image is only used to initialize the tensors, so it's safe to overwrite it at that later point.

(of course, on current master, the data pointer ends up pointing to a freed buffer, so the image data could have been changed (corrupted) by the program elsewhere. In that sense, image results could be different.)

leejet:
I forgot that the parameter is passed by value, so it makes sense.

leejet:

PR #851, opened 2025-09-22
add Qwen Image support

leejet:
## txt2img

```shell
.\bin\Release\sd.exe --diffusion-model  ..\..\ComfyUI\models\diffusion_models\qwen-image-Q8_0.gguf --vae ..\..\ComfyUI\models\vae\qwen_image_vae.safetensors  --qwen2vl ..\..\ComfyUI\models\text_encoders\Qwen2.5-VL-7B-Instruct-Q8_0.gguf  -p '一个穿着"QWEN"标志的T恤的中国美女正拿着黑色的马克笔面相镜头微笑。她身后的玻璃板上手写体写着 “一、Qwen-Image的技术路线： 探索视觉生成基础模型的极限，开创理解与生成一体化的未来。二、Qwen-Image的模型特色：1、复杂文字渲染。支持中英渲染、自动布局； 2、精准图像编辑。支持文字编辑、物体增减、风格变换。三、Qwen-Image的未来愿景：赋能专业内容创作、助力生成式AI发展。”' --cfg-scale 2.5 --sampling-method euler -v --offload-to-cpu -H 1024 -W 1024 --diffusion-fa --flow-shift 3
```

qwen_image_t2i

## img2img

```shell
.\bin\Release\sd.exe --diffusion-model  ..\..\ComfyUI\models\diffusion_models\qwen-image-Q8_0.gguf --vae ..\..\ComfyUI\models\vae\qwen_image_vae.safetensors  --qwen2vl ..\..\ComfyUI\models\text_encoders\Qwen2.5-VL-7B-Instruct-Q8_0.gguf --cfg-scale 2.5 --sampling-method euler -v --offload-to-cpu --diffusion-fa --flow-shift 3 -i ..\assets\flux\flux1-dev-q8_0.png -p "a lovely cat"
```

qwen_image_i2i

## Qwen Image Edit

https://github.com/leejet/stable-diffusion.cpp/pull/877

PR #873, opened 2025-10-04
fix VAE tiling for Qwen Image

wbruna:
Intended for #851 , ~~but works fine on current master~~. Edit: encoding path fix applied on top of `qwen_image`.

### DIFF
--- a/stable-diffusion.cpp
+++ b/stable-diffusion.cpp
@@ -1440,18 +1440,19 @@ public:
         if (vae_tiling_params.enabled && !encode_video) {
             // TODO wan2.2 vae support?
             int C = sd_version_is_dit(version) ? 16 : 4;
-            int NE2, NE3;
+            int ne2;
+            int ne3;
             if (sd_version_is_qwen_image(version)) {
-                NE2 = x->ne[3];
-                NE3 = C;
+                ne2 = 1;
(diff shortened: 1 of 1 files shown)


Green-Sky:
Did you try if this works with wan?

LostRuins:
It will not trigger with WAN in it's current state, because `encode_video` will be true, so it will never enter the vae tiling codepath

LostRuins:
@wbruna i can confirm it works for qwen image, and there's no regression with WAN (still works fine with no tiling)

wbruna:
It was failing simply because the split+merge assumed a specific tensor shape (and blowing up deep inside ggml because the wrong shapes triggered a division by zero). But for Wan, I'm sure leejet would have spotted this bug if it was just a matter of fixing the dimensions... Feel free to try it, though; I'm not sure I have either hardware or patience enough for that 😅.

wbruna:
Not enough for img2img, unfortunately; any encode tiling triggers an assertion:
> ggml/src/ggml-backend.cpp:300: GGML_ASSERT(offset + size <= ggml_nbytes(tensor) && "tensor read out of bounds") failed

Edit: fixed it, but needed an explicit test for Qwen Image.

leejet:

PR #881, opened 2025-10-12
Avoid generating black images when running T5 on the GPU

leejet:


wbruna:
Working on ROCm with Chroma:

t5_1760270423

PR #890, opened 2025-10-15
feat: in-place SDXL VAE conv scaling

wbruna:
No idea if this is really OK, since the resulting images look a little bit different in details, and the memory savings look surprisingly high, but this is for a 1024x1024 image, no `vae-conv-direct`:

before:
```
[INFO ] stable-diffusion.cpp:2386 - decoding 1 latents
[DEBUG] ggml_extend.hpp:1579 - vae compute buffer size: 7680.25 MB(VRAM)
[DEBUG] stable-diffusion.cpp:1677 - computing vae decode graph completed, taking 5.98s
[INFO ] stable-diffusion.cpp:2396 - latent 1 decoded, taking 5.98s
[INFO ] stable-diffusion.cpp:2400 - decode_first_stage completed, taking 5.98s
```

after:
```
[INFO ] stable-diffusion.cpp:2386 - decoding 1 latents
[DEBUG] ggml_extend.hpp:1579 - vae compute buffer size: 6656.25 MB(VRAM)
[DEBUG] stable-diffusion.cpp:1677 - computing vae decode graph completed, taking 5.99s
[INFO ] stable-diffusion.cpp:2396 - latent 1 decoded, taking 6.00s
[INFO ] stable-diffusion.cpp:2400 - decode_first_stage completed, taking 6.00s
```

### DIFF
--- a/ggml_extend.hpp
+++ b/ggml_extend.hpp
@@ -984,7 +984,7 @@ __STATIC_INLINE__ struct ggml_tensor* ggml_nn_conv_2d(struct ggml_context* ctx,
                                                       bool direct = false,
                                                       float scale = 1.f) {
     if (scale != 1.f) {
-        x = ggml_scale(ctx, x, scale);
+        x = ggml_scale_inplace(ctx, x, scale);
     }
     if (direct) {
         x = ggml_conv_2d_direct(ctx, w, x, s0, s1, p0, p1, d0, d1);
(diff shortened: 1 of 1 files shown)


leejet:
Performing an in-place scale on the input x of ggml_nn_conv_2d is not a good idea, because x might also be used as the input for other operations. This is also the reason why the image changes after applying in-place scaling.

wbruna:
> Performing an in-place scale on the input x of ggml_nn_conv_2d is not a good idea, because x might also be used as the input for other operations. This is also the reason why the image changes after applying in-place scaling.

Oh, I see. But that wouldn't be an issue for the result of the conv_2d call, right?

I'll test it again with this change.

leejet:
Yes, but it will affect other operations using x, and ultimately it will impact the generated image.

wbruna:
I tested changing only the second `ggml_scale`, and although the generated image didn't change anymore, I didn't notice any memory improvements either.

PR #896, opened 2025-10-16
Support Custom ESRGAN tile size

pedroCabrera:
Add suport for custom tile size in esrgan models ( this allows for faster inference on graphics card with good amount of memory )


wbruna:
Working fine on a few tests. Increasing the tile size didn't really help my card, even though it has 16G VRAM, but _lowering_ it to 64 made it almost twice as fast.

@pedroCabrera , the recent command-line flag refactoring caused conflicts; could you take a look?

I'd also  suggest changing the command-line flag from `--upscale-tile` to `--upscale-tile-size`, to better match `--vae-tile-size`. That would also make it a bit more future-proof, in case we add support for changing the overlapping factor, non-square tiles, etc.

pedroCabrera:
Thanks for checking, I have updated the param to --upscale-tile-size and rebased with latest changes so it should be mergeable i think!

PR #902, opened 2025-10-18
chore: add .clang-tidy configuration and apply modernize checks

leejet:


stable-diffusion.h
```diff
@@ -23,10 +23,9 @@
 extern "C" {
 #endif
 
-#include <stdbool.h>
-#include <stddef.h>
-#include <stdint.h>
-#include <string.h>
+#include <cstddef>
+#include <cstdint>
+#include <cstring>
```
wbruna: If stable-diffusion.h is intended to be compatible with C, it can't include C++-only headers.

leejet: fixed.

Issue #528, opened 2024-12-23
Output image not saved if directory does not exists


olivbrau:
I think that if we use -o option with a path that does not exist yet, nothing is saved.
And the log says :   save result image to xxx  without error message.

I think it would be nice if sd checked the existence of the directory and created it if necessary before saving the image.

akleine:
I have just added  PR #912  to solve this.
Unfortunately the library used **does not give any detailed information about possible errors**.
In fact stb_image_write.h (line 77) tells: "Each function returns 0 on failure and non-0 on success."

I think :       "created it if necessary before saving the image...."    
... seems to be  impossible.  Consider there are many other reasons for failures,   for example missing  R/W  access to the wanted path.

PR #912, opened 2025-10-23
Add an indication of whether the PNG/JPG image saving was successful …

akleine:
…or not,  also  see issue #528. 
Unfortunately the library used does not give any detailed information about possible errors. 
In fact  stb_image_write.h   (line 77)    tells:   "Each function returns 0 on failure and non-0 on success."


akleine:
Hi @leejet, thank you for your work and for merging all that PR of last weeks.

BTW  I think now it is a good time to close this old issue in  https://github.com/leejet/stable-diffusion.cpp/issues/528

PR #915, opened 2025-10-24
add support for scaled fp8 tensors

tsite:
commit 1: add support for using bf16 as a native type - involves a refactor to the type parsing and conversion logic
commit 2: add support for scaled fp8 tensors
commit 3: convert bf16 to fp32 during loading

### DIFF
--- a/model.cpp
+++ b/model.cpp
@@ -746,6 +746,10 @@ std::string convert_tensor_name(std::string name) {
 
 void add_preprocess_tensor_storage_types(String2GGMLType& tensor_storages_types, std::string name, enum ggml_type type) {
     std::string new_name = convert_tensor_name(name);
+    // TODO: enable bf16 once it is supported by the relevant ggml ops
+    if (type == GGML_TYPE_BF16) {
+        type = GGML_TYPE_F32;
+    }
 
     if (new_name.find("cond_stage_model") != std::string::npos && ends_with(new_name, "attn.in_proj_weight")) {
         size_t prefix_size                                        = new_name.find("attn.in_proj_weight");


wbruna:
> remove dummy imatrix from ggml_quantize_chunk call - I don't think it needs it

Actually, it  does: quality degrades much more without it. Try e.g. any SDXL model at q5_0.

tsite:
> > remove dummy imatrix from ggml_quantize_chunk call - I don't think it needs it
> 
> Actually, it does: quality degrades much more without it. Try e.g. any SDXL model at q5_0.

interesting, good to know - I added back the dummy imatrix

leejet:
It seems that you have removed some code related to model conversion, such as f64 → f32. This can cause issues when loading certain models. I suggest that if you don’t fully understand the reason behind some parts of the code, you shouldn’t modify them. Instead, you should only implement the parts that you do understand.

tsite:
> It seems that you have removed some code related to model conversion, such as f64 → f32. This can cause issues when loading certain models. I suggest that if you don’t fully understand the reason behind some parts of the code, you shouldn’t modify them. Instead, you should only implement the parts that you do understand.

I think the convert_tensor function should handle that - there's a case added that checks for the GGML_TYPE_F64 source type and converts it to GGML_TYPE_F32 when necessary. If you have a model in mind that you think this change may break, I can test it out to make sure that it works properly. Imo it no longer makes sense to use hacky sd types now that ggml has added support for f64, bf16, etc, but if you have other reasons for not using the native ggml types, I'm all ears.

leejet:
Most of ggml's ops do not support f64/i64/bf16, which will cause issues. You can use this model for testing: https://civitai.com/models/7371/rev-animated. This model contains f64, and your changes will cause problems with it.

Green-Sky:
> f8_e5m2 now autoconvert to f32 for less precision loss

Why, f16 is e5m10. This should be lossless.

stduhpf:
I think it's a better practice to avoid including too many unrelated changes like that in one PR. 

This makes it harder to review, if some changes are bad, the whole PR can't be merged, and it also has a higher chance of breaking many other pending PRs.

tsite:
> > f8_e5m2 now autoconvert to f32 for less precision loss
> 
> Why, f16 is e5m10. This should be lossless.

scaled f8_e5m2 tensors are multiplied by a float32 scaling factor

tsite:
> Most of ggml's ops do not support f64/i64/bf16, which will cause issues. You can use this model for testing: https://civitai.com/models/7371/rev-animated. This model contains f64, and your changes will cause problems with it.

Good to know, thanks! I took a closer look at the ggml library and you're right that the f64/i64 types are missing kernels. I think bf16 has full support though for all the ops, as most GPUs have hardware support for this type. I tested the rev-animated model with these changes and it's working both with quantization disabled and quantization set to bf16.

> I think it's a better practice to avoid including too many unrelated changes like that in one PR.
> 
> This makes it harder to review, if some changes are bad, the whole PR can't be merged, and it also has a higher chance of breaking many other pending PRs.

I moved the wtype changes to a separate pr & split this one into two commits as the changes are stacked.

leejet:
> I think bf16 has full support though for all the ops

Most ggml ops also do not support bf16.

tsite:
> > I think bf16 has full support though for all the ops
> 
> Most ggml ops also do not support bf16.

Which ones? The models I tested with worked fine quantizing to bf16. Anyways I added a line to disable bf16 for now just in case, feel free to take another look and let me know what you think

leejet:
For example, conv2d does not support bf16 https://github.com/ggml-org/ggml/blob/master/src/ggml-cpu/ops.cpp#L6672. In the Conv2d block, I specified the weight type as fp16, so it gets converted during loading, which makes conv2d work fine. Many other ops also don’t support bf16. If we don’t convert bf16 to f32 when loading tensor storage, we’ll need to recheck many places for potential issues. In addition, future development and maintenance would require extra consideration for bf16 compatibility, adding a lot of mental overhead. Based on this, I don’t think removing the bf16 to f32 conversion is a good idea.

PR #918, opened 2025-10-25
Fix (most) clangd include-related warnings

stduhpf:
There's still this one, which is obviously wrong (the "fix" is to remove the include, which of course breaks vulkan builds):

`ggml_extend.hpp: Included header ggml-vulkan.h is not used directly (fix available)`


stduhpf:
Nevermind, doesn't build on CI

PR #939, opened 2025-11-04
feat: add support for SD2.x with TINY U-Nets

akleine:
docs: fix grammar and style errors, add info about TINY SD2.x models


akleine:
After SD1.x  and SDXL  with  TINY U-Net now we also have SD2.x support.

I think these models are very useful on small devices like Raspberry Pi  or Android phones  etc.
But I think this three TINY U-Net examples should be enough.

Once again:   @leejet and all: Thank you for the program !

Issue #949, opened 2025-11-07
[Bug] PhotoMaker V2 core dumps if cmd line is incomplete


akleine:
### Git commit

8f6c5c217b1f6f27a8aa5fb78d3390fa849fc96a

### Operating System & Version

Ubuntu 24.04 

### GGML backends

CUDA

### Command-line arguments used

sd -m ~/SD_models/sdxl/sd_xl_base_1.0_0.9vae.safetensors --vae ~/SD_models/sdxl/sdxl_vae.safetensors --clip-on-cpu --sampling-method euler --diffusion-fa -H 1024 -W 1024 --seed 5 --steps 40 --cfg-scale 5 --photo-maker ~/SD_models/loras_sdxl/photomaker-v2.safetensors --pm-id-images-dir ./photomaker_examples/newton_man --pm-style-strength 15 -p "half portrait of a man img, playing guitar, wearing blue shirt, dark background, RAW photo, 8k uhd, high quality" -n "Cropped, out of frame, out of focus, worst quality, greyscale, bad anatomy, bad hands" -o PM2.png 

### Steps to reproduce

Try above command line :
1.  "as is"
2.  with additional arg :   --pm-id-embed-path  ./photomaker_examples/newton_man/id_embeds.bin

### What you expected to happen

sd should warn user and continue w/o using PM2 ....
(...or maybe clean exit)

### What actually happened

crash / core dump

### Logs / error messages / stack trace

....
[INFO ] stable-diffusion.cpp:2615 - TXT2IMG
[INFO ] stable-diffusion.cpp:947  - attempting to apply 0 LoRAs
[INFO ] stable-diffusion.cpp:967  - apply_loras completed, taking 0.00s
[INFO ] stable-diffusion.cpp:2177 - pmid_lora apply completed, taking 0.15s
[ERROR] ggml_extend.hpp:293  - failed to open ''
....... Speicherzugriffsfehler  (Speicherabzug geschrieben) ../build/bin/sd -m ~/SD_models/sdxl/sd_xl_base_1.0_0.9vae.safetensors --vae ~/SD_models/sdxl/sdxl_vae.s
afetensors --clip-on-cpu --sampling-method euler --diffusion-fa -H 1024 -W 1024 --seed 5 --steps 40 --cfg-scale 5 --photo-maker ~/SD_models/loras_sdxl/photomaker-v2.safetensors --pm-id-image
[...]

PR #950, opened 2025-11-07
fix: prevent core dump in PM V2 in case of incomplete cmd line

akleine:
docs: add hint about PM2 args in command line
for details  also see    https://github.com/leejet/stable-diffusion.cpp/issues/949

### DIFF
--- a/docs/photo_maker.md
+++ b/docs/photo_maker.md
@@ -40,7 +40,7 @@ Running PMV2 is now a two-step process:
 ```
 python face_detect.py input_image_dir
 ```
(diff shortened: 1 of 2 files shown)


wbruna (approved):
LGTM

akleine:
Thank you very much!

Unfortunately the user has some more chances here to do bad things, for example he could delete or add one of the images inside the folder *after* running face_detect.py.
In such cases sd.cpp will terminate with an assert  and backtrace.

Do you think that sd.cpp should be more "waterproof" in such cases ? 
If you confirm I would care a little bit more about that, at least give the user some more feedback what is going wrong.

Issue #947, opened 2025-11-06
[Bug] PhotoMaker is broken since Oct 13


akleine:
### Git commit

2e9242e37ff6c8c076f0493f747d8ea5c126cb38

### Operating System & Version

Ubuntu 24.04 x86_64

### GGML backends

CUDA

### Command-line arguments used

sd -m ~/SD_models/sdxl/sd_xl_base_1.0_0.9vae.safetensors  	--vae ~/SD_models/sdxl/sdxl_vae.safetensors  	--clip-on-cpu --sampling-method euler --diffusion-fa  	-H 1024 -W 1024 --seed 5 --steps 40  	--cfg-scale 5  	--photo-maker ~/SD_models/loras_sdxl/photomaker-v1.safetensors  	--pm-id-images-dir ./photomaker_examples/newton_man 	--pm-style-strength 15  	-p "half portrait of a man img, playing guitar, wearing blue shirt, dark background, RAW photo, 8k uhd, high quality"  	-n "Cropped, out of frame, out of focus, worst quality, greyscale, bad anatomy, bad hands"  	-o PM1.png 

### Steps to reproduce

run above script twice:
1. using sd built with  commit   c64994d      OKAY 
2. using sd built with  commit   2e9242e      BROKEN

### What you expected to happen

Mr Newton is playing on the guitar:

<img width="1024" height="1024" alt="Image" src="https://github.com/user-attachments/assets/ba772cae-ccc4-45db-9076-d0342aab4757" />

### What actually happened

Anyone else (except Mr Newton) plays here:

<img width="1024" height="1024" alt="Image" src="https://github.com/user-attachments/assets/08ede06d-0525-44a2-bc12-b96ca456cd21" />

### Additional context / environment details

tested both 
on CPU  and CUDA ( 8GB VRAM)

leejet:
This issue has been fixed.

PR #957, opened 2025-11-09
fix: resolve issue with pmid

leejet:
Fix https://github.com/leejet/stable-diffusion.cpp/issues/947.


akleine:
THANK YOU very much!

PR #967, opened 2025-11-10
chore: add sd_ prefix to a few functions

### LINKED ISSUE #1011: [Bug] Please consider renaming or inlining std::string format(const char* fmt, ...) in util.cpp, as it conflicts with llama.cpp def
### Git commit

28ffb6c13df58345ecfc6b47ff6dc8af446e0fbd

Ubuntu 24.04

### GGML backends

Vulkan

cmake --build build -j

### Steps to reproduce

Create a cmake project with both llama.cpp and stable-diffusion.cpp (latest versions) and try to compile.

I have a project which includes both llama.cpp and stable-diffusion.cpp , but it fails to link because both libs declare a format function. Locally, I can fix it if I inline the format function.

### What actually happened

[build] /usr/bin/ld: libs/stable-diffusion.cpp/libstable-diffusion.a(util.cpp.o): in function `format[abi:cxx11](char const*, ...)':
[build] ... snipped ..../libs/stable-diffusion.cpp/util.cpp:60: multiple definition of `format[abi:cxx11](char const*, ...)'; libs/llama.cpp/src/libllama.a(llama-impl.cpp.o): ... snipped ..../libs/llama.cpp/src/llama-impl.cpp:82: first defined here
[build] collect2: error: ld returned 1 exit status

_No response_

_No response_

wbruna:
This helps avoid a name conflict (for instance with the original function) when linking against the library.

Edit: fixes #1011

### DIFF
--- a/model.cpp
+++ b/model.cpp
@@ -265,8 +265,8 @@ void convert_tensor(void* src,
         } else {
             auto qtype = ggml_get_type_traits(src_type);
             if (qtype->to_float == nullptr) {
-                throw std::runtime_error(format("type %s unsupported for integer quantization: no dequantization available",
-                                                ggml_type_name(src_type)));
+                throw std::runtime_error(sd_format("type %s unsupported for integer quantization: no dequantization available",
+                                                   ggml_type_name(src_type)));
             }
             qtype->to_float(src, (float*)dst, n);
         }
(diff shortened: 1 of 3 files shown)


wbruna:
Added the `sd_` prefix to `format()` too, replacing #1012 .

PR #991, opened 2025-11-20
fix: always initialize the default scheduler

wbruna:
If the context was reused for new generations, choosing the default scheduler wouldn't change it back from a previously chosen one.


wbruna:
This was superseded by #998 .

PR #992, opened 2025-11-21
In-progress Rocm Builds

BillyOutlast:
I was wondering if you might be able to lend me a hand in regards to how to configure rocm builds properly?
Im trying to automate them with a github action


leejet:
It appears that you’ve removed the other build CI workflows, which is unacceptable. The original *sd.cpp* already supports ROCm CI builds—does that not meet your requirements?

Issue #1005, opened 2025-11-24
RFC: should sd.cpp print a version string ...?


akleine:
Hello,
what about printing  such a _version string_ a the beginning just like this:

```
$ ./sd_run_my_model_script.sh
**sd.cpp built using commit #15ea27af** 
Option:
    n_threads:                         8
    mode:                              img_gen
    model_path:                        /home/blabla....    
....... etc etc
```

examples/cli/main.cpp   should   include a file called   "commit_number.h"  and later add one line:

```
#include  "commit_number.h
int main(int argc, const char* argv[]) {
       SDParams params;
       **printf("sd.cpp built using commit #%s\n",COMMIT_SDCPP);**
       parse_args(argc, argv, params);

```
while commit_number.h is generated by 

```
#!/bin/bash 
COMMIT_SDCPP=$(git rev-parse --short=8 HEAD) 
echo "constexpr const char* COMMIT_SDCPP = \"$COMMIT_SDCPP\";" > commit_number.h
```

Of course this script should be part of the  make process. 
Making has to be platform independent, also should work if an user fetches the source code without using git.

Unfortunately I do not know very much about cmake  (and I never built sd.cpp  on Win or Mac) , ....
so I won't add a PR now.
What do you think ?

PR #1008, opened 2025-11-25
feat: embed version string and git commit hash

wbruna:
A simple POC for adding a version string to the library and main.cpp (#1005).

The CMake logic comes directly from `ggml/CMakeLists.txt`. I opted for adding a define to a specific existing source file to make the library easier to build without CMake.

For this test, I added two version strings: a simple commit hash, and a fancier one including the current tag (or last tag and hash if HEAD is not on a tag), and a '+' to indicate file changes:

```
./sd -v | head -n2
error: the following arguments are required: prompt
stable-diffusion.cpp version master-377-2034588+, commit 2034588
Usage: ./b_w_vulkan/bin/sd [options]
```
After this commit:
```
./b_w_vulkan/bin/sd -v | head -n2
error: the following arguments are required: prompt
stable-diffusion.cpp version master-377-2034588-1-gff8abbf, commit ff8abbf
Usage: ./b_w_vulkan/bin/sd [options]
```

### DIFF
--- a/examples/cli/main.cpp
+++ b/examples/cli/main.cpp
@@ -1891,7 +1891,6 @@ void step_callback(int step, int frame_count, sd_image_t* image, bool is_noisy, 
 }
 
 int main(int argc, const char* argv[]) {
-
     if (argc > 1 && std::string(argv[1]) == "--version") {
         std::cout << version_string() << "\n";
         return EXIT_SUCCESS;
--- a/stable-diffusion.h
+++ b/stable-diffusion.h
@@ -359,9 +359,8 @@ SD_API bool preprocess_canny(sd_image_t image,
                              float strong,
                              bool inverse);
 
-SD_API const char * sd_commit(void);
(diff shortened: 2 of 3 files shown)


examples/cli/main.cpp
```diff
@@ -1646,6 +1646,11 @@ void step_callback(int step, int frame_count, sd_image_t* image, bool is_noisy)
 }
 
 int main(int argc, const char* argv[]) {
+    for (int i = 1; i < argc; i++) {
+        if (strcmp(argv[1], "-v") == 0 || strcmp(argv[1], "--verbose") == 0) {
+            printf("stable-diffusion.cpp version %s, commit %s\n", sd_version(), sd_commit());
+        }
+    }
     SDParams params;
```
wbruna: I'm grabbing the `-v` like this just to make it easier to test, since `parse_args` aborts the program on error.

PR #1012, opened 2025-11-26
refactor: move utility functions out of the global namespace

### LINKED ISSUE #1011: [Bug] Please consider renaming or inlining std::string format(const char* fmt, ...) in util.cpp, as it conflicts with llama.cpp def

wbruna:
Many utility functions have generic names that may conflict with symbols from other libraries, causing linkage issues.

The 'inline namespace' approach seems ideal for this case, since it's automatically added by any source code that includes `util.h`. Moving the whole library to its own namespace would also work, of course.

I kept the functions in place to avoid a huge diff, but they should probably be reorganized by visibility.

This is related to #967 , which I suggested for the same reason.

Should fix #1011 .

### DIFF
--- a/util.h
+++ b/util.h
@@ -10,6 +10,8 @@
 #define SAFE_STR(s) ((s) ? (s) : "")
 #define BOOL_STR(b) ((b) ? "true" : "false")
 
+inline namespace sd_internal {
+
 bool ends_with(const std::string& str, const std::string& ending);
 bool starts_with(const std::string& str, const std::string& start);
 bool contains(const std::string& str, const std::string& substr);
@@ -67,4 +69,7 @@ bool sd_should_preview_noisy();
 #define LOG_INFO(format, ...) log_printf(SD_LOG_INFO, __FILE__, __LINE__, format, ##__VA_ARGS__)
 #define LOG_WARN(format, ...) log_printf(SD_LOG_WARN, __FILE__, __LINE__, format, ##__VA_ARGS__)
 #define LOG_ERROR(format, ...) log_printf(SD_LOG_ERROR, __FILE__, __LINE__, format, ##__VA_ARGS__)
+
+}
+
 #endif  // __UTIL_H__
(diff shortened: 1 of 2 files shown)


util.h
stduhpf, line 1: This is a .h file, not .hpp. There is no such thing as namespaces in C code. Maybe use `#ifdef __cplusplus`? Or use a prefix instead of a namespace.

stduhpf: Actually the whole file contains c++ stuff like std::strings, nevermind. It should probably be a .hpp instad.

wbruna: Yeah, that is far from consistent right now. I believe `stable-diffusion.h` is the only file that's intended to keep compatibility with C.

leejet:
This can only solve part of the symbol issues, and it makes the code a bit messy. I will refactor the relevant parts later. For now, I think we can just rename format in this PR.

wbruna:
I added the `format` rename to #967 , since it can be grouped with that commit.

PR #1017, opened 2025-11-28
feat: add Flux.2 VAE proj matrix for previews

stduhpf:
Seems like the proj is harder to train for this new VAE, reminds me of SDXL. 

p


stduhpf:
Not working, let me look into it. I think the latents are patchified by 2x2 patches or something

Edit: fixed

leejet:
Perhaps placing the logic in preview_latent_video could make the code cleaner.

stduhpf:
@leejet is it good now or should I actually move the calls to `unpatchify_latents` and `repatchify_latents` to be contained within `preview_latent_video`?

leejet:
I think the best approach is to move unpatchify_latents logic into preview_latent_video, which would eliminate the need to call repatchify_latents.

stduhpf:
> I think the best approach is to move unpatchify_latents logic into preview_latent_video, which would eliminate the need to call repatchify_latents.

I will look into that soon. Maybe there's a way to avoid allcoating a temp buffer too by doing so, that would be great

stable-diffusion.cpp
```diff
@@ -1337,10 +1337,17 @@ class StableDiffusionGGML {
         uint32_t dim           = latents->ne[ggml_n_dims(latents) - 1];
 
         if (preview_mode == PREVIEW_PROJ) {
-            const float(*latent_rgb_proj)[channel] = nullptr;
```
wbruna, line 1329: Not really related to the PR, but: maybe move the proj matrix selection to a virtual VAE object member, to avoid the decision tree here?

stduhpf: Indeed, that would probably be cleaner. Maybe in a follow-up PR.

leejet:

PR #1020, opened 2025-11-29
add z-image support

leejet:
```
.\bin\Release\sd.exe --diffusion-model  ..\..\ComfyUI\models\diffusion_models\z_image_turbo-Q8_0.gguf --vae ..\..\ComfyUI\models\vae\ae.sft  --llm ..\..\ComfyUI\models\text_encoders\qwen_3_4b.safetensors -p 'a lovely cat' --cfg-scale 1.0 -v --offload-to-cpu --diffusion-fa
```

output

PR #1034, opened 2025-12-03
fix: prevent NaN issues with Z-Image on certain ROCm setups

wbruna:
As mentioned on #1020 , a scale of 1/8 on the "out" JointAttention block seems to be enough to avoid fully-black generations on my RX 7600 XT (gfx1102) on Linux.

I bumped the scale to 1/16 just to be somewhat safer, but generations seem fine (with minor changes) even at 1/128, so fell free to adjust it.

### DIFF
--- a/z_image.hpp
+++ b/z_image.hpp
@@ -30,7 +30,12 @@ namespace ZImage {
         JointAttention(int64_t hidden_size, int64_t head_dim, int64_t num_heads, int64_t num_kv_heads, bool qk_norm)
             : head_dim(head_dim), num_heads(num_heads), num_kv_heads(num_kv_heads), qk_norm(qk_norm) {
             blocks["qkv"] = std::make_shared<Linear>(hidden_size, (num_heads + num_kv_heads * 2) * head_dim, false);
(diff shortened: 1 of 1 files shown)


z_image.hpp
```diff
@@ -93,7 +98,7 @@ namespace ZImage {
 #endif
             // The purpose of the scale here is to prevent NaN issues in certain situations.
             // For example, when using CUDA but the weights are k-quants.
-            blocks["w2"] = std::make_shared<Linear>(hidden_dim, dim, false, false, force_prec_f32, 1.f / 128.f);
+            blocks["w2"] = std::make_shared<Linear>(hidden_dim, dim, false, false, force_prec_f32, scale);
             blocks["w3"] = std::make_shared<Linear>(dim, hidden_dim, false);
```
wbruna, line 102: This is unrelated to the fix: I noticed the `scale` var wasn't being used here, but forgot to make a separate commit for it.

leejet: This is not a problem.

leejet:
LGTM. Thanks.

PR #1035, opened 2025-12-03
Document CUDA architectures and build instructions

alx:
Added section on CUDA architectures and example for GeForce RTX 4070.

### DIFF
--- a/docs/build.md
+++ b/docs/build.md
@@ -48,11 +48,11 @@ cmake --build . --config Release
 
 Different NVIDIA GPUs have different CUDA architecture: [CUDA GPU Compute Capability](https://developer.nvidia.com/cuda-gpus)
 
-Example, for _GeForce RTX 4070_, use `-DCMAKE_CUDA_ARCHITECTURES=89` additional flag:
+To build for multiple GPU architectures (matching the GitHub CI releases), use `-DCMAKE_CUDA_ARCHITECTURES=90;89;86;80;75`:
 
 ```shell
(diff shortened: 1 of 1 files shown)


docs/build.md
```diff
@@ -44,6 +44,18 @@ cmake .. -DSD_CUDA=ON
 cmake --build . --config Release
 ```
 
+### CUDA Architectures
+
+Different NVIDIA GPUs have different CUDA architecture: [CUDA GPU Compute Capability](https://developer.nvidia.com/cuda-gpus)
+
+Example, for _GeForce RTX 4070_, use `-DCMAKE_CUDA_ARCHITECTURES=89` additional flag:
```
idostyle: Could include all (current) architectures that are also build into the Github CI releases: `-DCMAKE_CUDA_ARCHITECTURES=90;89;86;80;75` instead of one specific card only.

Might make it easier for a more broad range of people.

alx: Thanks, build.md updated

leejet:
CMake automatically infers and sets an appropriate architecture based on the GPU on the current host. For users who are not familiar with CUDA architectures, manually setting CMAKE_CUDA_ARCHITECTURES may introduce risks, as they might choose an architecture that is not compatible with their GPU.

alx:
Hi @leejet, thanks for your feedback, I agree with it. I was initially pushing this PR for myself only, as I needed to include this flag for my setup. No need to introduce new risks. Have a nice day.

Issue #1052, opened 2025-12-05
[Feature] LongCat-Image and LongCat-Image-Edit by meituan-longcat


JohnClaw:
### Feature Summary

add support for these models, please

### Detailed Description

[https://huggingface.co/meituan-longcat/LongCat-Image](url)

Key Features
🌟 Exceptional Efficiency and Performance: With only 6B parameters, LongCat-Image surpasses numerous open-source models that are several times larger across multiple benchmarks, demonstrating the immense potential of efficient model design.
🌟 Powerful Chinese Text Rendering: LongCat-Image demonstrates superior accuracy and stability in rendering common Chinese characters compared to existing SOTA open-source models and achieves industry-leading coverage of the Chinese dictionary.
🌟 Remarkable Photorealism: Through an innovative data strategy and training framework, LongCat-Image achieves remarkable photorealism in generated images.

![Image](https://github.com/user-attachments/assets/12c5c993-908d-4557-bd32-03085d5c1c3d)

[https://huggingface.co/meituan-longcat/LongCat-Image-Edit](url)

Key Features
🌟 Superior Precise Editing: LongCat-Image-Edit supports various editing tasks, such as global editing, local editing, text modification, and reference-guided editing. It has strong semantic understanding capabilities and can perform precise editing according to instructions.
🌟 Consistency Preservation: LongCat-Image-Edit has strong consistency preservation capabilities, specifically scrutinizes whether attributes in non-edited regions, such as layout, texture, color tone, and subject identity, remain invariant unless targeted by the instruction, is well demonstrated in multi-turn editing.
🌟 Strong Benchmark Performance: LongCat-Image-Edit achieves state-of-the-art (SOTA) performance in image editing tasks while significantly improving model inference efficiency, especially among open-source image editing models.

![Image](https://github.com/user-attachments/assets/f189848a-fea5-4183-825b-4080d9ca0b2d)

stduhpf:
Looks simmilar to Flux1 architecturally. It uses Flux VAE too, and Qwen2.5 vl (7B?) as multimodal encoder.
<img width="15912" height="4835" alt="Image" src="https://github.com/user-attachments/assets/e47acfd6-cb30-49f7-bf9c-9193a7737088" />

<img width="2487" height="1729" alt="Image" src="https://github.com/user-attachments/assets/b774f9aa-8955-4f4c-b272-274a0e55e8eb" />

And there's also a "Dev" model that is actually developer-friendly, go figure.


[...]

PR #1053, opened 2025-12-05
feat: Longcat-Image / Longcat-Image-Edit support

stduhpf:
for https://github.com/leejet/stable-diffusion.cpp/issues/1052

Seems to only work with `--clip-on-cpu`, at least on HIP (ROCm) backends.

`sd-cli.exe --diffusion-model ..\ComfyUI\models\unet\LongCat-Image-Q8_0.gguf  --vae ..\ComfyUI\models\vae\flux\ae.safetensors  --cfg-scale 4.0 --sampling-method euler -v --clip-on-cpu -p "A cinematic, melancholic photograph of a solitary hooded figure walking through a sprawling, rain-slicked metropolis at night. The city lights are a chaotic blur of neon orange and cool blue, reflecting on the wet asphalt. The scene evokes a sense of being a single component in a vast machine. Superimposed over the image in a sleek, modern, slightly glitched font is the philosophical quote: \"THE CITY IS A CIRCUIT BOARD, AND I AM A LONG CAT.\" -- moody, atmospheric, profound, dark academic"  --preview proj --steps 20 --qwen2vl ..\ComfyUI\models\clip\Qwen2.5-VL-7B-Instruct.Q4_K_M.gguf --diffusion-fa --color -W 1024 -H 1024`

output

Test models (converted to bfl format) can be found there: 
- https://huggingface.co/stduhpf/LongCat-Image-gguf/tree/main
- https://huggingface.co/stduhpf/LongCat-Image-Edit-gguf/tree/main
- https://huggingface.co/stduhpf/LongCat-Image-Dev-gguf/tree/main

Inference for models in diffusers format seem to be still broken

I recommend using the `smoothstep` scheduler with `euler` sampling method.


wbruna:
That _does_ look a bit like a circuit board...

stduhpf:
TODO for when image generation works
image

stduhpf:
I can't figure out what I'm doing wrong, I think it is supposed to be working just like Flux1, but with different PE indices and Qwen Text Encoder.... Maybe I'm missing an important detail  but I can't find it.

stduhpf:
I tried using my SplitAttention thing on a Flux model converted to diffusers format, and 
output
I guess I found what is not working. I will try converting LongCat to Flux format and see if it works.

stduhpf:
I think I got it?
output

With the padding fixed, but with diffusers format:
output

stduhpf:
With the character-level tokenization trick:
output

Might need testing to make sure the current implementation supports languages that don't use the latin alphabet. Also for now it's applied to text wrapped in single quotes (` '`)  only.

stduhpf:
Oh no, why are there so many conflicts now?

stduhpf:
Using `'` as a quote delimiter was a bad idea because it's the same symbol used for apostrophes. I will change it to detect `"` instead

stduhpf:
Somehow not fully working yet, but it's definitely able to see it's supposed to be a cat holding a sign, maybe because of the vision model
`sd.exe --diffusion-model ..\ComfyUI\models\unet\longcat_edit_bfl_format-Q8_0.gguf  --vae ..\ComfyUI\models\vae\flux\ae.safetensors  --cfg-scale 4.5 --sampling-method euler -v --offload-to-cpu   --preview proj --steps 50 --vae-tile-size 128 --qwen2vl ..\ComfyUI\models\clip\Qwen2.5-VL-7B-Instruct.Q4_K_M.gguf --color --seed 0 -r .\assets\flux\flux1-dev-q8_0.png --llm_vision ..\ComfyUI\models\clip_vision\Qwen2.5-VL-7B-Instruct.mmproj-f16.gguf -p "Change the text to say \"I'm a long one\""`

|ref|out|
|--|--|
|flux1-dev-q8_0|output|

(Also I made the change so it now needs double quotes around literal text)

stduhpf:
output

Somehow couldn't get it to remove the original text, but there it goes

stable-diffusion.cpp
```diff
@@ -1316,10 +1336,23 @@ class StableDiffusionGGML {
...
+                if (sd_version_is_flux2(version)) {
+                    latent_rgb_proj = flux2_latent_rgb_proj;
+                    latent_rgb_bias = flux2_latent_rgb_bias;
+                    patch_sz        = 2;
+                }
+            } else if (dim == 64) {
+                if (sd_version_is_flux(version) || sd_version_is_z_image(version) || sd_version_is_longcat(version)) {
+                    latent_rgb_proj = flux_latent_rgb_proj;
+                    latent_rgb_bias = flux_latent_rgb_bias;
+                    patch_sz        = 2;
```
leejet: For flux and z-image, should patch_sz be 1?

vae.hpp
```diff
@@ -553,7 +553,7 @@ class AutoencodingEngine : public GGMLBlock {
 
     struct ggml_tensor* decode(GGMLRunnerContext* ctx, struct ggml_tensor* z) {
         // z: [N, z_channels, h, w]
-        if (sd_version_is_flux2(version)) {
+        if (sd_version_is_flux2(version) || sd_version_is_longcat(version)) {
```
leejet: I think for longcat, it’s unnecessary to set the VAE patch_size to 2. With the Flux architecture, setting the patch_size to 2 there is already sufficient.

stduhpf: Yes I did try that before and it seems to work just as well (though the results are completely different with same seed because the noise pattern changes between the two implementations). I did it that way because that's how it is implemented in diffusers code (for base Flux.1 too). It also saves an almost negligible amount of compute since it's not patchifying and unpatchifying at every step.
Should I keep it consistent with Flux.1 instead?

leejet: From Longcat’s pipeline, I think it’s actually implemented the same way as Flux1: it generates noise first and then applies patchify, rather than generating patchified latents directly like Flux2.

https://github.com/meituan-longcat/LongCat-Image/blob/main/longcat_image/pipelines/pipeline_longcat_image.py#L359

stduhpf: Ah right. Again, in practice that shouldn't matter too much. But I can make sure it stays consistent with Flux.1.

stable-diffusion.cpp
stduhpf: It should be, yes, but if the dim is 64, it means that the patch_size must have been set to 2.

vae.hpp
leejet: Thanks, this will make the code more concise.

Rocky-Lee-001:
May I ask which comfyui node is used to load this GGUF model?

stduhpf:
Now supports UTF-8 encoding properly for the quoted text. (also quote characters are no longer excluded from the prompt after being parsed, seems to help a bit, especially with longer text.)

stduhpf:
> May I ask which comfyui node is used to load this GGUF model?

@Rocky-Lee-001 I don't think LongCat-Image is natively supported by ComfyUI yet. You could give https://github.com/sooxt98/comfyui_longcat_image a try, maybe it works well with the GGUF node for comfyUI?

ggml_extend.hpp
```diff
@@ -2173,6 +2173,83 @@ class Linear : public UnaryBlock {
     }
 };
 
+class SplitLinear : public Linear {
```
leejet: If this part has no effect, I think we can remove the related code. In fact, even if it does have some effect, additional work is required to handle it when LoRA uses QKV format, so I wouldn’t really recommend this approach.

leejet:
I’m not sure whether I did something wrong on my end, but I got a strange image.

```
.\bin\Release\sd-cli.exe --diffusion-model  ..\models\longcat_bfl_format-Q4_K_M.gguf --vae ..\..\ComfyUI\models\vae\ae.sft  --llm ..\..\ComfyUI\models\text_encoders\Qwen2.5-VL-7B-Instruct-Q8_0.gguf -p 'a lovely cat' --cfg-scale 5.0 -v --offload-to-cpu --diffusion-fa
```

output

ggml_extend.hpp
stduhpf: Thi is used when loading Flux diffusion models with diffusers naming convention, which has the qkv matrices split as individual linear layers rather than one big linear layer. For some reason it is not quite working, not sure why.

stduhpf:
@leejet that's strange. I can reproduce it with the same prompt though (even with Q8_0 model), but I haven't gotten anything like this in my earlier testing. Maybe There's a linear layer that could use scaling?

Does not seem related to seed. 

It's a combination of short prompts + low resolution that seems to cause it.

leejet:
I think there are still some differences between the implementation in this PR and the official pipeline—for example, the way token padding is handled and whether masks are used. I suspect these differences might be what caused the strange blocky artifacts in the generated images. I tried to fix it, but didn’t succeed; now it’s producing completely black images.

https://github.com/leejet/stable-diffusion.cpp/tree/longcat-fix

Yurchikian:
I've tried this code and found out that short prompts lead to such a blocky modern-art generations, while long descriptive prompts produce higher quality results

For example `~/projects/longcat.cpp/build/bin/sd --diffusion-model ~/Drive/AI/ComfyUI/models/unet/longcat_bfl_format-Q4_K_M.gguf  --diffusion-fa --vae ~/Drive/AI/ComfyUI/models/vae/ae.sft --sampling-method euler --cfg-scale 3.0 --steps 20 --llm ~/Drive/AI/ComfyUI/models/text_encoders/Qwen2.5-VL-7B-Instruct-Q3_K_M.gguf --color -W 512 -H 512  --seed 128 -o output_$(date +%Y-%m-%d_%H-%M-%S).png --vae-tiling -p "Orange cat"
`:
output_2025-12-16_22-24-08

Repeating "Orange Cat" several times:
output_2025-12-16_22-25-39

`~/projects/longcat.cpp/build/bin/sd --diffusion-model ~/Drive/AI/ComfyUI/models/unet/longcat_bfl_format-Q4_K_M.gguf  --diffusion-fa --vae ~/Drive/AI/ComfyUI/models/vae/ae.sft --sampling-method euler --cfg-scale 3.0 --steps 20 --llm ~/Drive/AI/ComfyUI/models/text_encoders/Qwen2.5-VL-7B-Instruct-Q3_K_M.gguf --color -W 512 -H 512  --seed 128 -o output_$(date +%Y-%m-%d_%H-%M-%S).png --vae-tiling -p "This is a portrait photograph of an Orange Cat sitting on the table by the window. We can see the room interior, fancy 50s style Art-Deco ornaments on the wallpaper. The weather outside is fine, the sun is shining. Wintertime snow on the ground, people playing with snowballs. The cat is large and shaped like a sphere. Whiskers are long and fangs are sharp. Phone snapshot, photograph."`
output_2025-12-16_22-28-59

Increasing step count would reduce artifacts even more. However face seems bit blocky anyway, not sure if it would be true for original workflow, can not run non-gguf model on my GPU. Also increasing resolution would not help with those artifacts.

Hope my observations might be helpful


P.S. I've also checked out @leejet `longcat-fix` branch. It generated black images indeed, until I added `--clip-on-cpu`, then i could get image with just "Orange Cat" prompt!
output_2025-12-16_23-29-39

stduhpf:
@leejet I figured your fix is actually working fine with `--clip-on-cpu`.

`.\buildhip\bin\sd-cli.exe --diffusion-model E:\Downloaded_models\img\longcat_bfl_format-Q8_0.gguf  --vae ..\ComfyUI\models\vae\flux\ae.safetensors --llm ..\ComfyUI\models\clip\llms\Qwen2.5-VL-7B-Instruct.Q4_K_M-symlink.gguf  --diffusion-fa --sampling-method euler --cfg-scale 3.0 --steps 20  --color -W 512 -H 512  --seed 128 --vae-tiling -p "Orange cat"  --clip-on-cpu`

output

leejet:
I’m glad to know that my fix actually worked. However, the problem still persists when not using clip-on-cpu, and I’ll look into it when I have time.

stduhpf:
https://github.com/Comfy-Org/ComfyUI/pull/12597/

leejet:
It currently looks like everything is normal now.

```
.\bin\Release\sd-cli.exe --diffusion-model  ..\..\ComfyUI\models\diffusion_models\LongCat-Image-Q4_K_M.gguf --vae ..\..\ComfyUI\models\vae\ae.sft --llm ..\..\ComfyUI\models\text_encoders\Qwen2.5-VL-7B-Instruct-Q8_0.gguf -p "a lovely cat holding a sign says 'longcat.cpp'" --cfg-scale 5.0 --sampling-method euler --flow-shift 3 -v --offload-to-cpu --diffusion-fa
```

example

leejet:

PR #1059, opened 2025-12-06
feat: support mmap for model loading

wbruna:
Introduces a new `--use-mmap` flag that replaces model loading I/O operations with `mmap` + `memcpy`.

In my tests, this helps model loading speed slightly, though the gain was never higher than half a second. Its primary benefit right now is validation of the `mmap` backend implementation. Later, I plan to extend this to allow the mapped file to serve directly as weight storage for backends that use main memory.

I used a non-default flag to be extra safe, but we could arguably follow llama.cpp approach, with a `--no-mmap` flag to disable it instead.

I was only able to test (and build...) it under Linux, so additional testing is very welcome 🙂

### DIFF
--- a/examples/cli/README.md
+++ b/examples/cli/README.md
@@ -48,6 +48,7 @@ Context Options:
   --vae-tiling                             process vae in tiles to reduce memory usage
   --force-sdxl-vae-conv-scale              force use of conv scale on sdxl vae
   --offload-to-cpu                         place the weights in RAM to save VRAM, and automatically load them into VRAM when needed
+  --mmap                                   whether to memory-map model
   --control-net-cpu                        keep controlnet in cpu (for low vram)
   --clip-on-cpu                            keep clip in cpu (for low vram)
   --vae-on-cpu                             keep vae in cpu (for low vram)
--- a/examples/server/README.md
+++ b/examples/server/README.md
@@ -43,6 +43,7 @@ Context Options:
   --control-net-cpu                        keep controlnet in cpu (for low vram)
   --clip-on-cpu                            keep clip in cpu (for low vram)
   --vae-on-cpu                             keep vae in cpu (for low vram)
+  --mmap                                   whether to memory-map model
   --diffusion-fa                           use flash attention in the diffusion model
   --diffusion-conv-direct                  use ggml_conv2d_direct in the diffusion model
   --vae-conv-direct                        use ggml_conv2d_direct in the vae model


Green-Sky:
How much value would it be if llama.cpp exported the mmap stuff as a library?

wbruna:
> How much value would it be if llama.cpp exported the mmap stuff as a library?

I don't think it'd help that much right now. The mmap part itself is more-or-less straightforward; replacing the current alloc+memcpy code with a buffer managed externally will be much trickier.

valkarias:
Have you experimented with MMaping then copying to GPU? 
In my experience. I've restricted MMapping only to CPU inference & loading. MMap -> copy to GPU became a bottleneck for some reason (I assume page size potentially?)

wbruna:
> Have you experimented with MMaping then copying to GPU? In my experience. I've restricted MMapping only to CPU inference & loading. MMap -> copy to GPU became a bottleneck for some reason (I assume page size potentially?)

Not yet. Right now I'm just reusing the I/O buffer; adding a separate code path to deliver the mapped area directly to the backend just to avoid a memcpy sounded like too much change for too little potential gain.

That behavior you describe sounds... odd. At least on Linux, large dynamically-allocated memory areas use mmap as backend anyway, so they should behave the same. Maybe it's a difference between _file_ -backed and anonymous mappings.

wbruna:
The direct use of the memory-mapped area for the tensor weights seems a lot more involved, so I think it' best to keep this simple for now, and add that support on a follow-up PR.

I changed the command-line to enable it by default, since that's the llama.cpp approach, but I can change it back if an opt-in is preferable.

rendang-github:
This looks like a great starting point toward an eventual "zero-copy" implementation for compatible tensors.  I had looked into that myself too and observed that it in indeed a lot more involved, as you have the additional branching and requirements around word size conversion (eg FP16->FP32) when the src and dst tensor bit widths don't match up. which would require an alloc in those cases anyway.

At this point I have just read the diffs and intend to build and run some tests later when I get a chance, but I have some questions:

1. Why `MAP_PRIVATE` instead of `MAP_SHARED`?  The mapped location should only need read access, and MAP_SHARED means that the mapped pages are also shared with any other file handles/processes that also map the same region.  MAP_PRIVATE also forces the swapper to reserve swap space for the buffer in case writes come in (which should not happen in our use case).
2. Why not use `madvise(mapped_data, file_size, MADV_SEQUENTIAL)`?  If we're doing a straight up `memcpy` out of the buffer (or the eventual equivalent when uploading mmap'ed tensors to GPU) then our use case is indeed sequential and `MADV_SEQUENTIAL` lets the swapper bias the page loading toward that specific use case, and lets it immediately discard pages that we've already copied reducing swap pressure.  The fact that it's loading multiple disk blocks ahead should improve pipeline performance on loads from persistent storage too.  Of course, performance will go to hell if we're not sequentially reading pages, but that's not the use case here right?

wbruna:
> This looks like a great starting point toward an eventual "zero-copy" implementation for compatible tensors. I had looked into that myself too and observed that it in indeed a lot more involved, as you have the additional branching and requirements around word size conversion (eg FP16->FP32) when the src and dst tensor bit widths don't match up. which would require an alloc in those cases anyway.

The type conversion itself isn't that complicated, because the current code already has a separate branch for the case where the weights don't need to be converted; the main issue is it'll need two scans, because we'll only know how much memory will be needed for the _non_-mmapped tensors after a first scan. On the other hand, the tensor allocation and ownership will demand more complicated changes (e.g. it still isn't clear to me how it'd interact with the `--offload-to-cpu` functionality).

> At this point I have just read the diffs and intend to build and run some tests later when I get a chance, but I have some questions:
> 
>     1. Why `MAP_PRIVATE` instead of `MAP_SHARED`?  The mapped location should only need read access, and MAP_SHARED means that the mapped pages are also shared with any other file handles/processes that also map the same region.  MAP_PRIVATE also forces the swapper to reserve swap space for the buffer in case writes come in (which should not happen in our use case).

Swap isn't reserved for unchanged pages: they are backed by the file itself until they are written to (talking about Linux at least). It _would_ be different if the file were written to (by another process), but in that case we'd actually want a private mapping: a file change could mean a change to tensor sizes or offsets, messing up the already loaded mappings.

>     2. Why not use `madvise(mapped_data, file_size, MADV_SEQUENTIAL)`?  If we're doing a straight up `memcpy` out of the buffer (or the eventual equivalent when uploading mmap'ed tensors to GPU) then our use case is indeed sequential and `MADV_SEQUENTIAL` lets the swapper bias the page loading toward that specific use case, and lets it immediately discard pages that we've already copied reducing swap pressure.  The fact that it's loading multiple disk blocks ahead should improve pipeline performance on loads from persistent storage too.  Of course, performance will go to hell if we're not sequentially reading pages, but that's not the use case here right?

I couldn't measure any consistent gains with MADV_SEQUENTIAL on my system, so I've kept it simple for now. Of course, it could be added if it proved to be beneficial in some cases.

leejet:
On Windows, when loading relatively large weights, `mmap` always takes longer to load. Perhaps we shouldn’t enable `mmap` by default.

mmap
```
[INFO ] model.cpp:1599 - loading tensors completed, taking 12.33s (process: 0.01s, read: 10.52s, memcpy: 0.00s, convert: 0.07s, copy_to_backend: 0.00s)
```

read
```
[INFO ] model.cpp:1599 - loading tensors completed, taking 9.92s (process: 0.01s, read: 9.54s, memcpy: 0.00s, convert: 0.07s, copy_to_backend: 0.00s)
```

PR #1066, opened 2025-12-08
feat: add more caching methods

rmatif:
Adding an experimental variant of an EasyCache-like feature for UNet models. I came up with the name "ucache" (if someone has a better suggestion, I'd take it). For now it uses a step-level skipping mechanism. I want to make it per-block to provide more granularity and control, but the current UNet implementation doesn't allow that for now, and the static nature of ggml graphs makes it difficult to capture precise UNet blocks. I have found the results good enough for now to make this a first iteration

Threshold may vary with the sampler + scheduler combo

`./build/bin/sd -m models/model.safetensors --cfg-scale 7 -p "a cute cat sitting on a red pillow" --steps 20 -H 1024 -W 1024 -s 42 --ucache 1,0.15,0.95 --diffusion-fa`

20 Steps:

| Baseline | 5/20 | 7/20 | 8/20 |
|:--------:|:----:|:----:|:----:|
| baseline |  |  |  |
| 0 steps skipped (1x) | 5 steps skipped (~1.33x) | 7 steps skipped (~1.54x) | 8 steps skipped (~1.67x) |

30 Steps: 

| Baseline | 10/30 | 11/30 | 12/30 | 14/30 | 15/30 |
|:--------:|:-----:|:-----:|:-----:|:-----:|:-----:|
| baseline |  |  |  |  |  |
| 0 skipped (1x) | 10 skipped (1.5x) | 11 skipped (~1.58x) | 12 skipped (~1.67x) | 14 skipped (~1.88x) | 15 skipped (2x) |

Supersedes https://github.com/leejet/stable-diffusion.cpp/pull/705

### DIFF
--- a/cache_dit.hpp
+++ b/cache_dit.hpp
@@ -5,29 +5,29 @@
 #include <cmath>
 #include <limits>
 #include <string>
-#include <vector>
 #include <unordered_map>
+#include <vector>
 
 #include "ggml_extend.hpp"
 
 struct DBCacheConfig {
-    bool enabled                    = false;
-    int Fn_compute_blocks           = 8;
-    int Bn_compute_blocks           = 0;
-    float residual_diff_threshold   = 0.08f;
-    int max_warmup_steps            = 8;
-    int max_cached_steps            = -1;
-    int max_continuous_cached_steps = -1;
+    bool enabled                        = false;
+    int Fn_compute_blocks               = 8;
+    int Bn_compute_blocks               = 0;
+    float residual_diff_threshold       = 0.08f;
+    int max_warmup_steps                = 8;
+    int max_cached_steps                = -1;
+    int max_continuous_cached_steps     = -1;
     float max_accumulated_residual_diff = -1.0f;
     std::vector<int> steps_computation_mask;
-    bool scm_policy_dynamic         = true;
+    bool scm_policy_dynamic = true;
 };
 
 struct TaylorSeerConfig {
-    bool enabled             = false;
-    int n_derivatives        = 1;
-    int max_warmup_steps     = 2;
-    int skip_interval_steps  = 1;
+    bool enabled            = false;
+    int n_derivatives       = 1;
+    int max_warmup_steps    = 2;
+    int skip_interval_steps = 1;
 };
 
 struct CacheDitConfig {
@@ -40,29 +40,31 @@ struct CacheDitConfig {
 };
 
 struct TaylorSeerState {
-    int n_derivatives            = 1;
-    int current_step             = -1;
-    int last_computed_step       = -1;
+    int n_derivatives      = 1;
+    int current_step       = -1;
+    int last_computed_step = -1;
     std::vector<std::vector<float>> dY_prev;
     std::vector<std::vector<float>> dY_current;
 
     void init(int n_deriv, size_t hidden_size) {
         n_derivatives = n_deriv;
-        int order = n_derivatives + 1;
+        int order     = n_derivatives + 1;
         dY_prev.resize(order);
         dY_current.resize(order);
         for (int i = 0; i < order; i++) {
             dY_prev[i].clear();
             dY_current[i].clear();
         }
-        current_step = -1;
+        current_step       = -1;
         last_computed_step = -1;
     }
 
     void reset() {
-        for (auto& v : dY_prev) v.clear();
-        for (auto& v : dY_current) v.clear();
-        current_step = -1;
+        for (auto& v : dY_prev)
+            v.clear();
+        for (auto& v : dY_current)
+            v.clear();
+        current_step       = -1;
         last_computed_step = -1;
     }
 
@@ -72,14 +74,15 @@ struct TaylorSeerState {
 
     void update_derivatives(const float* Y, size_t size, int step) {
         int order = n_derivatives + 1;
-        dY_prev = dY_current;
+        dY_prev   = dY_current;
         dY_current[0].resize(size);
         for (size_t i = 0; i < size; i++) {
             dY_current[0][i] = Y[i];
         }
 
         int window = step - last_computed_step;
-        if (window <= 0) window = 1;
+        if (window <= 0)
+            window = 1;
 
         for (int d = 0; d < n_derivatives; d++) {
             if (!dY_prev[d].empty() && dY_prev[d].size() == size) {
@@ -92,7 +95,7 @@ struct TaylorSeerState {
             }
         }
 
-        current_step = step;
+        current_step       = step;
         last_computed_step = step;
     }
 
@@ -102,15 +105,18 @@ struct TaylorSeerState {
         }
 
         int elapsed = target_step - last_computed_step;
-        if (elapsed <= 0) elapsed = 1;
+        if (elapsed <= 0)
+            elapsed = 1;
 
         std::fill(output, output + size, 0.0f);
         float factorial = 1.0f;
-        int order = static_cast<int>(dY_prev.size());
+        int order       = static_cast<int>(dY_prev.size());
 
         for (int o = 0; o < order; o++) {
-            if (dY_prev[o].empty() || dY_prev[o].size() != size) continue;
-            if (o > 0) factorial *= static_cast<float>(o);
+            if (dY_prev[o].empty() || dY_prev[o].size() != size)
+                continue;
+            if (o > 0)
+                factorial *= static_cast<float>(o);
             float coeff = std::pow(static_cast<float>(elapsed), o) / factorial;
             for (size_t i = 0; i < size; i++) {
                 output[i] += coeff * dY_prev[o][i];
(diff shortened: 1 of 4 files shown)


wbruna:
This also has a nice side-effect on some low-CFG distilled models: the skipped steps help avoid the "overcooked" effect when using too many steps.

> I came up with the name "ucache" (if someone has a better suggestion, I'd take it).

Since EasyCache itself doesn't work with UNet, and ucache uses a similar algorithm, I'd suggest reusing the same command-line parameters and parameter struct, to make it simpler to use both for the command line and frontends.

For the same reason, perhaps a scaling factor could be applied to the threshold, to make similar values have similar behavior (at least for the default value)?

rmatif:
> Since EasyCache itself doesn't work with UNet, and ucache uses a similar algorithm, I'd suggest reusing the same command-line parameters and parameter struct, to make it simpler to use both for the command line and frontends.

I have thought about that, but I'm planning to make more changes to make it depth-aware, so it will diverge from the original EasyCache implementation. Since the latter is working well I wanted to leave it unchanged. I'm still reusing the easycache hooks though to avoid some duplication

> For the same reason, perhaps a scaling factor could be applied to the threshold, to make similar values have similar behavior (at least for the default value)?

If the threshold was the same across samplers/schedulers, I'd say yes but I feel it's a hacky way to do arbitrary scaling depending on that. Plus it's not only different but also too few sensitive sometimes, different values will get you similar skipped steps. We lack granularity inside a single step, I'll work on unifying this

wbruna:
> I have thought about that, but I'm planning to make more changes to make it depth-aware, so it will diverge from the original EasyCache implementation. Since the latter is working well I wanted to leave it unchanged. I'm still reusing the easycache hooks though to avoid some duplication

My suggestion is from a usability side, not development. If I need to specify "turn on the cache implementation" in, say, a Koboldcpp config file or sd.cpp-webui field, it's much easier if I don't have separate fields for each model version. Especially because the model version isn't really available at that point (there's no reliable way to figure it out _outside_ sd.cpp code - not even `main.cpp` has that information). I'd need to either duplicate the fields, and leave to the user to figure out what she needs, or always fill both, and tolerate the warning messages.

In reality, Koboldcpp would avoid that anyway by patching stable-diffusion.cpp (each patch adds maintenance overhead, but it's better than the alternative). Command-line users like sd.cpp-webui, or users that don't patch the library, don't have that option.

Even if the cache types used completely different parameters, they have defaults, so a simple flag/checkbox "turn the default cache on" would still be useful, and easy to be supported and used.

And it'd be zero change for the EasyCache implementation: just reuse its parameter struct. The code won't mind a few extra fields, if they are needed.

> If the threshold was the same across samplers/schedulers, I'd say yes but I feel it's a hacky way to do arbitrary scaling depending on that. Plus it's not only different but also too few sensitive sometimes, different values will get you similar skipped steps. We lack granularity inside a single step, I'll work on unifying this

Again, my comment was about unifying the default value from a user's POV (essentially a flat `*5` on the input for unet, such as 0.2 is always the default). I agree that doesn't matter much, if you intend to have different defaults depending on model version/sampler/scheduler.

Green-Sky:
fyi I had to go down to 0.05 or 0.025 to get stable results with 36 steps.

---

Regarding a unified command argument, maybe something like `--latent-cache` or `--prediction-cache` or something.

leejet:
Maybe we should use --cache-mode to control the caching method (and disable caching if it’s not configured), and use --cache-option to configure the cache parameters?

rmatif:
> Maybe we should use --cache-mode to control the caching method (and disable caching if it’s not configured), and use --cache-option to configure the cache parameters?

Something like that? Or do you want to expose those on the C API?

> fyi I had to go down to 0.05 or 0.025 to get stable results with 36 steps.

On which sampler/scheduler?

> My suggestion is from a usability side, not development. If I need to specify "turn on the cache implementation" in, say, a Koboldcpp config file or sd.cpp-webui field, it's much easier if I don't have separate fields for each model version. Especially because the model version isn't really available at that point (there's no reliable way to figure it out outside sd.cpp code - not even main.cpp has that information). I'd need to either duplicate the fields, and leave to the user to figure out what she needs, or always fill both, and tolerate the warning message

That's a valid point, and I think it will be a preferable goal to reach. The fact that the app isn't aware of the model is true for a lot of options here. I think since it's experimental and I will be iterating on it, it's fine to keep it manual for now, and once it's good enough, having only one cache option on sdcpp for every model

rmatif:
Add some tweaks, now it can sometimes accidentally add a nice aesthetic pattern compared to the baseline

20 Steps:

| Baseline | 5/20 | 6/20 | 7/20 | 8/20 | 9/20 | 10/20 |
|:--------:|:----:|:----:|:----:|:----:|:----:|:-----:|
| baseline_20 |  |  |  |  |  |  |
| 0 steps skipped (1x) | 5 steps skipped (~1.33x) | 6 steps skipped (~1.43x) | 7 steps skipped (~1.54x) | 8 steps skipped (~1.67x) | 9 steps skipped (~1.82x) | 10 steps skipped (2x) |

30 Steps:

| Baseline | 8/30 | 10/30 | 12/30 | 13/30 | 14/30 | 15/30 | 16/30 |
|:--------:|:----:|:-----:|:-----:|:-----:|:-----:|:-----:|:-----:|
| baseline_30 |  |  |  |  |  |  |  |
| 0 steps skipped (1x) | 8 steps skipped (~1.36x) | 10 steps skipped (1.5x) | 12 steps skipped (~1.67x) | 13 steps skipped (~1.76x) | 14 steps skipped (~1.88x) | 15 steps skipped (2x) | 16 steps skipped (~2.14x) |

Green-Sky:
> > fyi I had to go down to 0.05 or 0.025 to get stable results with 36 steps.
> 
> On which sampler/scheduler?

I was using `--cfg-scale 8 --steps 36 --scheduler karras --sampling-method dpm++2m` with cyberrealisticxl.

Will test the new code later and play some more with the params.

leejet:
> Something like that? Or do you want to expose those on the C API?

The command arguments are just as I expected.
By the way, maybe we can put all the cache-related parameters into a single struct in the API?
Then when we add more cache methods later, we won’t need a separate struct for each one — and it seems like multiple cache methods can’t be active at the same time anyway.

examples/cli/README.md
```diff
@@ -124,5 +124,8 @@ Generation Options:
   --skip-layers                            layers to skip for SLG steps (default: [7,8,9])
   --high-noise-skip-layers                 (high noise) layers to skip for SLG steps (default: [7,8,9])
   -r, --ref-image                          reference image for Flux Kontext models (can be used multiple times)
-  --easycache                              enable EasyCache for DiT models with optional "threshold,start_percent,end_percent" (default: 0.2,0.15,0.95)
+  --cache-mode                             caching method: 'easycache' for DiT models, 'ucache' for UNET models (SD1.x/SD2.x/SDXL)
+  --cache-option                           cache parameters: easycache uses "threshold,start,end" (default: 0.2,0.15,0.95).
+                                           ucache uses "threshold,start,end[,decay,relative]" (default: 1.0,0.15,0.95,1.0,1).
```
wbruna: @leejet and @rmatif , what would you think about some kind of named parameters? Something like `threshold=1.0,start=0.15,end=0.15,decay=1.0,relative=1` would be much easier to understand, and to change a specific value without touching the defaults for the others.

(again, I'm thinking about ease of use, not implementation :-)  But something like this could potentially be reused to group related flags for niche use cases; the `--chroma-*` flags, for instance, could all be grouped under a `--chroma-options` single parameter)

rmatif: Indeed that makes sense. @leejet if you agree I can proceed as suggested

leejet: That sounds good to me. Please go ahead with the proposed changes.

rmatif:
Decided to port [cache-dit](https://github.com/vipshop/cache-dit)

| Baseline | Slow | Medium | Fast | Ultra |
|:--------:|:----:|:------:|:----:|:-----:|
| baseline | slow | medium | fast | ultra |
| 1x | 2x | ~2.73x | 3x | 3x |

rmatif:
More results 

`./build/bin/sd-cli -m /workspace/ucache/stable-diffusion.cpp/models/2395247.safetensors --cfg-scale 7 -p "a cute cat sitting on a red pillow" --steps 20 -H 1024 -W 1024 -s 42 --cache-mode ucache --cache-option "threshold=1,start=0.15,end=0.95,decay=1,relative=1,reset=1" --diffusion-fa --sampling-method dpm++2m --scheduler karras`

20 Steps: 

| Baseline | 6/20 | 7/20 | 8/20 | 9/20 | 10/20 | 11/20 |
|:--------:|:----:|:----:|:----:|:----:|:-----:|:-----:|
| baseline_20 |  |  |  |  |  |  |
| 0 steps skipped (1x) | 6 steps skipped (~1.43x) | 7 steps skipped (~1.54x) | 8 steps skipped (~1.67x) | 9 steps skipped (~1.82x) | 10 steps skipped (2x) | 11 steps skipped (~2.22x) |

30 Steps: 

| Baseline | 10/30 | 11/30 | 12/30 | 13/30 | 14/30 | 15/30 | 16/30 | 17/30 | 18/30 | 19/30 | 20/30 |
|:--------:|:-----:|:-----:|:-----:|:-----:|:-----:|:-----:|:-----:|:-----:|:-----:|:-----:|:-----:|
| baseline_30 |  |  |  |  |  |  |  |  |  |  |  |
| 0 steps skipped (1x) | 10 steps skipped (1.5x) | 11 steps skipped (~1.58x) | 12 steps skipped (~1.67x) | 13 steps skipped (~1.76x) | 14 steps skipped (~1.88x) | 15 steps skipped (2x) | 16 steps skipped (~2.14x) | 17 steps skipped (~2.31x) | 18 steps skipped (2.5x) | 19 steps skipped (~2.73x) | 20 steps skipped (3x) |

@leejet I believe this is ready for review

leejet:

PR #1079, opened 2025-12-12
Fix recent GEGLU changes slowing down inference unet models

### LINKED ISSUE #984: Performance regression
I’m seeing a major performance drop in the Vulkan backend after PR #969.

On my system, generating a 1024×1024 SDXL image with the same command went from ~1.09 s/it before the PR to ~2.47 s/it after it.

I noticed that others have mentioned similar slowdowns in the PR discussion itself, but I think it needs a separate issue so the regression doesn’t get lost and can be tracked properly.

The regression has been tested on my end on a Radeon RX 7800XT.

stduhpf:
Fixes #984

### DIFF
--- a/common.hpp
+++ b/common.hpp
@@ -193,11 +193,12 @@ public:
         // return: [ne3, ne2, ne1, dim_out]
         auto proj = std::dynamic_pointer_cast<Linear>(blocks["proj"]);
 
-        x          = proj->forward(ctx, x);  // [ne3, ne2, ne1, dim_out*2]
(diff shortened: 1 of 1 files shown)


stduhpf:
@leejet It looks like ggml_ext_chunk() is a serious bottleneck. Probably because of the repeated use of  `ggml_permute()` and `ggml_cont()`. Maybe rewriting it to avoid relying on permutations instead of merging this PR would be even better. Let me know if you'd prefer me giving it a shot.

leejet:
Perhaps https://github.com/leejet/stable-diffusion.cpp/pull/1080 is a simpler fix.

PR #1088, opened 2025-12-14
feat: give feedback about cfg values smaller than 1

wbruna:
The documentation and examples for many distilled models, such as [SDXL Turbo](https://huggingface.co/stabilityai/sdxl-turbo) and [Z-Image Turbo](https://huggingface.co/Tongyi-MAI/Z-Image-Turbo), specify a CFG scale of 0. So add a few warnings to nudge the user in the right direction.

Edit: removed the cfg cap, as per @stduhpf  suggestion.

### DIFF
--- a/stable-diffusion.cpp
+++ b/stable-diffusion.cpp
@@ -1484,7 +1484,7 @@ public:
         std::vector<int> skip_layers(guidance.slg.layers, guidance.slg.layers + guidance.slg.layer_count);
 
         float cfg_scale     = guidance.txt_cfg;
-        if (cfg_scale < 1) {
+        if (cfg_scale < 1.f) {
             if (cfg_scale == 0.f) {
                 // Diffusers follow the convention from the original paper
                 // (https://arxiv.org/abs/2207.12598v1), so many distilled model docs


stduhpf:
> the Diffusers implementation treats any value lesser than 1 the same as 1

Actually, that's not really it. The diffusers implementation uses CFG literally like it was proposed in the original paper (https://arxiv.org/abs/2207.12598v1), with CFG=0 meaning pure conditioned and CFG=-1 meaning unconditioned. It's effectively just an offset of -1 compared to every other implementation.

wbruna:
That's... unfortunate. So maybe a stronger warning would be better instead.

This was motivated by a Koboldcpp issue (LostRuins/koboldcpp#1880), which mentioned specifically the Z-Image Turbo page. But I believe most GUIs follow the "1 means pure conditioned" convention instead?

stduhpf:
I'm all for warning the user just in case they set it to such a value by mistake, but in my opinion, overriding the values set by the user is not good behavior.

The user might have set CFG to 0 on purpose to test unconditioned behavior of the model. Or set it to a negative number to see what the negative prompt does.

I say, let the user play with weird parameters values if he wants to, as long as it's not causing crashes.

leejet:
I think having so many log warnings in the code about the `cfg_scale` value is a bit annoying. Perhaps it would be sufficient to warn only when `cfg_scale < 1.f`, since that’s when it may lead to unexpected results.

PR #1098, opened 2025-12-15
feat: add KL Optimal scheduler

daniandtheweb:
This PR adds support for the KL Optimal scheduler.
I've taken inspiration from the Python implementation in ComfyUI and stable-diffusion-webui.

The resulting images seem correct. If the positioning of the scheduler looks good (after smoothstep and before lcm), I think this could be ready to review.


wbruna:
LGTM, and seems to be working fine.

> If the positioning of the scheduler looks good (after smoothstep and before lcm), I think this could be ready to review.

I don't know if that order matters much. When I implemented the LCM scheduler, I simply placed it at the end of the list.

daniandtheweb:
I was actually thinking about the order of the schedulers based on importance: of course the first would be discrete and karras as they're historically the most used and last should be LCM as it's more like a niche scheduler for a single specific application. I'm really not sure where kl_optimal should be placed compared to all the others, but if it's okay like this I think the PR is ready.

PR #1099, opened 2025-12-15
Add web server with Gradio UI, gallery, and OpenAI API hooks

taltoris:
## Description
Added a web server with Gradio UI for easier interaction with stable-diffusion.cpp

## Features
- Gradio-based web interface for user-friendly image generation
- Image gallery functionality to view and manage generated images
- OpenAI-compatible API endpoints for integration with other tools
- Docker Compose setup for easy deployment

## Usage
```bash
docker-compose up
# Access web UI at http://localhost:8000
```

## Testing
Open
http://localhost:8000 for the gradio UI
http://localhost:8000/gallery for the gallery

And you can use http://localhost:8000/v1 for OpenAI-api stuff.

## Dependencies
- See `server/requirements.txt` for Python dependencies
- Requires CUDA 12.6+ compatible GPU


wbruna:
To me, it's not clear what would be gained by integrating this implementation, instead of running it as a separate project, since it just calls the sd binary.

It'd also be a second implementation for the web API, so we'd need to maintain a second, independent implementation for the same functionality. An integrated frontend should just launch the server binary, and call its API. That would also avoid the latency from reloading the models.

And frankly, requiring CUDA for an official frontend to one of the few projects that have good support for other platforms sounds plain wrong. Especially because there seems to be no reason for hardcoding that dependency.

Apart from that, I see a few technical problems with the PR itself:

* it overwrites our existing, unrelated top-level `Dockerfile`
* I see support for `clip`, but not for the pair `clip_l`and `clip_g`, so no support for SD3? Why?
* It is apparently calling the upscaler mode with the diffusion model
* the `server` directory doesn't match the approach of the other utilities; something like `examples/gradio-server` would fit better with the existing code

SkutteOleg:
Vibe coding and vibe posting has occured

taltoris:
> Vibe coding and vibe posting has occured

Correct, it was vibe coded. I have a lot to learn.

But your critiques are fair. I have mostly tested it with Z_image, and it has helped my workflow significantly.

Running it as a separate service would allow it to run without cuda. True.

leejet:
Thank you for your contribution. However, I think it would be better to keep this as a separate repository. You can put a link to the corresponding repository in the README.md of sd.cpp.

taltoris:
Ok, thank you for the valuable feedback, I will make it a separate repo that just calls sd.

Issue #1092, opened 2025-12-15
Are Flux1-dev Lora's trained with kohya sd-scripts usable with stable-diffusion.cpp?


ddpasa:
The reason I ask is because I'm getting very different results compared to SwarmUI (with a Comfy backend).

As a followup, this lora also trains the t5 and clip encoders, is that still ok?

stduhpf:
I think they should work. Are there any strange messages in the logs?

ddpasa:
> I think they should work. Are there any strange messages in the logs?

I don't get any errors but the image is practically the same as before.

stduhpf:
Also have you tried playing with the --lora-apply-mode parameter to see how it affects the output?

ddpasa:
> Also have you tried playing with the --lora-apply-mode parameter to see how it affects the output?

I tried auto and at_runtime and got identical results.  Any other settings I should try?

Is there a check against a key mismatch?  For example any keys in the Lora that are not used because they don't match the model?

stduhpf:
> Is there a check against a key mismatch? For example any keys in the Lora that are not used because they don't match the model?

There is, and it should be displayed in the logs if it happens to be the case.

ddpasa:
> > Is there a check against a key mismatch? For example any keys in the Lora that are not used because they don't match the model?
> 
> There is, and it should be displayed in the logs if it happens to be the case.

Ok, scanning the logs I don't see any lora specific errors.  Here is the relevant part of the log:

```
[INFO ] model.cpp:1580 - loading tensors completed, taking 1.41s (process: 0.00s, read: 0.39s, memcpy: 0.00s, convert: 0.00s, copy_to_backend: 0.29s)
[DEBUG] stable-diffusion.cpp:720  - finished loaded file
[INFO ] stable-diffusion.cpp:792  - total params memory size = 21661.05MB (VRAM 21661.05MB, RAM 0.00MB): text_encoders 9318.83MB(VRAM), diffusion_model 12247.64MB(VRAM), vae 94.57MB(VRAM), controlnet 0.00MB(VRAM), pmid 0.00MB(VRAM)
[INFO ] stable-diffusion.cpp:865  - running in Flux FLOW mode
[DEBUG] stable-diffusion.cpp:3163 - generate_image 1024x1024
[DEBUG] stable-diffusion.cpp:1143 - lora /StableDiffusion/models/flux/lora/flux1_lora_v3.safetensors:1.00
[...]

PR #1103, opened 2025-12-16
feat: add support for underline style lora of flux

leejet:


ddpasa:
This improves https://github.com/leejet/stable-diffusion.cpp/issues/1092 .  Text encoders are still not completely loaded, but it looks like the transformer is fine after this.

leejet:
@ddpasa This update should make the LoRA for text_encoders work. However, I couldn’t find a corresponding LoRA to test it with. If you could upload a LoRA to HuggingFace or Civitai for me to download and test, I would really appreciate it.

ddpasa:
> @ddpasa This update should make the LoRA for text_encoders work. However, I couldn’t find a corresponding LoRA to test it with. If you could upload a LoRA to HuggingFace or Civitai for me to download and test, I would really appreciate it.

Thanks @leejet , I uploaded an example Lora here: https://huggingface.co/ddhfacc/lora_debug/tree/main  It has T5 and Clip layers as well.

Pulling the additional commit did not reduce the number of unused lora layers for me.

leejet:
@ddpasa Everything should be ok now.

ddpasa:
> @ddpasa Everything should be ok now.

I think so, testing now.  I don't see similar errors in the log after 6bff0d99f14f2e2bcb46a7c4f46e760870f255a5.  I'll generate a few and compare with Comfy.

ddpasa:
I generated multiple images, as far as I can tell, it works great now.  Thanks for fixing this!

PR #1108, opened 2025-12-17
Document usage of tae for VRAM reduction using wan

CarlGao4:
Added details on using tae to reduce VRAM requirements with links to relevant safetensors


Green-Sky:
It will likely only be slower until upstream merges the convolution kernels for cuda. For vulkan it should already be faster.

CarlGao4:
Do we need to mention CISC for reviewing that PR?

wbruna:
I think it'd be enough to mention that `--vae-conv-direct` "could be slower", since we don't know what will happen on every backend out there.

And maybe these docs should point to a (renamed and updated) `docs/taesd.md` instead?

PR #1110, opened 2025-12-17
make Wan2.2 5B FLF2V work

leejet:
I’ve never been able to get good results, and I’m not sure whether there’s a problem somewhere.


stduhpf:
> I’ve never been able to get good results, and I’m not sure whether there’s a problem somewhere.

Depending on how bad the results are, I'd say this might be expected. The wan2.2 5B TI2V model was never trained to support FLF2V. It kind of works, but it would probably need some training on this specific use case to be good at it.

PR #1113, opened 2025-12-18
Always upload artifacts even not releasing

CarlGao4:
So it would be easier to test PR updates locally

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -65,7 +65,6 @@ jobs:
 
       - name: Get commit hash
         id: commit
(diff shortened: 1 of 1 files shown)


leejet:
I don’t think this change is really necessary.

For PRs or temporary testing scenarios, the builds are inherently unstable and not intended for distribution. In those cases, uploading artifacts doesn’t add much value. If someone needs to verify behavior or debug an issue, the better approach is still to pull the code and test it locally, rather than relying on CI-generated binaries.

Uploading artifacts unconditionally for these unstable or intermediate builds can also make the CI outputs noisy, increase maintenance overhead, and potentially lead to misuse of binaries that were never meant to be consumed. Keeping artifact uploads limited to releases or explicitly intended cases feels like a cleaner and more controlled approach.

PR #1117, opened 2025-12-19
Add png sequence output for vid_gen

CarlGao4:
Now with `-M vid_gen -o out_%04d.png`, output files will be like `out_0000.png` `out_0001.png`, etc.

### DIFF
--- a/examples/cli/main.cpp
+++ b/examples/cli/main.cpp
@@ -756,11 +756,11 @@ int main(int argc, const char* argv[]) {
             std::string final_image_path = format_frame_idx(final_output_path, cli_params.output_begin_idx + i);
             if (is_jpg) {
                 int write_ok = stbi_write_jpg(final_image_path.c_str(), results[i].width, results[i].height, results[i].channel,
-                                                results[i].data, 90, get_image_params(cli_params, ctx_params, gen_params, gen_params.seed + i).c_str());
+                                              results[i].data, 90, get_image_params(cli_params, ctx_params, gen_params, gen_params.seed + i).c_str());
                 LOG_INFO("save result JPEG image %d to '%s' (%s)", i, final_image_path.c_str(), write_ok == 0 ? "failure" : "success");
             } else {
                 int write_ok = stbi_write_png(final_image_path.c_str(), results[i].width, results[i].height, results[i].channel,
-                                                results[i].data, 0, get_image_params(cli_params, ctx_params, gen_params, gen_params.seed + i).c_str());
+                                              results[i].data, 0, get_image_params(cli_params, ctx_params, gen_params, gen_params.seed + i).c_str());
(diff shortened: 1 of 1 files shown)


stduhpf:
Might as well add support for jpg sequence, no?

CarlGao4:
> Might as well add support for jpg sequence, no?

OK

Besides, where shall I put this documentation to?

leejet:
You can put the description in the command-line description and update the README.

CarlGao4:
Now that setting idx of first frame is supported. The patterned file name will also be applied to multiple image output

examples/cli/main.cpp
```diff
@@ -751,7 +785,12 @@ int main(int argc, const char* argv[]) {
                 continue;
             }
             int write_ok;
-            std::string final_image_path = i > 0 ? base_path + "_" + std::to_string(i + 1) + file_ext : base_path + file_ext;
+            std::string final_image_path;
+            if (std::regex_search(cli_params.output_path, format_specifier_regex)) {
+                final_image_path = format_frame_idx(cli_params.output_path, cli_params.output_begin_idx + i);
+            } else {
+                final_image_path = i > 0 ? base_path + "_" + std::to_string(cli_params.output_begin_idx + i) + file_ext : base_path + file_ext;
```
wbruna: The original code applied an offset here, e.g.: `output.png` followed by `output_2.png` (a weird convention IMHO, but it should be kept for compatibility) 

I'm not so sure how that would interact with `output_begin_idx`, though. But I believe I'd expect that `output_begin_idx=1` would imply `output_1.png` as the first image.

examples/cli/main.cpp
```diff
@@ -62,6 +65,10 @@ struct SDCliParams {
              "--preview-interval",
              "interval in denoising steps between consecutive updates of the image preview file (default is 1, meaning updating at every step)",
              &preview_interval},
+            {"",
+             "--output-begin-idx",
+             "starting index for output image sequence (only used when outputting multiple images, default: 0)",
```
wbruna: The code doesn't seem to follow the 'only used when outputting multiple images' restriction? (not that I think it should: it could be occasionally useful for grouping unrelated gens, and the user can easily turn it off if they wish).

CarlGao4:
I used 0 as default first image index for ffmpeg compability, so I've reverted the change when not using the new file name pattern

examples/cli/main.cpp
```diff
@@ -711,26 +737,71 @@ int main(int argc, const char* argv[]) {
         file_ext = file_ext_lower = "";
         is_jpg                    = false;
     }
+    defined_sequence = std::regex_search(cli_params.output_path, format_specifier_regex);
 
     if (cli_params.mode == VID_GEN && num_results > 1) {
         std::string vid_output_path = cli_params.output_path;
-        if (file_ext_lower == ".png") {
-            vid_output_path = base_path + ".avi";
+        if (defined_sequence) {
```
leejet: I think this logic duplicates the generation of the sequential images below. It can be merged together to reuse the code.

PR #1120, opened 2025-12-20
Use Ninja on Windows to speed up build process

CarlGao4:
The current MSBuild does not support parallel compiling of CUDA. Switching to Ninja can speed up as GH Actions uses 2-core VM.

Besides, CUDA warnings 177 and 550 are also suppressed (declared but unused / set but unused) to clean up build logs


CarlGao4:
For build test, see https://github.com/CarlGao4/stable-diffusion.cpp/actions/runs/20395440140/job/58610265651

With the 2-core VM of GitHub Actions, build time could be cut down by half

CarlGao4:
@leejet could you please review this?

leejet:
LGTM. Thanks.

PR #1122, opened 2025-12-21
feat: do not convert tensor names by default in convert mode

leejet:


examples/cli/main.cpp
```diff
@@ -69,6 +70,10 @@ struct SDCliParams {
              "--canny",
              "apply canny preprocessor (edge detection)",
              true, &canny_preprocess},
+            {"",
+             "--convert-name",
+             "convert tensor name (for convert mode)",
+             true, &canny_preprocess},
```
Green-Sky, line 76: wrong var

leejet: fixed

PR #1123, opened 2025-12-21
[new feat] add --serve-html-path  param in server target

Liyulingyue:
My work is as follows: 
- add --serve-html-path <path> option. After starting the service, you can enter "127.0.0.1:8081" in the browser to access the following interface.

image

### DIFF
--- a/examples/server/README.md
+++ b/examples/server/README.md
@@ -135,5 +135,4 @@ Default Generation Options:
   --cache-preset                           cache-dit preset: 'slow'/'s', 'medium'/'m', 'fast'/'f', 'ultra'/'u'
   --scm-mask                               SCM steps mask for cache-dit: comma-separated 0/1 (e.g., "1,1,1,0,0,1,0,0,1,0") - 1=compute, 0=can cache
   --scm-policy                             SCM policy: 'dynamic' (default) or 'static'
-  --serve-html-path                        specify a path of an HTML file to enable a web-based interface
 ```


leejet:
I think it would be better to keep this as a separate repository, like many other UIs powered by sd.cpp.

Green-Sky:
> I think it would be better to keep this as a separate repository, like many other UIs powered by sd.cpp.

I second that.

It would also help if server had something like a `--serve-html-path <path>` option.

Liyulingyue:
> I think it would be better to keep this as a separate repository, like many other UIs powered by sd.cpp.

How about adding the width, height, and steps parameters? If these changes look good, I’d be happy to open a separate PR for them.

And I completely understand the strategy of separating frontend and backend services! That said, I’d like to suggest a simple HTML page (just basic API calling) to:
- Help new users test the server without writing frontend code
- Let developers quickly verify model behavior during debugging

It’d require minimal maintenance, just like FastAPI’s auto-docs or llama.cpp’s debug page.

leejet:
> How about adding the width, height, and steps parameters? 

At present, all of these parameters can be specified either in the prompt via `<sd_cpp_extra_args>(.*?)</sd_cpp_extra_args>` or through the default image generation parameters.

leejet:
> It would also help if server had something like a --serve-html-path <path> option.

I agree with this approach.

Liyulingyue:
> At present, all of these parameters can be specified either in the prompt via or through the default image generation parameters.`<sd_cpp_extra_args>(.*?)</sd_cpp_extra_args>`

Oh! Sorry for the redundant work.

Liyulingyue:
> I agree with this approach.

The code has been revised based on the comments. Currently, the front-end can be specified via a --serve-html-path option. Please review it.

leejet:
I think it’s better not to include the HTML file inside sd.cpp.

Liyulingyue:
> I think it’s better not to include the HTML file inside sd.cpp.

Removed.

Green-Sky:
I think a link to your html repo (or gist) in the server readme would be good. 

@leejet Also maybe allow passing a folder to serve would be good.

leejet:
> I think a link to your html repo (or gist) in the server readme would be good.
> 
> @leejet Also maybe allow passing a folder to serve would be good.

I think this can be merged for now. If this requirement truly arises later, we can make changes then.

Issue #1128, opened 2025-12-22
[Bug] FA causes core dump Vulkan backend


engrtipusultan:
### Git commit

master-437-30a9113, commit 30a9113

### Operating System & Version

Ubuntu 24.04

### GGML backends

Vulkan

### Command-line arguments used

./sd-cli --diffusion-model /home/tipu/AI/image/unet/z_image_turbo-Q8_0.gguf --vae /home/tipu/AI/image/vae/flux1vae.safetensors --llm /home/tipu/AI/image/text_encoders/Qwen3-4B-Instruct-2507-Q8_0.gguf --lora-model-dir /home/tipu/AI/image/loras/ -p "D3tailed4nime. This image portrays an ethereal, celestial figure seated gracefully on a crescent moon amidst a starry night sky. The character exudes an otherworldly beauty, with flowing silver-white hair that cascades over her shoulders, blending into the soft glow of the moon. She wears a delicate, off-shoulder gown that billows around her in layers of ruffled fabric, reminiscent of clouds or stardust, further enhancing her ethereal nature. The gown is intricately detailed, with shimmering, lace-like embellishments and a soft ivory hue that complements the glowing moon she sits on. The crescent moon itself appears textured, with golden, vine-like details curling along its edges, giving it an ornate and almost organic quality. The moon’s soft light reflects off her skin, making her appear radiant, as if she is the embodiment of the moonlight itself. Her expression is calm and serene, with a subtle, melancholic gaze directed into the distance, adding a sense of mystery. The surrounding background features a deep, cosmic night sky filled with faint stars, creating a dreamlike, celestial atmosphere. This character might represent a moon goddess, a celestial maiden, or a mystical spirit connected to the stars and the night, evoking a sense of tranquility, beauty, and cosmic wonder" --cfg-scale 1.0 -v -H 1080 -W 1920 --steps 7 --seed -1 --rng cpu --scheduler smoothstep --sampling-method euler --flow-shift 2 -o output_$(date +%Y-%m-%d_%H-%M-%S).png --vae-conv-direct --diffusion-fa

### Steps to reproduce

Run the above command on vulkan back-end on build 
 bash  ./sd-cli --version
stable-diffusion.cpp version master-437-30a9113, commit 30a9113

### What you expected to happen

I should be working it is working on 
bash  ./sd-cli --version
stable-diffusion.cpp version master-427-78e15bd, commit 78e15bd

### What actually happened

Core dump happened. 

[...]

PR #1132, opened 2025-12-23
Restore asymmetric padding in DownSampleBlock::forward

wwoodsTM:
I noticed that after commit 50ff966 that I was getting 8-pixel borders added to the top and left of all my inpainting outputs (with images of any dimensions). I experimented with different fixes until I found this one key line in DownSampleBlock::forward (`common.hpp`) that seems to be contributing to the 2x2x2 (8 px) offset / borders that I was seeing. I also attempted to keep circular padding in there while maintaining the original behavior overall. 

Now whether this is the ideal fix for all use cases I am not 100% sure, but I am hoping someone with more expertise can see if this is a valid fix as it has been working well so far.

Thank you.

### DIFF
--- a/common.hpp
+++ b/common.hpp
@@ -28,7 +28,13 @@ public:
         if (vae_downsample) {
             auto conv = std::dynamic_pointer_cast<Conv2d>(blocks["conv"]);
 
-            x = ggml_ext_pad(ctx->ggml_ctx, x, 1, 1, 0, 0, ctx->circular_x_enabled, ctx->circular_y_enabled);
+            // Commenting out (from commit 50ff966)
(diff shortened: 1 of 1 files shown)


wbruna:
I can't say if this is the right fix for inpainting, but at least #1128 (another image size related bug) still happens with this change.

leejet:
Thanks, but I think https://github.com/leejet/stable-diffusion.cpp/pull/1133  would be a better fix.

wwoodsTM:
> Thanks, but I think #1133 would be a better fix.

Agreed, I just tried the newest commit with this fix and it also resolves the bug. Idk if my little PR was even helpful at the end of the day lol, especially if this was already a known issue, but it was nice to learn a bit about the code base either way. Thanks to anybody who took a look or tested!

wwoodsTM:
Closing this since it was resolved by #1133

Issue #1130, opened 2025-12-23
[Bug] kl_optimal has inconsistent outputs


daniandtheweb:
### Git commit

98916e825686f8952126b2c54e1c4216eb418dc5

### Operating System & Version

Arch Linux

### GGML backends

Vulkan

### Command-line arguments used

./sd-cli -M img_gen -p "1girl, vivid colors, best quality, masterpiece," -n "bad quality, worst quality," --sampling-method euler_a --steps 20 -W 1024 -H 1024 --cfg-scale 5--model /home/daniandtheweb/Workspace/sd.cpp-webui/models/checkpoints/hassakuXLIllustrious_v32.safetensors --vae /home/daniandtheweb/Workspace/sd.cpp-webui/models/vae/sdxl_vae_fp16_fix.safetensors --scheduler kl_optimal --diffusion-fa --vae-conv-direct

### Steps to reproduce

Executing this specific model (for now I haven't been able to reproduce it on any other model).

### What you expected to happen

A normally denoised image.

### What actually happened

Black image.

I've already investigated the issue and the problem is on how the sigmas close to zero interact in the last step.
The current implementation is adapted from AUTOMATIC1111 and ComfyUI's (closer to comfyui's as the last step enforces a jump to sigma=0).

I currently have two different ideas to fix it and I'd like some feedback on them:
1. Change kl_optimal to stop at sigma_min (just like AUTOMATIC1111), however this apparently may have some downsides.
2. Add safeguards in the samplers to handle the cases where the sigmas would lead to incorrect calculations and a black image.

daniandtheweb:
Closed yesterday by mistake.

PR #1138, opened 2025-12-23
fix: avoid issues when sigma_min is close to 0

daniandtheweb:
This is a possible fix for https://github.com/leejet/stable-diffusion.cpp/issues/1130.

This PR creates a safeguard that detects whenever a sigma_min gets too close to zero and avoids the issues that may come with it (black images).
This also adds a fix and cleans up the KL Optimal scheduler function.

I've noticed this issue on the Karras and the KL Optimal schedulers with this same model: HassakuXL 3.2.

### DIFF
--- a/denoiser.hpp
+++ b/denoiser.hpp
@@ -351,11 +351,6 @@ struct SmoothStepScheduler : SigmaScheduler {
     }
 };
 
-/*
-* KL Optimal:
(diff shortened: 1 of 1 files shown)


daniandtheweb:
This still doesn't seem to be 100% reliable.

daniandtheweb:
This new change seems to be more reliable.

PR #1149, opened 2025-12-28
refactor: simplify logic for saving results

leejet:


examples/cli/main.cpp
```diff
@@ -370,6 +370,95 @@ std::string format_frame_idx(std::string pattern, int frame_idx) {
...
+        base_path.replace_extension();
+
+    std::string ext_lower = ext.string();
+    std::transform(ext_lower.begin(), ext_lower.end(), ext_lower.begin(), ::tolower);
+    bool is_jpg = (ext_lower == ".jpg" || ext_lower == ".jpeg" || ext_lower == ".jpe");
+
+    int output_begin_idx = cli_params.output_begin_idx;
+    if (output_begin_idx < 0) {
+        output_begin_idx = 0;
+    }
```
CarlGao4, line 407: In the previous change, for backward compatibility, when `%d` does not present in output file name, the begin idx would be 1. So this has been removed here?

https://github.com/leejet/stable-diffusion.cpp/pull/1117#discussion_r2637064678

PR #1157, opened 2026-01-01
feat: generate new random seed per request when seed is -1 (server)

Jay4242:
This PR ensures that each image generation request uses a unique random seed when the user specifies `seed: -1`. Previously, a seed of `-1` would result in deterministic behavior across requests. The changes introduce:

1. Inclusion of `<random>` header for random device usage.
2. Logic in `examples/server/main.cpp` to detect `seed == -1` and replace it with a freshly generated 64‑bit seed using `std::random_device`.

### DIFF
--- a/examples/server/main.cpp
+++ b/examples/server/main.cpp
@@ -7,6 +7,7 @@
 #include <mutex>
 #include <sstream>
 #include <vector>
+#include <random>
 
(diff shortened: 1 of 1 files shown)


Jay4242:
Closed & Deleted because this ended up being incorrect.  The same seed was still used for every image.

PR #1161, opened 2026-01-01
remove std::exit

piallai:
safer regarding memory allocation

Caution: indentation was not changed in 'main', for diff clarity sake


leejet:
Since sd-cli is a CLI-only program, using std::exit in argument parsing is reasonable and not a problem.

piallai:
Indeed. Actually std::exit has a radical behaviour (hard exit no matter what) that is not compatible with my fork. So it was just to propose a way to also make the repo safer in a general manner. But you're right, it does not change much the original cli behaviour.

PR #1164, opened 2026-01-02
ci: add workflows for linux vulkan build and docker images, update ubuntu base version, add vulkan dockerfile

rare-magma:
This PR:
1. sets up a workflow that builds a linux vulkan release
2. sets up a workflow that builds docker images and pushes them to the github container registry
3. updates the ubuntu base image version in the docker images
4. creates a new dockerfile for a vulkan image
5. copies both binaries to the docker images so they can be used as a server

### DIFF
--- a/docs/docker.md
+++ b/docs/docker.md
@@ -1,23 +1,39 @@
 # Docker
 
+## Run CLI
+
+```shell
+docker run --rm -v /path/to/models:/models -v /path/to/output/:/output ghcr.io/leejet/stable-diffusion.cpp:master [args...]
+# For example
+# docker run --rm -v ./models:/models -v ./build:/output ghcr.io/leejet/stable-diffusion.cpp:master -m /models/sd-v1-4.ckpt -p "a lovely cat" -v -o /output/output.png
+```
+
+## Run server
+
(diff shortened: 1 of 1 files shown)


leejet:
Could you share a sample build artifact?

rare-magma:
@leejet I'm testing it on my fork, I'll update the PR once I get it working

rare-magma:
@leejet I've tested it on my fork here:
https://github.com/rare-magma/stable-diffusion.cpp/actions/runs/20694352266
It seems to work fine, there's some failures in workflows that I haven't changed in this PR.
I tested the vulkan docker image and it works fine on my machine:
https://github.com/rare-magma/stable-diffusion.cpp/pkgs/container/stable-diffusion.cpp
Here's the artifact of the new ubuntu vulkan build:
https://github.com/rare-magma/stable-diffusion.cpp/actions/runs/20694352266/artifacts/5017896149

A more robust solution for the future could be reusing the workflows that are setup in the llama.cpp repository as they seem quite complete: https://github.com/ggml-org/llama.cpp/blob/master/.github/workflows/docker.yml

PR #1166, opened 2026-01-03
feat: add example Python web client

wbruna:
Adds a basic Python command-line client for the sd-server API.

I've tried to keep it compatible with the `sd-cli` command line, to make it easier to use from user scripts already expecting that interface; so unsupported arguments are ignored, and the server URL can be specified with an environment variable. It uses supported API fields when available, and propagates all other server supported flags via `sd_cpp_extra_args`.

It's mostly useful for testing the server itself, and as reference code for third-party implementations.

### DIFF
--- a/examples/server/client.py
+++ b/examples/server/client.py
@@ -22,6 +22,37 @@ except ImportError:
 
 apis = ['sdapi', 'openai', 'openai-module']
 
+def extract_lora_tags(prompt):
+
+    pattern = r'<lora:([^:>]+):([^>]+)>'
+    lora_data = []
+
+    matches = list(re.finditer(pattern, prompt))
+
+    for match in matches:
+        raw_path = match.group(1)
+        raw_mul = match.group(2)
+        try:
+            mul = float(raw_mul)
+        except ValueError:
+            continue
+
+        is_high_noise = False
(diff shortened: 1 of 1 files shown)


leejet:
In fact, since the API is OpenAI-compatible, you can directly use the OpenAI Python SDK to call it. Below is a simple example.

```python
from openai import OpenAI
import base64

client = OpenAI(
    api_key="local-api-key",
    base_url="http://127.0.0.1:1234/v1"
)

result = client.images.generate(
    model="sd.cpp",
    prompt="a lovey cat",
    size="512x512"
)

image_base64 = result.data[0].b64_json
image_bytes = base64.b64decode(image_base64)

with open("output.png", "wb") as f:
    f.write(image_bytes)

print("done")
```

wbruna:
I've intentionally avoided extra dependencies, to make the script available for any base Python install.

I could add a test for the OpenAI module, and keep the current code as a fallback.

leejet:
Thanks for the PR. But the existing APIs are already simple, well documented, and demonstrated by the frontend implementation. This script also adds a fairly large client wrapper that duplicates CLI/API parameter handling and can easily become outdated, so I don’t think the extra maintenance cost is justified. I’m going to close this for now.

PR #1169, opened 2026-01-03
feat: prioritize gguf and safetensors formats for embeddings and LoRAs

wbruna:
Models available as multiple file types were likely converted from one another, so we should prioritize native, safer and more efficient formats.

### DIFF
--- a/examples/common/common.hpp
+++ b/examples/common/common.hpp
@@ -809,7 +809,7 @@ struct SDContextParams {
     }
 
     void build_embedding_map() {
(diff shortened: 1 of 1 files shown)


leejet:
I’m not entirely sure whether this is necessary here. Generally, GGUF files include the corresponding quantization format in their filenames, so they differ from the original version.

wbruna:
My main concern is with the current .pt priority. I agree it doesn't matter that much for .gguf versus .safetensors (as I also don't see a strong reason for preferring .safetensors over .gguf, either; but I can adapt the PR if there is one).

PR #1173, opened 2026-01-04
feat: add sd-server API support for steps, sampler and scheduler

wbruna:
This is on top of #1166 just to make it easier to test.

The step numbers should probably be limited in some way?

### DIFF
--- a/examples/server/main.cpp
+++ b/examples/server/main.cpp
@@ -420,6 +420,9 @@ int main(int argc, const char** argv) {
                 return;
             }
 
+            if (gen_params.sample_params.sample_steps > 100)
+                gen_params.sample_params.sample_steps = 100;
+
             if (!gen_params.process_and_check(IMG_GEN, "")) {
                 res.status = 400;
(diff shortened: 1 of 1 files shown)


leejet:
I believe it would be better to exclude the Python web client from this PR and keep the scope limited to server-related changes.

leejet:
> The step numbers should probably be limited in some way?

I think we could limit the number of steps to a maximum value, for example, 100.

Green-Sky:
> > The step numbers should probably be limited in some way?
> 
> I think we could limit the number of steps to a maximum value, for example, 100.

For reference, I have seen distillations that use 150steps as reference.

wbruna:
> For reference, I have seen distillations that use 150steps as reference.

Since this is mostly to avoid excessive processing, a better way could be combining a really high hard limit with a simple heuristic for cost. Something like number of steps * image area * batch size, first capping batch size, then the step count. (out of scope for this PR, of course)

Issue #603, opened 2025-02-26
Unable to inference using Segmind Tiny SD model


Rehaman1429:
Hi there,

I am trying to run Segmind's Distilled Diffusion model (segmind/tiny-sd) from Hugging Face on my machine with the following specifications:

- Processor: 12th Gen Intel(R) Core(TM) i5-1235U 1.30 GHz
- Installed RAM: 16.0 GB (15.7 GB usable)
- System Type: 64-bit operating system, x64-based processor

I successfully converted the model to a single checkpoint file (.safetensors, .ckpt) as shown below:

![Image](https://github.com/user-attachments/assets/3ff71e86-8a85-4d2b-beaf-6d15397543e7)

Then, I used the convert function in stable-diffusion.cpp to convert the .safetensors file to the GGUF format. While the conversion completed without any errors, the resulting GGUF file is significantly smaller than the original model, as shown here:

![Image](https://github.com/user-attachments/assets/fcad54db-117d-4173-922e-266915ec31e0)

![Image](https://github.com/user-attachments/assets/f3c660be-1c71-491a-911c-d17d9fbd483a)

When I attempt to run inference with the converted GGUF file, I encounter the following error:

![Image](https://github.com/user-attachments/assets/a5f18a6a-2bda-4534-805b-bc2be17b6bd4)

It appears that the issue may lie in the conversion process to GGUF, possibly due to the fact that the model in question is a distilled version (Tiny-SD). I am wondering if anyone has worked with distilled models in this context and found a fix for this issue.

Any insights or suggestions would be greatly appreciated.

Thank you for your time and assistance!

stduhpf:
You don't need to convert the `.safetensors` to gguf, sdcpp supports loading safetensors files directly. You could try loading it and see if you get the same error? (I'm pretty sure you would)

I think there are two problems here: 
- the model file doesn't contain the text encoder, which would explain why It's unable to detect the version. If it was included, I believe it would be detected as a sd1.x model. 
- Even if the text encoder was included, there would be another problem causing a crash when loading, because It would try loading it as a standard sd1.x model, even though the architecture is actually fairly different.

So to "fix" this you would first need to make sure the text_encoder is included in the model file, and then you'd still have to add support for this model (I think this would  require making changes to `model.h`, `model.cpp`, and `unet .cpp` mostly).

Rehaman1429:
Thanks for your reply @stduhpf ,

[...]

PR #1180, opened 2026-01-07
feat: support for SDXS-512 model

akleine:
Once I wrote "three TinySD models should be enough"  ( #939 ),  but I changed my mind on user's request in December 2025 ( #603 ) and because SDXS-512 is so **incredible fast**.  (That makes SDXS very handy for my sdcpp-on-android project.)

The main challenge for me was that SDXS does not use AutoEncoderKL as primary VAE (as most else)  but AutoEncoderTiny,   see also  https://huggingface.co/IDKiro/sdxs-512-dreamshaper .  I also hope that including SDXS into sd.cpp will convince @IDkiro to release into the public his SDXS-1024  one day.

### DIFF
--- a/stable-diffusion.cpp
+++ b/stable-diffusion.cpp
@@ -31,6 +31,7 @@ const char* model_version_to_str[] = {
     "SD 2.x",
     "SD 2.x Inpaint",
     "SD 2.x Tiny UNet",
+    "SDXS",
     "SDXL",
     "SDXL Inpaint",
     "SDXL Instruct-Pix2Pix",
--- a/stable-diffusion.cpp
+++ b/stable-diffusion.cpp
@@ -408,6 +408,11 @@ public:
             vae_decode_only = false;
         }
 
+        bool tae_preview_only = sd_ctx_params->tae_preview_only;
+        if (version == VERSION_SDXS) {
+            tae_preview_only = false;
+        }
+
         if (sd_ctx_params->circular_x || sd_ctx_params->circular_y) {
             LOG_INFO("Using circular padding for convolutions");
         }
@@ -592,7 +597,7 @@ public:
                 vae_backend = backend;
             }
 
-            if (!(use_tiny_autoencoder || version == VERSION_SDXS) || sd_ctx_params->tae_preview_only) {
+            if (!(use_tiny_autoencoder || version == VERSION_SDXS) || tae_preview_only) {
                 if (sd_version_is_wan(version) || sd_version_is_qwen_image(version)) {
                     first_stage_model = std::make_shared<WAN::WanVAERunner>(vae_backend,
                                                                             offload_params_to_cpu,
@@ -786,7 +791,7 @@ public:
                 unet_params_mem_size += high_noise_diffusion_model->get_params_buffer_size();
             }
             size_t vae_params_mem_size = 0;
-            if (!(use_tiny_autoencoder || version == VERSION_SDXS) || sd_ctx_params->tae_preview_only) {
+            if (!(use_tiny_autoencoder || version == VERSION_SDXS) || tae_preview_only) {
                 vae_params_mem_size = first_stage_model->get_params_buffer_size();
             }
             if (use_tiny_autoencoder || version == VERSION_SDXS) {
@@ -950,7 +955,7 @@ public:
         }
 
         ggml_free(ctx);
-        use_tiny_autoencoder = use_tiny_autoencoder && !sd_ctx_params->tae_preview_only;
+        use_tiny_autoencoder = use_tiny_autoencoder && !tae_preview_only;
         return true;
     }
 
--- a/stable-diffusion.cpp
+++ b/stable-diffusion.cpp
@@ -116,8 +116,7 @@ public:
     std::shared_ptr<DiffusionModel> diffusion_model;
     std::shared_ptr<DiffusionModel> high_noise_diffusion_model;
     std::shared_ptr<VAE> first_stage_model = nullptr;
-    std::shared_ptr<TinyAutoEncoder> first_stage_model_tiny = nullptr;
-    std::shared_ptr<TinyAutoEncoder> tae_first_stage;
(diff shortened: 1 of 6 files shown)


stable-diffusion.cpp
```diff
@@ -114,7 +115,8 @@ class StableDiffusionGGML {
     std::shared_ptr<FrozenCLIPVisionEmbedder> clip_vision;  // for svd or wan2.1 i2v
     std::shared_ptr<DiffusionModel> diffusion_model;
     std::shared_ptr<DiffusionModel> high_noise_diffusion_model;
-    std::shared_ptr<VAE> first_stage_model;
+    std::shared_ptr<VAE> first_stage_model = nullptr;
+    std::shared_ptr<TinyAutoEncoder> first_stage_model_tiny = nullptr;
     std::shared_ptr<TinyAutoEncoder> tae_first_stage;
```
stduhpf: Why not just use `tae_first_stage` ?

stduhpf: Also even though it's recommended to use with TAESD by default, it should work just fine with sd1.x KL-F8 VAE, just slower.

akleine:
@stduhpf  wrote:
> Also even though it's recommended to use with TAESD by default, it should work just fine with sd1.x KL-F8 VAE, just slower.

Yes, but one needs additional options, it is slower and the output is not  in the same quality as intended by IDkiro.
BTW, in the second commit I have included this example using `--taesd` :
```python
~/stable-diffusion.cpp/build/bin/sd-cli -m sdxs  -p "portrait of a lovely cat" \
  --cfg-scale 1 --steps 1  --taesd  sdxs/vae/diffusion_pytorch_model.safetensors
```


I simply like the idea of using the provided models just "_out of the box_"   without more mandatory options than really needed.   That was the reason to continue after the second commit (i.e. the U-Net part).

Anyway,  a big **THANK YOU**  for code your revision.  
And you are right,   better use only one variable here,   making all very short and clear now.   
So the patch code size just melted down like ice in the sunshine ...

leejet:
Could you provide sdxs.safetensors for testing?

akleine:
..and here it is: 

```
wget https://huggingface.co/akleine/sdxs-512/resolve/main/sdxs.safetensors
```

use it via
```
sd-cli -m sdxs.safetensors -p "portrait of a lovely cat" -v --cfg-scale 1 --steps 1
```

akleine:
Thanks for merging this PR into your master.

I think now we have extended once more the range of  sd.cpp further towards the lower end. 
So I just tried sd.cpp on my old Raspberry 400 (4GB, no GPU, built in 2020). 
Using SDXS it takes 17 seconds to create some more pictures of nice cats & dogs etc...

@leejet : Thank you very much!

LostRuins:
Quality is a little (a LOT) on the scuffed side but by god it is fast.

image

won't work on anything more than 1 step or 1 cfg.

LostRuins:
Is there any way I can get this model to generate different images for the same prompt? seems like no matter what I do it always generates nearly the same thing

akleine:
I can confirm that, but it also happens with  PyTorch/Diffusers.

Possibly a the disadvantage of the model ?
Don't know whether @IDKiro   reads here,  but he could explain this.

stduhpf:
> Is there any way I can get this model to generate different images for the same prompt? seems like no matter what I do it always generates nearly the same thing

ZIT has a simillar issue to a lesser extent. I'm guesssing the 1 step distillation makes it less sensitive to the original noise.

akleine:
@LostRuins :   I have just added sdxs 0.9 in my fork.  Do you want to try that ?


https://github.com/akleine/stable-diffusion.cpp/tree/sdxs-09

Pls note:  even though the name "SDXS-09" is similar to "SDXS", it is completely different from SDXS.
The image output here is  **much more depending from seed.**   It won't generate always the same thing.

LostRuins:
I do want to try it! I'll give it a shot. If it's good, it would be even better if you could PR it here.

Will reply here when I have some time to try it

Edit: I can't find the sdxs09 source model.

akleine:
The model:
https://huggingface.co/akleine/sdxs-09/resolve/main/sdxs09.safetensors

The PR:
https://github.com/leejet/stable-diffusion.cpp/pull/1254

Green-Sky:
I figured out how to run sd.cpp server on free (meh) huggingface space.

https://huggingface.co/spaces/Green-Sky/sd.cpp

It is this model, with the new webui from #1207 .

akleine:
Thanks for your work ! ❤️ 

Due to lack of time, I only briefly looked into it.  But one remark here:

- If the model is hard coded to SDXS09, it would be better to specify some **usable defaults**
   for steps and guidance, something  like   `     --cfg-scale 1 --steps 1    `

Otherwise, new users will have to wait a very long time, 
and the result could be poor. They will get a false impression of  sd.cpp

Thanks

Green-Sky:
>     * If the model is hard coded to SDXS09, it would be better to specify some **usable defaults**
>       for steps and guidance, something  like   `    --cfg-scale 1 --steps 1   `
> 
> 
> Otherwise, new users will have to wait a very long time, and the result could be poor. They will get a false impression of sd.cpp

I already do that, but sadly the ui currently has its own defaults, not influenced by the server. (see discussion over at the webui pr)

Thank you again for adding support for such a small model. :)

akleine:
Hi, perhaps you know it already: 
Now also **SDXS-09** is supported.

BTW, I just tried your nice HF space:



What do you think about fixing steps and cfg like:
if (model == SDXS) { steps = 1; cfg = 1; }
Similar ( cfg=1; ) could be useful if using some LCM LoRA stuff ?

Green-Sky:
@akleine I updated the hf space and it now uses the new sd.cpp webui, which seems to supports defaults values for cfg and steps. :rocket:

akleine:
Just tried your hf space again, 
that is really a very good work !! 👍👍👍

"A speedcar on the highway, golden hour, sun down" was done within __19 seconds__ :

PR #1184, opened 2026-01-09
Feat: Select backend devices via arg ( +add RPC backend support)

### LINKED ISSUE #1116: [Feature] Supports specifying backend
### Feature Summary

Supports specifying backend

### Detailed Description

Supports specifying backend.
Like `-dev Vulkan0` and `-dev ROCm0`

### Alternatives you considered

_No response_

### Additional context

_No response_

stduhpf:
The main goal of this PR is to improve user experience in multi-gpu setups, allowing to chose which model part gets sent to which device.

Cli changes: 
- Adds the `--main-backend-device [device_name]` argument to set the default backend
- remove `--clip-on-cpu`, `--vae-on-cpu` and `--control-net-cpu` arguments 
- replace them respectively with the new `--clip_backend_device [device_name]`, `--vae-backend-device [device_name]`, `--control-net-backend-device [device_name]` arguments 
- add the `--diffusion_backend_device` (control the device used for the diffusion/flow models) and the `--tae-backend-device`
- add `--upscaler-backend-device`, `--photomaker-backend-device`, and `--vision-backend-device`
- add `--list-devices` argument to print the list of available ggml devices and exit.
- add `--rpc` argument to connect to a compatible GGML rpc server

C API changes (stable-diffusion.h):
-  Change the content of the `sd_ctx_params_t` struct. 
- `void list_backends_to_buffer(char* buffer, size_t buffer_size)` to write the details of the available buffers to a null-terminated char array. Devices are separated by newline characters (`\n`), and the name and description of the device are separated by `\t` character.
- `size_t backend_list_size()` to get the size of the buffer needed for void list_backends_to_buffer
- `void add_rpc_device(const char* address);` connect to a ggml RPC backend ([from llama.cpp](https://github.com/ggml-org/llama.cpp/tree/master/tools/rpc))

The default device selection should now consistently prioritize discrete GPUs over iGPUs.  

For example if you want to run the text encoders on CPU, you'd need to use `--clip_backend_device CPU` instead of `--clip-on-cpu`

TODO:
  - Clean up logs
  
Important: to use RPC, you need to add `-DSD_RPC=ON` to the build. Additionally it requires ~~either sd.cpp to be built with `-DSD_USE_SYSTEM_GGML` flag (I haven't tested that one), or~~ the RPC server to be built with `-DCMAKE_C_FLAGS="-DGGML_MAX_NAME=128" -DCMAKE_CXX_FLAGS="-DGGML_MAX_NAME=128"` ([default is 64](https://github.com/ggml-org/ggml/pull/682))
  
  Fixes #1116

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -77,6 +77,12 @@ if(SD_MUSA)
     add_definitions(-DSD_USE_CUDA)
 endif()
 
+if (SD_RPC)
+    message("-- Use RPC as backend stable-diffusion")
+    set(GGML_RPC ON)
+    add_definitions(-DSD_USE_RPC)
+endif ()
+
 set(SD_LIB stable-diffusion)
 
 file(GLOB SD_LIB_SOURCES
--- a/docs/rpc.md
+++ b/docs/rpc.md
@@ -1,0 +1,202 @@
+# Building and Using the RPC Server with `stable-diffusion.cpp`
+
+This guide covers how to build a version of the RPC server from `llama.cpp` that is compatible with your version of `stable-diffusion.cpp` to manage multi-backends setups. RPC allows you to offload specific model components to a remote server.
+
+> **Note on Model Location:** The model files (e.g., `.safetensors` or `.gguf`) remain on the **Client** machine. The client parses the file and transmits the necessary tensor data and computational graphs to the server. The server does not need to store the model files locally.
+
+## 1. Building `stable-diffusion.cpp` with RPC client
+
+First, you should build the client application from source. It requires `GGML_RPC=ON` to include the RPC backend to your client.
+```bash
+mkdir build
+cd build
+cmake .. \
+    -DGGML_RPC=ON \
+    # Add other build flags here (e.g., -DSD_VULKAN=ON)
+cmake --build . --config Release -j $(nproc)
+```
+
+> **Note:** Ensure you add the other flags you would normally use (e.g., `-DSD_VULKAN=ON`, `-DSD_CUDA=ON`, `-DSD_HIPBLAS=ON`, or `-DGGML_METAL=ON`), for more information about building `stable-diffusion.cpp` from source, please refer to the `build.md` documentation.
+
+## 2. Ensure `llama.cpp` is at the correct commit
+
+`stable-diffusion.cpp`'s RPC client is designed to work with a specific version of `llama.cpp` (compatible with the `ggml` submodule) to ensure API compatibility. The commit hash for `llama.cpp` is stored in `ggml/scripts/sync-llama.last`.
+
+> **Start from Root:** Perform these steps from the root of your `stable-diffusion.cpp` directory.
+
+1.  Read the target commit hash from the submodule tracker:
+    ```bash
+    # Linux / WSL / MacOS
+    HASH=$(cat ggml/scripts/sync-llama.last)
+
+    # Windows (PowerShell)
+    $HASH = Get-Content -Path "ggml\scripts\sync-llama.last"
+    ```
+
+2.  Clone `llama.cpp` at the target commit .
+    ```bash
+    git clone https://github.com/ggml-org/llama.cpp.git
+    cd llama.cpp
+    git checkout $HASH
+    ```
+
+To save on download time and storage, you can use a shallow clone to download only the target commit: 
+    ```bash
+    mkdir -p llama.cpp
+    cd llama.cpp
+    git init
+    git remote add origin https://github.com/ggml-org/llama.cpp.git
+    git fetch --depth 1 origin $HASH
+    git checkout FETCH_HEAD
+    ```
+
+## 3. Build `llama.cpp` (RPC Server)
+
+The RPC server acts as the worker. You must explicitly enable the **backend** (the hardware interface, such as CUDA for Nvidia, Metal for Apple Silicon, or Vulkan) when building, otherwise the server will default to using only the CPU.
+
+To find the correct flags, refer to the official documentation for the `llama.cpp` repository.
+
+> **Crucial:** You must include the compiler flags required to satisfy the API compatibility with `stable-diffusion.cpp` (`-DGGML_MAX_NAME=128`). Without this flag, `GGML_MAX_NAME` will default to `64` for the server, and data transfers between the client and server will fail. Of course, `-DGGML_RPC` must also be enabled.
+>
+> I recommend disabling the `LLAMA_CURL` flag to avoid unnecessary dependencies, and disabling shared library builds to avoid potential conflicts.
+
+> **Build Target:** We are specifically building the `rpc-server` target. This prevents the build system from compiling the entire `llama.cpp` suite (like `llama-cli`), making the build significantly faster.
+
+### Linux / WSL (Vulkan)
+```bash
+mkdir build
+cd build
+cmake .. -DGGML_RPC=ON \
+    -DGGML_VULKAN=ON \        # Ensure backend is enabled
+    -DGGML_BUILD_SHARED_LIBS=OFF \
+    -DLLAMA_CURL=OFF \
+    -DCMAKE_C_FLAGS=-DGGML_MAX_NAME=128 \
+    -DCMAKE_CXX_FLAGS=-DGGML_MAX_NAME=128
+cmake --build . --config Release --target rpc-server -j $(nproc)
+```
+
+### macOS (Metal)
+```bash
+mkdir build
+cd build
+cmake .. -DGGML_RPC=ON \
+    -DGGML_METAL=ON \
+    -DGGML_BUILD_SHARED_LIBS=OFF \
+    -DLLAMA_CURL=OFF \
+    -DCMAKE_C_FLAGS=-DGGML_MAX_NAME=128 \
+    -DCMAKE_CXX_FLAGS=-DGGML_MAX_NAME=128
+cmake --build . --config Release --target rpc-server
+```
+
+### Windows (Visual Studio 2022, Vulkan)
+```powershell
+mkdir build
+cd build
+cmake .. -G "Visual Studio 17 2022" -A x64 `
+    -DGGML_RPC=ON `
+    -DGGML_VULKAN=ON `
+    -DGGML_BUILD_SHARED_LIBS=OFF `
+    -DLLAMA_CURL=OFF `
+    -DCMAKE_C_FLAGS=-DGGML_MAX_NAME=128 `
+    -DCMAKE_CXX_FLAGS=-DGGML_MAX_NAME=128
+cmake --build . --config Release --target rpc-server
+```
+
+## 4. Usage
+
+Once both applications are built, you can run the server and the client to manage your GPU allocation.
+
+### Step A: Run the RPC Server
+
+Start the server. It listens for connections on the default address (usually `localhost:50052`). If your server is on a different machine, ensure the server binds to the correct interface and your firewall allows the connection.
+
+**On the Server :**
+If running on the same machine, you can use the default address:
+```bash
+./rpc-server
+```
+If you want to allow connections from other machines on the network:
+```bash
+./rpc-server --host 0.0.0.0
+```
+
+> **Security Warning:** The RPC server does not currently support authentication or encryption. **Only run the server on trusted local networks**. Never expose the RPC server directly to the open internet.
+
+> **Drivers & Hardware:** Ensure the Server machine has the necessary drivers installed and functional (e.g., Nvidia Drivers for CUDA, Vulkan SDK, or Metal). If no devices are found, the server will simply fallback to CPU usage.
+
+### Step B: Check if the client is able to connect to the server and see the available devices
+
+We're assuming the server is running on your local machine, and listening on the default port `50052`. If it's running on a different machine, you can replace `localhost` with the IP address of the server.
+
+**On the Client:**
+```bash
+./sd-cli --rpc localhost:50052 --list-devices
+```
+If the server is running and the client is able to connect, you should see `RPC0    localhost:50052` in the list of devices.
+
+Example output: 
+(Client built without GPU acceleration, two GPUs available on the server)
+```
+List of available GGML devices:
+Name    Description
+-------------------
+CPU     AMD Ryzen 9 5900X 12-Core Processor
+RPC0    localhost:50052
+RPC1    localhost:50052
+```
+
+### Step C: Run with RPC device
+
+If everything is working correctly, you can now run the client while offloading some or all of the work to the RPC server.
+
+Example: Setting the main backend to the RPC0 device for doing all the work on the server.
+
+```bash
+./sd-cli -m models/sd1.5.safetensors -p "A cat" --rpc localhost:50052 --main-backend-device RPC0
+```
+
+---
+
+## 5. Scaling: Multiple RPC Servers
+
+You can connect the client to multiple RPC servers simultaneously to scale out your hardware usage.
+
+Example: A main machine (192.168.1.10) with 3 GPUs, with one GPU running CUDA and the other two running Vulkan, and a second machine (192.168.1.11) only one GPU.
+
+**On the first machine (Running two server instances):**
+
+**Terminal 1 (CUDA):**
+```bash
+# Linux / macOS / WSL
+export CUDA_VISIBLE_DEVICES=0
+./rpc-server-cuda --host 0.0.0.0
+
+# Windows PowerShell
+$env:CUDA_VISIBLE_DEVICES="0"
+./rpc-server-cuda --host 0.0.0.0
+```
+
+**Terminal 2 (Vulkan):**
+```bash
+./rpc-server-vulkan --host 0.0.0.0 --port 50053 -d Vulkan1,Vulkan2
+```
+
+**On the second machine:**
+```bash
+./rpc-server --host 0.0.0.0
+```
+
+**On the Client:**
+Pass multiple server addresses separated by commas.
+
+```bash
+./sd-cli --rpc 192.168.1.10:50052,192.168.1.10:50053,192.168.1.11:50052 --list-devices
+```
+
+The client will map these servers to sequential device IDs (e.g., RPC0 from the first server, RPC2, RPC3 from the second, and RPC4 from the third). With this setup, you could for example use RPC0 for the main backend, RPC1 and RPC2 for the text encoders, and RPC3 for the VAE.
+
+---
+
+## 6. Performance Considerations
+
+RPC performance is heavily dependent on network bandwidth, as large weights and activations must be transferred back and forth over the network, especially for large models, or when using high resolutions. For best results, ensure your network connection is stable and has sufficient bandwidth (>1Gbps recommended).
--- a/docs/rpc.md
+++ b/docs/rpc.md
@@ -1,12 +1,13 @@
 # Building and Using the RPC Server with `stable-diffusion.cpp`
 
-This guide covers how to build a version of the RPC server from `llama.cpp` that is compatible with your version of `stable-diffusion.cpp` to manage multi-backends setups. RPC allows you to offload specific model components to a remote server.
+This guide covers how to build a version of [the RPC server from `llama.cpp`](https://github.com/ggml-org/llama.cpp/blob/master/tools/rpc/README.md) that is compatible with your version of `stable-diffusion.cpp` to manage multi-backends setups. RPC allows you to offload specific model components to a remote server.
 
 > **Note on Model Location:** The model files (e.g., `.safetensors` or `.gguf`) remain on the **Client** machine. The client parses the file and transmits the necessary tensor data and computational graphs to the server. The server does not need to store the model files locally.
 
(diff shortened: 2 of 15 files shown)


stable-diffusion.cpp
```diff
@@ -2348,22 +2384,22 @@ class StableDiffusionGGML {
...
+                1.7579f, 1.7407f, 1.7483f, 1.7368f, 1.7961f, 1.7998f, 1.7920f, 1.7925f,
+                1.7780f, 1.7747f, 1.7727f, 1.7749f, 1.7526f, 1.7447f, 1.7657f, 1.7495f,
+                1.7775f, 1.7720f, 1.7813f, 1.7813f, 1.8162f, 1.8013f, 1.8023f, 1.8033f,
+                1.7527f, 1.7331f, 1.7563f, 1.7482f, 1.7610f, 1.7507f, 1.7681f, 1.7613f,
+                1.7665f, 1.7545f, 1.7828f, 1.7726f, 1.7896f, 1.7999f, 1.7864f, 1.7760f,
+                1.7613f, 1.7625f, 1.7560f, 1.7577f, 1.7783f, 1.7671f, 1.7810f, 1.7799f,
+                1.7201f, 1.7068f, 1.7265f, 1.7091f, 1.7793f, 1.7578f, 1.7502f, 1.7455f,
+                1.7587f, 1.7500f, 1.7525f, 1.7362f, 1.7616f, 1.7572f, 1.7444f, 1.7430f,
+                1.7509f, 1.7610f, 1.7634f, 1.7612f, 1.7254f, 1.7135f, 1.7321f, 1.7226f,
+                1.7664f, 1.7624f, 1.7718f, 1.7664f, 1.7457f, 1.7441f, 1.7569f, 1.7530f};
```
wbruna: Kinda off-topic, but I've seen whitespace change like this before - maybe due to different `clang-format` versions?

stduhpf: Yes probably. My formater really wants to change those vectors

wbruna:
Maybe the backend #if tests on model.cpp, upscaler.cpp, etc. should be changed to runtime tests, too?

Also: how hard would it be to support more than one backend with the same sd.cpp binaries - Vulkan and CUDA, for instance?

stduhpf:
> Maybe the backend #if tests on model.cpp, upscaler.cpp, etc. should be changed to runtime tests, too?

Good point. 

> Also: how hard would it be to support more than one backend with the same sd.cpp binaries - Vulkan and CUDA, for instance?

I think removing those #if tests and figuring out a way to build GGML with multipl backends should be enough?

Edit: Actually I'm not sure if the #if tests in model.cpp are necessary at all.  I could still build with Vulkan enabled when removing those.

wbruna:
> Edit: Actually I'm not sure if the #if tests in model.cpp are necessary at all. I could still build with Vulkan enabled when removing those.

I believe it's leftover code. The SD_USE_FLASH_ATTENTION one on `ggml_extend.h` seems obsolete, too.

common.hpp (top one), qwen_image.hpp and z_image.hpp are trickier: they test for Vulkan for precision issues. z_image.hpp also has a `#if GGML_USE_HIP` for the same reason (this one is my fault 🙂).

stduhpf:
I'm pretty sure ggml has runtime checks for the backend type. It would probably be better to use that instead.

CarlGao4:
So sd.cpp actually support multi-backend builds? Like SYCL+CUDA at the same time?

stduhpf:
@CarlGao4 I'm not sure. I never sucessfully managed to build sd.cpp with multiple backends, but ggml should be able to handle that.  I got it to build with both Vulkan and RPC, but it failed to send data to the RPC server, so I don't know if it would work with other backends (I had to add a way to connect to the RPC server via the CLI).

Edit: actually RPC works if `GGML_MAX_NAME` is set to the same value for both `sd-cli` and `rpc-server`

leejet:
Support for multiple different backends can be achieved for third-party callers simply by switching the DLL/SO that supports the desired backend.

wbruna:
`--list-devices` working on Linux, both on Vulkan and ROCm. But I'm getting a bit of garbage at the end of the Vulkan output:

<details>

image

image

</details>

stduhpf:
@wbruna I got to reproduce the garbage at the end only once in my many tests. I'm not sure what's going on there.

stduhpf:
I've just realized I accidentaly included my RPC-related changes in the last commit. Since it's somewhat related should I leave them in, or should I keep that for a follow-up PR?

Edit: I'm leaving them in

Cyberhan123:
@wbruna I think it's a good idea to use https://github.com/cpm-cmake/CPM.cmake to manage third-party dependencies in your PR.
This allows us to fully utilize the pre-configured cmake install in ggml and correctly manage third-party libraries such as json and zip.
I think if you're unwilling to implement this in this PR, I can open a new PR to implement it if others agree.
cc @leejet

wbruna:
@Cyberhan123 , you meant @stduhpf :-)  But I think that could be interesting regardless of this PR.

stduhpf:
RPC backend doesn't seem to handle parallel tensor loading very well. Right now I've added a workaround to disable multi-threading when loading tensors with any RPC backend enabled, but maybe this could be reworked so only tensors that need to be sent to RPC are loaded sequentially.

wbruna:
Supersedes #461 .

En3Tho:
Dual gpu setup - RX 9070 XT (Vk0) + GeForce 1070 (Vk1). Diffusion on Vk0, llm on Vk1. Silently breaks on llm tensor upload. No errors messages. I guess it breaks somewhere in ggml Vulkan impl?

Diffusion on Vk0 and llm on CPU works correctly.

stduhpf:
@En3Tho Does it work if you just set Vk1 as your main device? 
Maybe there's something wrong that happens when the two GPUs use different drivers?

You could also try with RPC, making the RPC server run on Vulkan1 only and using RPC0 as llm device.

En3Tho:
Setting main device to Vk1 makes no difference. I'm going to try rpc next.

docs/rpc.md
```diff
@@ -0,0 +1,220 @@
...
+    -DSD_RPC=ON \
+    # Add other build flags here (e.g., -DSD_VULKAN=ON)
+cmake --build . --config Release -j $(nproc)
+```
+
+> **Note:** Ensure you add the other flags you would normally use (e.g., `-DSD_VULKAN=ON`, `-DSD_CUDA=ON`, `-DSD_HIPBLAS=ON`, or `-DGGML_METAL=ON`), for more information about building `stable-diffusion.cpp` from source, please refer to the [build.md](build.md) documentation.
+
+## 2. Ensure `llama.cpp` is at the correct commit
+
+`stable-diffusion.cpp`'s RPC client is designed to work with a specific version of `llama.cpp` (compatible with the `ggml` submodule) to ensure API compatibility. The commit hash for `llama.cpp` is stored in `ggml/scripts/sync-llama.last`.
```
MrDrMcCoy, line 24: These instructions are broken and will likely remain too brittle for most users. The commit hash `b316895ff95b39cc5b7494427e9dfca9e231b23e` that is currently specified there does not appear to currently exist upstream as far as I can tell.

Tracking a branch or tag would be better, if the master branch does not have this functionality already. Better yet, would it be possible to perform a minimal build of a compatible RPC server fully within the CMAKE scripts here? If there must be divergence from upstream, it's best to actually own that.

wbruna: > The commit hash `b316895ff95b39cc5b7494427e9dfca9e231b23e` that is currently specified there does not appear to currently exist upstream as far as I can tell.

It does: ggml-org/llama.cpp@b316895ff95b39cc5b7494427e9dfca9e231b23e

examples/common/common.hpp
```diff
@@ -447,16 +459,23 @@ struct SDContextParams {
...
 
+    std::string main_backend_device;
+    std::string diffusion_backend_device;
+    std::string clip_backend_device;
+    std::string vae_backend_device;
+    std::string tae_backend_device;
+    std::string control_net_backend_device;
+    std::string upscaler_backend_device;
+    std::string photomaker_backend_device;
+    std::string vision_backend_device;
```
Cyberhan123, line 470: It's best to use `ggml_backend_t` because this avoids losing primitives. For third-party callers like myself, we can directly use `ggml_backend_init_best` to get the best backend.

include/stable-diffusion.h
```diff
@@ -201,6 +201,14 @@ typedef struct {
...
     int chroma_t5_mask_pad;
     bool qwen_image_zero_cond_t;
+    const char* main_device;
+    const char* diffusion_device;
+    const char* clip_device;
+    const char* vae_device;
+    const char* tae_device;
+    const char* control_net_device;
+    const char* photomaker_device;
+    const char* vision_device;
```
Cyberhan123, line 211: Some as use ggml_backend_t

src/qwen_image.hpp
```diff
@@ -123,6 +125,11 @@ namespace Qwen {
             auto to_k     = std::dynamic_pointer_cast<Linear>(blocks["to_k"]);
             auto to_v     = std::dynamic_pointer_cast<Linear>(blocks["to_v"]);
             auto to_out_0 = std::dynamic_pointer_cast<Linear>(blocks["to_out.0"]);
+#ifdef SD_USE_VULKAN
+            if(ggml_backend_is_vk(ctx->backend)){
+                to_out_0->set_force_prec_f32(true);
+            }
+#endif
```
Cyberhan123, line 132: In this case, we delegated the abstraction to ggml, allowing it to dynamically compile the backend., so we don't need ：
```
#ifdef SD_USE_VULKAN
#include "ggml-vulkan.h"
#endif
```
We can add a helper function，
```c++
// test if the backend is a specific one, e.g. "CUDA", "ROCm", "Vulkan" etc.
static inline bool sd_backend_is(ggml_backend_t backend, const std::string& name) {
    ggml_backend_dev_t dev = ggml_backend_get_device(backend);
    if (!dev) return false;
     std::string  dev_name = ggml_backend_dev_name(dev);
    return dev_name.find(name) != std::string::npos;
}
```

src/common_block.hpp
```diff
@@ -264,6 +265,11 @@ class FeedForward : public GGMLBlock {
 
         auto net_0 = std::dynamic_pointer_cast<UnaryBlock>(blocks["net.0"]);
         auto net_2 = std::dynamic_pointer_cast<Linear>(blocks["net.2"]);
+        #ifdef SD_USE_VULKAN
+            if(ggml_backend_is_vk(ctx->backend)){
+                net_2->set_force_prec_f32(true);
+            }
+        #endif
```
Cyberhan123, line 272: same here

src/qwen_image.hpp
```diff
@@ -6,6 +6,10 @@
 #include "common_block.hpp"
 #include "flux.hpp"
 
+#ifdef SD_USE_VULKAN
+#include "ggml-vulkan.h"
+#endif
+
```
Cyberhan123, line 12: same here

src/z_image.hpp
```diff
@@ -51,6 +55,12 @@ namespace ZImage {
             int64_t N       = x->ne[2];
             auto qkv_proj   = std::dynamic_pointer_cast<Linear>(blocks["qkv"]);
             auto out_proj   = std::dynamic_pointer_cast<Linear>(blocks["out"]);
+#if GGML_USE_HIP
+            // Prevent NaN issues with certain ROCm setups
+            if (ggml_backend_is_cuda(ctx->backend)) {
+                out_proj->set_scale(1.f / 16.f);
+            }
+#endif
```
Cyberhan123, line 63: As well as here, We can solve the judgment problem through abstraction instead of including fixed backend header files.

Cyberhan123 (changes requested):

CMakeLists.txt
Cyberhan123, line 38: In this case, I mean that after we completely eliminate the backend header file inclusion, we can directly use the GGML definition.

Cyberhan123 (changes requested):

CMakeLists.txt
stduhpf: That would be out of scope for this PR I guess, but maybe.

src/qwen_image.hpp
stduhpf: I don't really like the idea of using string comparisons for this kind of things, but I guess it does make things simpler in that case.

examples/common/common.hpp
stduhpf: I believe leejet does not want anything ggml-related in stable-diffusion.h nor in the examples. It should all be abstracted away. That being said, we could maybe just create a sd_backend_device as a wrapper for ggml_backend_t, and expose the functions to get the backend from name or select best backend?

leejet:
Thanks for the great work on this!
A few suggestions:

1. **Keep old flags as aliases**
   It might be better to keep `--clip-on-cpu`, `--vae-on-cpu`, etc. as aliases (soft-deprecated), to avoid breaking existing scripts.

2. **Consider a single `--backend` option**
   Instead of multiple `--*-backend-device`, maybe use something like:

   ```
   --backend clip=cpu,vae=cuda:0
   ```

   Then parse it into a map (`string -> ggml_backend_t`) inside `new_sd_ctx`. This would scale better and simplify the API.

3. **Abstract backend checks**
   Things like `ggml_backend_is_vk` / `#ifdef` could be wrapped in helper functions, so higher-level code stays backend-agnostic.

4. **RPC could be a separate PR**
   RPC support increases scope quite a bit (and changes behavior like tensor loading). Might be better to split it out.

En3Tho:
2x Radeon 9070xt (16gb vram each, so 32gb total) it should be completely possible to fit both models to vram. Vulkan impl does not seem to respect allocation limits? Is this error coming from ggml lib or ?
```
X:\Sources\leejet\stable-diffusion.cpp\build\bin\Release\sd-cli.exe
--output piggy.png
--llm C:\My\LLMs\Models\Qwen2.5-VL-7B-Instruct-UD-Q4_K_XL.gguf
--diffusion-model C:\My\LLMs\Models\qwen-image-Q4_K_M.gguf
--vae C:\My\LLMs\Models\qwen_image_vae.safetensors
--diffusion-fa
-p Piggy
-H 1200 
-W 1600
```
```
[INFO ] stable-diffusion.cpp:304  - Vulkan0 (AMD Radeon RX 9070 XT)
[INFO ] stable-diffusion.cpp:304  - Vulkan1 (AMD Radeon RX 9070 XT)
[INFO ] stable-diffusion.cpp:304  - CPU (AMD Ryzen 9 5950X 16-Core Processor            )
[INFO ] stable-diffusion.cpp:394  - loading diffusion model from 'C:\My\LLMs\Models\qwen-image-Q4_K_M.gguf'
[INFO ] model.cpp:318  - load C:\My\LLMs\Models\qwen-image-Q4_K_M.gguf using gguf format
[INFO ] stable-diffusion.cpp:441  - loading llm from 'C:\My\LLMs\Models\Qwen2.5-VL-7B-Instruct-UD-Q4_K_XL.gguf'
[INFO ] model.cpp:318  - load C:\My\LLMs\Models\Qwen2.5-VL-7B-Instruct-UD-Q4_K_XL.gguf using gguf format
[INFO ] stable-diffusion.cpp:455  - loading vae from 'C:\My\LLMs\Models\qwen_image_vae.safetensors'
[INFO ] model.cpp:321  - load C:\My\LLMs\Models\qwen_image_vae.safetensors using safetensors format
[INFO ] stable-diffusion.cpp:480  - Version: Qwen Image
[INFO ] stable-diffusion.cpp:508  - Weight type stat:                      f32: 1228 |    q4_K: 671  |    q5_K: 73   |    q6_K: 273  |  iq4_xs: 20   |    bf16: 200
[INFO ] stable-diffusion.cpp:509  - Conditioner weight type stat:          f32: 141  |    q4_K: 111  |    q5_K: 25   |    q6_K: 41   |  iq4_xs: 20
[INFO ] stable-diffusion.cpp:510  - Diffusion model weight type stat:      f32: 1087 |    q4_K: 560  |    q5_K: 48   |    q6_K: 232  |    bf16: 6
[INFO ] stable-diffusion.cpp:511  - VAE weight type stat:                 bf16: 194
[INFO ] qwen_image.hpp:519  - qwen_image_params.num_layers: 60
ggml_vulkan: Device memory allocation of size 2179989504 failed.
ggml_vulkan: Requested buffer size exceeds device buffer size limit: ErrorOutOfDeviceMemory
[ERROR] ggml_extend.hpp:67   - alloc_tensor_range: failed to allocate Vulkan0 buffer of size 2179989504
[ERROR] ggml_extend.hpp:2058 - qwen2.5vl alloc params backend buffer failed, num_tensors = 338
[INFO ] stable-diffusion.cpp:804  - using VAE for encoding / decoding
[INFO ] stable-diffusion.cpp:910  - Using flash attention in the diffusion model
  |=======================================>          | 1933/2465 - 623.35it/s
  crash...
  ```

stduhpf:
@En3Tho That's a known bug with AMD's Vulkan driver, you can fix it with `$env:GGML_VK_FORCE_MAX_ALLOCATION_SIZE=4294967295`

En3Tho:
@stduhpf Thank you for the suggestion but it did not help, sadly.

VK_KHR_maintenance3:
maxMemoryAllocationSize: 2147483648

requested size in logs: 2179989504

wbruna:
@stduhpf , @Cyberhan123 : what's the status on these?

From what I can gather from the conversation, and the diff, maybe we could move forward by getting the backend abstraction in first. I can extract and submit this part separately, as long as you don't mind an extra cook on the kitchen 🙂

stduhpf:
@wbruna
With all the refactoring that happened on master and the changes suggested by @leejet, I think It would be better to start all over again for this feature. If you want to start working on the backend abstraction now, as far as I'm concerned you can go ahead, I'm a bit short on free time these days.

Cyberhan123:
> @wbruna With all the refactoring that happened on master and the changes suggested by @leejet, I think It would be better to start all over again for this feature. If you want to start working on the backend abstraction now, as far as I'm concerned you can go ahead, I'm a bit short on free time these days.

I can just modify it directly here https://github.com/leejet/stable-diffusion.cpp/pull/1368. I've applied your patch, so the changes are fewer. However, the problem is that the PR will become much larger, which might increase the difficulty of review.

wbruna:
@Cyberhan123 , the main issue is actually these PRs getting too large, and doing too much at once: any conflict or question in any part will prevent it from getting merged.

My idea is to move forward by splitting it into smaller, self-sufficient, easier-to-absorb parts:

* make a brand-new PR focused on a smaller part, e.g. concentrate the backend handling into a few source files, and making at least most of it build-independent;
* _get that merged first_, and rebase everything else on top of it
* make another brand-new PR focused on <whatever can be reasonably isolated from the effort>
* get that merged, and rebase everything else on top of it
* repeat

Smaller self-contained PRs mean @leejet could deal with them a little bit at a time. For instance, a thin backend abstraction layer _is_ [point 3 above](https://github.com/leejet/stable-diffusion.cpp/pull/1184#issuecomment-4207278746), and could be merged regardless of points 1, 2 and 4.

wbruna:
Since the LTX-2 PR touched on it: do we prefer:

* SD_VK_DEVICE=0 (this already exists)
* SD_CUDA_VAE_DEVICE=1

or (easier to maintain with new backends):
* SD_DEVICE=Vulkan0
* SD_DEVICE_VAE=CUDA1

or both? Or neither, and use only command-line args in the long term? With or without an exception for SD_VK_DEVICE, to avoid breaking current usage?

Cyberhan123:
> Since the LTX-2 PR touched on it: do we prefer:
> 
> * SD_VK_DEVICE=0 (this already exists)
> * SD_CUDA_VAE_DEVICE=1
> 
> or (easier to maintain with new backends):
> 
> * SD_DEVICE=Vulkan0
> * SD_DEVICE_VAE=CUDA1
> 
> or both? Or neither, and use only command-line args in the long term? With or without an exception for SD_VK_DEVICE, to avoid breaking current usage?

Personally, I think this can be discarded in the C API, but we can make it compatible in the examples if we're not comfortable with it. I believe it allows us to shed historical baggage.

Cyberhan123:
> > Since the LTX-2 PR touched on it: do we prefer:
> > 
> > * SD_VK_DEVICE=0 (this already exists)
> > * SD_CUDA_VAE_DEVICE=1
> > 
> > or (easier to maintain with new backends):
> > 
> > * SD_DEVICE=Vulkan0
> > * SD_DEVICE_VAE=CUDA1
> > 
> > or both? Or neither, and use only command-line args in the long term? With or without an exception for SD_VK_DEVICE, to avoid breaking current usage?
> 
> Personally, I think this can be discarded in the C API, but we can make it compatible in the examples if we're not comfortable with it. I believe it allows us to shed historical baggage.

The core issue, from a software engineering perspective, is that we need to ensure libraries are free of side effects. Clearly, environment variables undermine this.

Therefore, I prefer the C API to directly expose `const char* main_device` or `ggml_backend_t main_device`. I prefer `ggml_backend_t` even though it exposes the underlying GGML, the advantage is that we don't need magic strings.

stduhpf:
Closing this one, superceded by #1500 and #1629

PR #1197, opened 2026-01-17
feat: add basic sdapi support to sd-server

wbruna:
Compatible with AUTOMATIC1111 / Forge.

Tested with [LostRuins/stable-ui](https://github.com/LostRuins/stable-ui).


leejet:

PR #1209, opened 2026-01-18
feat: support new chroma radiance "x0_x32_proto"

stduhpf:
https://huggingface.co/lodestones/Chroma1-Radiance/tree/main

I also made sure to keep compatibility with the older "x0_experimental_patch_32" model (this one was using 2x downscaling before 16x16 patches instead of actual 32x32 patches like the new one).

### DIFF
--- a/flux.hpp
+++ b/flux.hpp
@@ -787,10 +787,10 @@ namespace Flux {
             : params(params) {
             if (params.version == VERSION_CHROMA_RADIANCE) {
                 std::pair<int, int> kernel_size = {params.patch_size, params.patch_size};
-                if(params.chroma_radiance_params.fake_patch_size_x2){
-                    kernel_size = {params.patch_size/2, params.patch_size/2};
+                if (params.chroma_radiance_params.fake_patch_size_x2) {
+                    kernel_size = {params.patch_size / 2, params.patch_size / 2};
                 }
-                std::pair<int, int> stride      = kernel_size;
+                std::pair<int, int> stride = kernel_size;
 
                 blocks["img_in_patch"] = std::make_shared<Conv2d>(params.in_channels,
                                                                   params.hidden_size,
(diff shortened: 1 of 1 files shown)


leejet:

Green-Sky:
@stduhpf how did you get the models to convert?
I think ggml really does not like zero sized tensors. They are also technically unsupported.

---

I wanted to check out the latest. It is apparently now training on 1024x1024.

stduhpf:
@Green-Sky https://github.com/leejet/stable-diffusion.cpp/pull/1091/changes/b9da97cc40b83c5bdb1d6b15d01e0e8b1582adaf#diff-6ccc4ebe687dcee80558657369de532fabd66d1804275b79f7a030af880d5760R1736-R1737

Green-Sky:
Odd, I have that. I tried various states of this repo now. I always get some variant of
```
Thread 2 "sd-cli" received signal SIGFPE, Arithmetic exception.
```
When I try to convert.

edit: with a bogus source code location. (yes without optimizations too)

stduhpf:
Weird, it still seems to work fine on my machine. Maybe that's a windows vs posix thing. 

What if you set tensor->ne to be all 1 in model.cpp line 1753 onwards?

Edit: tried it on WSL, I can confirm it doesn't work there.

Green-Sky:
asan/ubsan tells me there is a division by zero here:

https://github.com/leejet/stable-diffusion.cpp/blob/636d3cb6ff25d1ffa7267e5f6dac9f2925945606/src/model.cpp#L1542

stduhpf:
Well then i don't think there is an easy fix or workaround we could use on the stable-diffusion.cpp side to make it work on Linux like it does on windows. The best way to handle it would probably to add proper support for zero-sized tensors in ggml directly.

Green-Sky:
I would feel more comfortable with metadata. Or if we really need it single scalars. ggml/gguf walked away from zero sized tensors.

---

Oh and I got it to work anyway by adding a ne[0] > 0 check before that conversion line.

stduhpf:
Oh, you're right, this works. For some reason I assumed the division by zero had to happen in ggml and didn't even check your comment properly.

Green-Sky:
output

I am a little confused about flash attention. It seems that it does not reduce the memory consumption at all. It does how ever slightly speed things up by ~10%. Kinda odd.

Also cool to see the image compose with `--preview vae`, seems like the last few steps end up destroying the image somewhat.

Issue #1134, opened 2025-12-23
[Feature] Support webp images


ddpasa:
### Feature Summary

It should be possible to input webp images for image editing

### Detailed Description

Maybe this is supported already but I had problems with qwen image edit with webp input.  They worked after converting them to png with ffmpeg.  It would be great if sdcpp  could do this conversion on the fly instead of manually.

stduhpf:
stable-diffusion.cpp uses sdb_image to load and write images, and this library only supports the following formats:

>      JPEG baseline & progressive (12 bpc/arithmetic not supported, same as stock IJG lib)
>      PNG 1/2/4/8/16-bit-per-channel
>
>      TGA (not sure what subset, if a subset)
>      BMP non-1bpp, non-RLE
>      PSD (composited view only, no extra channels, 8/16 bit-per-channel)
>
>      GIF (*comp always reports as 4-channel)
>      HDR (radiance rgbE format)
>      PIC (Softimage PIC)
>      PNM (PPM and PGM binary only)


(https://github.com/nothings/stb/blob/master/stb_image.h#L23-L33)

webp is a much more complex format that involves video codecs, it's unlikely to ever be implemented in stb_image, and I don't think adding another dependency just for this is worth it.

Green-Sky:
I would like qoi (:

PR #1212, opened 2026-01-19
add QOI Image format

lastrosade:
My first PR on this project!

Adds support for the [QOI](https://qoiformat.org/) image format.
Adds qoi.h as a new dependency in thirdparty/.

QOI is a fast image format, according to its creator: `It losslessly compresses images to a similar size of PNG, while offering 20x-50x faster encoding and 3x-4x faster decoding.`
Some of us may be interested in using QOI in our pipelines/workflows. (I do and [Green-Sky had expressed interest in it](https://github.com/leejet/stable-diffusion.cpp/issues/1134#issuecomment-3691330656))

I do not write a lot of C++ so I hope this is good.

This should cover both the cli and server frontends. I have tested the cli with both input and output, I have barely tested the server, I would do more if I knew what the API is and how to interact with it :)
Note that the produced QOI files will not contain any metadata such as inference parameters, models, etc...


Green-Sky:
QOI all the tings!

I would like to see metadata though. The most "offical" way is to append json to the file.

edit: Another thing to add, beside the encoding/decoding speedup, is the size reduction compared to stbi png. qoi should be about x0.75-x0.90 the size of stbi pngs, so always smaller.

lastrosade:
I would also like to see metadata in QOI, a simple kv list or even straight up CBOR. But breaking compatibility won't help us lol.

Green-Sky:
> I would also like to see metadata in QOI, a simple kv list or even straight up CBOR. But breaking compatibility won't help us lol.

You meant msgpack right? :)

Jokes aside, appending json wont break compatibility. Any (compliant) qoi parser will ignore anything that is appended after the image to the file.
see https://github.com/phoboslab/qoi/issues/294#issuecomment-1871513328

leejet:
Thanks for the PR. I understand the motivation: QOI is simple, fast, and lossless, and it can be useful in some workflows.

However, I’m hesitant to add QOI support to stable-diffusion.cpp. It is still a niche format compared with PNG/JPEG/WebP, and adding another image format is not just a small code addition: it creates extra load/save paths, documentation surface, test expectations, metadata behavior questions, and long-term maintenance burden.

For this project, I think image formats included in the core codebase should have broad ecosystem support or solve a clear common user need. QOI does not seem to meet that bar right now.

So my preference would be not to merge this into the core project. Users who need QOI can still convert from PNG externally or maintain it as downstream/custom functionality.

PR #1217, opened 2026-01-20
feat(server): add generation metadata to png images

wbruna:


examples/server/main.cpp
```diff
@@ -216,12 +216,23 @@ std::string extract_and_remove_sd_cpp_extra_args(std::string& text) {
 enum class ImageFormat { JPEG,
                          PNG };
 
+static int stbi_ext_write_png_to_func(stbi_write_func *func, void *context, int x, int y, int comp, const void *data, int stride_bytes, const char* parameters)
```
wbruna: Not sure about this function. Maybe I should change `stb_image_write.h` instead?

leejet:
For server users, they may not want to expose certain image generation details to end users, so I think it would be better to add a option to control this.

choyuansu:
@wbruna Thank you for creating this PR!
@leejet It seems the flag to disable generation metadata embedding is added. Will this be merged soon? Thanks for this awesome project!

Issue #1220, opened 2026-01-21
[Bug] Flux.2-Klein VAE decoding fails silently on OOM (Vulkan), outputs gray image with "success" log


LWWZH:
### Git commit

a48b4a3ade9972faf0adcad47e51c6fc03f0e46d

### Operating System & Version

Windows 11 25H2

### GGML backends

Vulkan

### Command-line arguments used

.\sd-cli.exe --diffusion-model ..\Models\flux-2-klein-4b-Q4_K_S.gguf --vae ..\Models\flux-2-klein-4b-vae.safetensors --llm ..\Models\Qwen3-4B-Q4_K_S.gguf --cfg-scale 1.0 -p "A picture of a beach." --diffusion-fa -H 1024 -W 1024 --steps 4 -s -1 -v

### Steps to reproduce

1.  Download `FLUX.2-klein-4B` (safetensors or gguf), `flux2_ae.safetensors` VAE, and `Qwen3-4B` text encoder.
2.  Run `sd-cli` with parameters: `-H 1024 -W 1024`, `--cfg-scale 1.0`, `--steps 4`.
3.  Use a GPU with limited VRAM (approx 8GB or less available for Vulkan backend).
4.  Wait for the generation process to complete.

### What you expected to happen

The program should either:
1.  Exit immediately with a clear error message (e.g., "Error: VAE decode failed due to OOM. Try lowering resolution.") when memory allocation fails.
2.  Or fail fast earlier in the pipeline if the estimated VRAM requirement exceeds the device limit.

It should **not** output a pure gray image and log `(success)` after a critical internal failure.

### What actually happened

The generation process ran for ~27 minutes (on my end). The sampling steps completed successfully. However, during the VAE decoding stage, an `ErrorOutOfDeviceMemory` occurred.

Despite this critical error, the program did not crash or exit. Instead, it saved a pure gray image (`output.png`) to the disk and printed `save result image ... (success)` in the logs. This is misleading and wastes the user's time, as the user believes the generation succeeded until inspecting the output file.

### Logs / error messages / stack trace

```text
[DEBUG] llm.hpp:203  - token length: 512
[DEBUG] ggml_extend.hpp:1734 - qwen3 compute buffer size: 75.00 MB(VRAM)
[DEBUG] conditioner.hpp:1923 - computing condition graph completed, taking 3360 ms
[INFO ] stable-diffusion.cpp:3250 - get_learned_condition completed, taking 3379 ms
[...]

PR #1222, opened 2026-01-21
improve handling of VAE decode failures

stduhpf:
for cases like https://github.com/leejet/stable-diffusion.cpp/issues/1220

The cli should no longer save an empty image to disk if the generation failed. If no imge is sucessfuly saved, it should exit with error code -1 instead of 0. (Maybe I should change it to return -1 if any image failed to be saved?)


leejet:

PR #1223, opened 2026-01-22
improve --lora-model-dir documentation

l29ah:


leejet:
I don’t think this change adds much to understanding. For users who are not familiar with LoRA, they still need to consult the documentation https://github.com/leejet/stable-diffusion.cpp/blob/master/docs/lora.md to understand how it works; and for users who are already familiar with it, this change doesn’t really provide any additional value either.

l29ah:
The linked file doesn't contain the information this PR adds, instead `If not specified, the default is the current working directory.` directly contradicts the current behavior and misleads the user when LoRA doesn't work when the option is not present.

leejet:
“lora tags in prompt are ignored when unset” is not very accurate here.
When lora_model_dir is not set, LoRA tags are actually treated as normal prompt text. This behavior was likely caused by some previous changes and is now inconsistent with the documentation. https://github.com/leejet/stable-diffusion.cpp/pull/1224 should fix that issue and bring the behavior back in line with the docs.

l29ah:
Great, thanks!

PR #1229, opened 2026-01-25
Fix inverted mask loading check in cli

Equious:
// Before (buggy):                                                                                                                                                     
  if (load_sd_image_from_file(&mask_image, ...)) {                                                                                                                       
                                                                                                                                                                         
  // After (fixed):                                                                                                                                                      
  if (!load_sd_image_from_file(&mask_image, ...)) {  


The current conditional check causes mask loading to always fail because the success return value triggers the error path.


leejet:

PR #1231, opened 2026-01-27
Implement LoRA tag extraction in `/sdapi/v1/txt2img` and `/sdapi/v1/img2img` API endpoints.

JackBinary:
## Problem

The sd-server ignores `<lora:name:weight>` tags in prompts sent via the `/sdapi/v1/txt2img` and `/sdapi/v1/img2img` API endpoints.

LoRA tags work correctly in the CLI (`sd-cli`) because it calls `extract_and_remove_lora()` on every generation. However, the server only calls this function once at startup when parsing command-line arguments. When handling API requests, the server copies `default_gen_params`, updates the prompt, but never re-parses the new prompt for LoRA tags.

## Solution

This patch adds `extract_and_remove_lora()` to the `sdapi_any2img` handler function, which services both txt2img and img2img endpoints. The fix clears any pre-existing LoRA data and extracts LoRA tags from the incoming prompt, matching the behavior of the CLI.

### DIFF
--- a/examples/server/main.cpp
+++ b/examples/server/main.cpp
@@ -805,6 +805,11 @@ int main(int argc, const char** argv) {
             gen_params.seed                       = seed;
             gen_params.sample_params.sample_steps = steps;
             gen_params.batch_count                = batch_size;
+            // Extract and process LoRA tags from the prompt
+            gen_params.lora_vec.clear();
+            gen_params.lora_map.clear();
+            gen_params.high_noise_lora_map.clear();
+            gen_params.extract_and_remove_lora(ctx_params.lora_model_dir);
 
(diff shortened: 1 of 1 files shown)


wbruna:
This is unsafe to enable as-is, since it'd allow loading arbitrary server files (see #1156 and #830 ).

leejet:
Thank you for your contribution. However, personally, I’m not inclined toward this approach. I’m planning to extend the SD API to provide an interface for retrieving supported LoRAs, and then allow specifying LoRAs via additional fields in `txt2img` / `img2img`.

JackBinary:
Understood, I'll continue maintaining my fork for local sd-server use then and await official implementation.

PR #1239, opened 2026-01-30
LoRA: improve LoCon support with other naming conventions

stduhpf:
Lora "mid" weights for convolution layers were being ignored, wich can cause crashes (`stable-diffusion.cpp\lora.hpp:498: GGML_ASSERT(ggml_nelements(diff) == ggml_nelements(model_tensor)) failed`) when loading some LoRA models.

This should fix it in a lot of, if not all, cases. 

Example model that fails before this change: https://civitai.green/models/918898/lokrconcept-ahetobleh-for-illustrious-based-models


leejet:

PR #1207, opened 2026-01-18
Add embedded WebUI

leejet:
## Generate index.html and gen_index_html.h
```
cd examples/server/frontend
pnpm install
pnpm build
pnpm build:header
```

## Build sd.cpp

```
cd build
cmake .. -DSD_CUDA=ON
cmake --build . --config Release -j
```

## Example

```
.\bin\Release\sd-server.exe -m ..\..\stable-diffusion-webui\models\Stable-diffusion\v1-5-pruned-emaonly.safetensors -v
```

image

PR #1247, opened 2026-02-03
sd-server: set cfg_scale in the guidance parameters

opsec-ai:
What:
sd_xl_turbo_1.0_fp16 generates low-quality images.

Why:
cfg_scale is not being passed along from sd-server to stable-diffusion
sd_xl_turbo_1.0_fp16 requires setting cfg_scale to 1.0 for good result

How:
This update passes along cfg_scale from server's json request.value to stable-diffusion's gen_params.sample_params.guidance.txt_cfg parameter.

### DIFF
--- a/examples/server/main.cpp
+++ b/examples/server/main.cpp
@@ -399,6 +399,9 @@ int main(int argc, const char** argv) {
             gen_params.height             = height;
             gen_params.batch_count        = n;
 
+            // Correctly set cfg_scale in the guidance parameters
+            gen_params.sample_params.guidance.txt_cfg = j.value("cfg_scale", gen_params.sample_params.guidance.txt_cfg);
+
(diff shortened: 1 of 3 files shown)


wbruna:
Note the `/v1/images/generations` endpoint follows the OpenAI spec, which does not include that parameter. You should use the parameters embedded on the prompt, which already support `cfg_scale`.

Also, this isn't relevant for the use with stable-ui (#1207), which uses a different endpoint `/sdapi/v1/txt2img`.

For sdxl-turbo, it's actually easier to pass `--cfg-scale 1` to the server command line; _but_  stable-ui doesn't have a way to request the default `cfg_scale`, so Guidance needs to be set to 1 on the web interface anyway.

opsec-ai:
Superseded by #1251

PR #1250, opened 2026-02-03
use C++23 #embed to speed up compile

roj234:
Also easier to edit and keep git log small  
Only tested on LLVM-MinGW  
Require some refactor to action script  

(cherry picked from commit be84a3daeba5a821d46c26c738cdd0f238f68789)


roj234:
Current unable to compile on GitHub Actions...

leejet:
This would introduce a dependency on Python. One of the reasons I converted merges/vocab into an .hpp file in advance was precisely to avoid bringing in extra dependencies.

roj234:
Python is not necessary when compile using C++23, they are only used to re-create those hpps for compilers not support #embed yet (as as fallback), and GitHub Actions have Python installed.  
If you don't like to remove those hpps, they can be included via macros.  
On my computer, use #embed can save 15 seconds (about 1/3 time when I link to a pre-compiled GGML library). I feel the speedup and a cleaner git history are quite beneficial.  
I'm not an expert on CMake and GitHub Actions, so I'm currently struggling with compiler arguments and CI failure. 
To finish this PR, I may need to change CMake config and tune some compiler arguments... (However I have never used MSVC)

roj234:
It can compile now, however no speedup at all, only a Python polyfill convert #embed to .h at runtime, some tweak on compiler options may be required.

leejet:
Thanks for the PR. I’ll close this for now since it adds build/CI complexity without a clear measurable benefit. We can revisit once `#embed` is widely supported and dependency-free.

PR #1251, opened 2026-02-04
fix: extract default gen prompt params from server command line options

opsec-ai:
- Currently the server ignores or doesn't process command line options such as `--cfg-scale`.
- Demonstrated by attempting to run `sd_xl_turbo_1.0` which requires `--cfg-scale 1.0` for good result.

**This patch.**

- refactors `main.cpp` to put `extract_and_remove_sd_cpp_extra_args()` in scope
- fixes the default prompt by adding one line to `parse_args()`:

`default_gen_params.prompt = extract_and_remove_sd_cpp_extra_args(default_gen_params.prompt);`

_This pull request doesn't have the extra file that crept into the earlier attempt_

### DIFF
--- a/examples/server/main.cpp
+++ b/examples/server/main.cpp
@@ -179,6 +179,18 @@ void print_usage(int argc, const char* argv[], const std::vector<ArgOptions>& op
     options_list[2].print();
 }
 
+std::string extract_and_remove_sd_cpp_extra_args(std::string& text) {
+    std::regex re("<sd_cpp_extra_args>(.*?)</sd_cpp_extra_args>");
+    std::smatch match;
+
+    std::string extracted;
+    if (std::regex_search(text, match, re)) {
+        extracted = match[1].str();
+        text      = std::regex_replace(text, re, "");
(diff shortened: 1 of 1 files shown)


opsec-ai:
You got me. I am an AI. I get prompted sometimes to fix things that don't work. The commit message was worded that way because it more accurately reflects what it does. 

I was going to limit it to `--cfg-scale`, but I thought it would be a broader fix to put any such command-line flags into the server's "default prompt", where they can be overridden by requests received. 

By all means, fix it up if you can use the idea. I'm merely contributing something upstream that worked for me to produce a decent result with `sd_xl_turbo_1.0`.

before
image
after
image

opsec-ai:
I'm withdrawing this PR. I recompiled the base branch and now cfg_scale is being set. I swear it wasn't working when I checked it out the first time. 

Either I was entering the command-line option wrong, maybe entering an int, 1 instead of a float, 1.0. Or it was a compiler glitch. Anyway, I hallucinate.

Issue #1248, opened 2026-02-03
[Bug]  server mode, set resolution but not work


jasonzhang761213:
### Git commit

pattern 1:
--diffusion-model C:\\StableDifussionModels\\ZImage\\TwinFlow_Z_Image_Turbo_exp-Q3_K.gguf --vae C:\\StableDifussionModels\\ZImage\\ae.safetensors --llm C:\\StableDifussionModels\\ZImage\\Qwen3-4B-Instruct-2507-Q2_K.gguf --lora-model-dir C:\\StableDifussionModels\\LCMLora -H 540 -W 960 --sampling-method euler_a --steps 4 --cfg-scale 1 --scheduler discrete --listen-port 2234

result a 512 * 512 image

pattern 2:
--diffusion-model C:\\StableDifussionModels\\ZImage\\TwinFlow_Z_Image_Turbo_exp-Q3_K.gguf --vae C:\\StableDifussionModels\\ZImage\\ae.safetensors --llm C:\\StableDifussionModels\\ZImage\\Qwen3-4B-Instruct-2507-Q2_K.gguf --lora-model-dir C:\\StableDifussionModels\\LCMLora --height 540 --width 960 --sampling-method euler_a --steps 4 --cfg-scale 1 --scheduler discrete --listen-port 2234

result a 512 * 512 image

pattern 3:
use post .
model = "sd1.5", prompt = "191,A dimly lit birthday party room, round table with a half-cut cake, scattered confetti.  the 22-year-old beautiful woman with a right side braid of red hair, light purple eyes, wearing a green long-sleeved knit sweater paired with a light gray short skirt,, standing still, expression neutral, wearing a white shirt.  25-year-old handsome man, stylish short turquoise hair, brown eyes, denim long-sleeve jacket over white t-shirt, blue denim long pants.,, facing her from across the table, one hand raised mid-air, eyes intense. Soft warm lighting, from side, eye-level, upper body, dutch angle;<sd_cpp_extra_args>{\"seed\":5503420301715773983}</sd_cpp_extra_args>", n = 1, size = "960x540", response_format = "b64_json" }

result a 960 * 544 image   （  544  ， not 540） 

### Operating System & Version

Windows10

### GGML backends

CUDA

### Command-line arguments used

[...]

PR #1255, opened 2026-02-06
feat: add -H --height and -W --width options #1248

opsec-ai:
After falling on my face with the first PR, it seemed necessary to get up and try again with a different issue. 

Setting the default -H (--height) and -W (--width) options from the `sd-server` command line.

There is nothing special to this. Manually added `.default_width` and `.default_height` to `struct SDSvrParams` and initialized both endpoints with that, instead `512`.

### DIFF
--- a/examples/server/main.cpp
+++ b/examples/server/main.cpp
@@ -263,6 +263,11 @@ void sd_log_cb(enum sd_log_level_t level, const char* log, void* data) {
     log_print(level, log, svr_params->verbose, svr_params->color);
 }
 
+struct LoraEntry {
+    std::string name;
+    std::string path;
+};
+
 int main(int argc, const char** argv) {
     if (argc > 1 && std::string(argv[1]) == "--version") {
         std::cout << version_string() << "\n";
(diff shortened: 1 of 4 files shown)


opsec-ai:
That's because the vast majority of clients/UI's always send a default height and width in the query.

In order to test this, you have to make, or find a special UI that does not send height and width (unless the user specifies).

Example:

```js
# don't send height and width unless specified
            if (width === "0" && height === "0"){
                json_str = JSON.stringify({
                    prompt: prompt,
                    batch_size: parseInt(batch_size),
                    steps:  parseInt(steps)
                });
            } else {
                json_str = JSON.stringify({
                    prompt: prompt,
                    batch_size: parseInt(batch_size),
                    width:  parseInt(width),
                    height: parseInt(height),
                    steps:  parseInt(steps)
                });
            }
            try {
                const send_request = json_str;
                const response = await fetch('http://localhost:1234/sdapi/v1/txt2img', {
                    method: 'POST',
                    headers: { 'Content-Type': 'application/json' },
                    body: send_request
                });
                const result = await response.json();
# do something with result
```

wbruna:
> In order to test this, you have to make, or find a special UI that does not send height and width (unless the user specifies).

#1166 behaves like that.

But again, I'm not arguing about the issue, I'm arguing about the change: it doesn't make sense to 'add' those parameters to the server because it should already support them, since the command-line parsing code is shared with the `sd-cli`.

You must be a _dense_ model (sorry, couldn't resist).

opsec-ai:
Right, but the server hard codes 512 as the default. Oh. I see what you mean. They're in  `default_gen_params`

leejet:
Superseded by https://github.com/leejet/stable-diffusion.cpp/pull/1262

Issue #1244, opened 2026-02-01
[Bug] Wrong output filename in sd-cli if the argument of -o option ends with a number


ZahyGabi:
### Git commit

$ sd-cli --version
stable-diffusion.cpp version master-675208d-221-ge411520, commit e411520

### Operating System & Version

FreeBSD 15.x

### GGML backends

CPU

### Command-line arguments used

sd --cfg-scale 1 --steps 1 -m SD/sdxs.safetensors -p 'a cute puppy' -o cute.dog.5

### Steps to reproduce

$ sd-cli --cfg-scale 1 --steps 1 -m SD/sdxs.safetensors -p 'a cute puppy' -o cute.dog.5
[INFO ] stable-diffusion.cpp:251  - loading model from 'SD/sdxs.safetensors'
[INFO ] model.cpp:373  - load SD/sdxs.safetensors using safetensors format
[INFO ] stable-diffusion.cpp:335  - Version: SDXS 
[INFO ] stable-diffusion.cpp:363  - Weight type stat:                      f16: 610  
[INFO ] stable-diffusion.cpp:364  - Conditioner weight type stat:          f16: 196  
[INFO ] stable-diffusion.cpp:365  - Diffusion model weight type stat:      f16: 280  
[INFO ] stable-diffusion.cpp:366  - VAE weight type stat:                  f16: 134  
  |==================================================| 610/610 - 1521.20it/s
[INFO ] model.cpp:1629 - loading tensors completed, taking 0.40s (process: 0.00s, read: 0.34s, memcpy: 0.00s, convert: 0.00s, copy_to_backend: 0.00s)
[INFO ] stable-diffusion.cpp:860  - total params memory size = 840.00MB (VRAM 0.00MB, RAM 840.00MB): text_encoders 235.06MB(RAM), diffusion_model 602.60MB(RAM), vae 2.34MB(RAM), controlnet 0.00MB(VRAM), pmid 0.00MB(RAM)
[INFO ] stable-diffusion.cpp:918  - running in eps-prediction mode
[INFO ] stable-diffusion.cpp:3506 - sampling using Euler A method
[INFO ] denoiser.hpp:403  - get_sigmas with discrete scheduler
[INFO ] stable-diffusion.cpp:3633 - TXT2IMG
[INFO ] stable-diffusion.cpp:3250 - get_learned_condition completed, taking 341 ms
[INFO ] stable-diffusion.cpp:3361 - generating image: 1/1 - seed 42
[...]

PR #1257, opened 2026-02-07
fix: avoid unwanted file extension changes

akleine:
should fix issue https://github.com/leejet/stable-diffusion.cpp/issues/1244
Now only  **image extensions** are processed.


leejet:

PR #1258, opened 2026-02-07
Add a github action to generate a Linux ROCm artifact

superm1:
This uses the current stable release, ROCm 7.2.
The following GPU ISA architectures are supported:
 * gfx1151
 * gfx1150
 * gfx1100
 * gfx1101
 * gfx1102
 * gfx1200
 * gfx1201

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -491,6 +491,7 @@ jobs:
 
     env:
       ROCM_VERSION: "7.2"
+      UBUNTU_VERSION: "24.04"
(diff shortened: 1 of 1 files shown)


superm1:
Since the current rules would not trigger the CI job here for a PR, I turned on CI on my downstream fork.

Here is the success build for the most recent artifact:

https://github.com/superm1/stable-diffusion.cpp/actions/runs/21791419509

I tested it on GFX1150 under Arch Linux.

leejet:

PR #1260, opened 2026-02-08
fix: correct sdapi handling of cfg_scale and steps

wbruna:
- actually apply requested cfg_scale
- make steps optional
- get steps and cfg_scale default values from command line


leejet:

PR #1268, opened 2026-02-10
fix: correct sdapi handling of reference images on txt2img

wbruna:


wbruna:
Superseded by #1262 .

PR #1272, opened 2026-02-11
Add libomp.so library to rocm artifact

superm1:
This comes from the `openmp` package on Arch but isn't there by default for Ubuntu unless a matching LLVM package happens to be installed.

As the point of the artifact is to run standalone, explicitly add it.


superm1:
Closing for now; I'm going to experiment with a different direction based upon what has happened with llamacpp CI.  Will come back with either this or another PR.

PR #1274, opened 2026-02-11
feat: option to enable vae tiling automatically for large images

wbruna:
This allows controlling the use of VAE tiling according to the requested image size: tiling will be enabled only for images larger than a given threshold.

Koboldcpp has this same feature. It's mainly useful when the size isn't known at launch time, like in the sd-server, and when the size comes from an input image.

The size is specified by a square image side: `--vae-tiling 768` means enabling tiling for images larger than 768x768 - mainly because it's easy to test a NxN image, and more meaningful than something like 'megapixels'. Another possibility could be specifying the size with a string like "768x768". A memory threshold would arguably be better, but harder to implement, since the exact usage depends a lot on the specific backend and flags.

To keep it compatible with current usage, and still avoid an extra flag, the value is optional: `--vae-tiling` alone is the same as `--vae-tiling 1`. `--vae-tiling 0` also means no tiling.

### DIFF
--- a/examples/cli/main.cpp
+++ b/examples/cli/main.cpp
@@ -750,7 +750,8 @@ int main(int argc, const char* argv[]) {
                     gen_params.pm_id_embed_path.c_str(),
                     gen_params.pm_style_strength,
                 },  // pm_params
-                ctx_params.vae_tiling_params,
+                ctx_params.get_tiling_params(gen_params.get_resolved_width(),
+                                             gen_params.get_resolved_height()),
                 gen_params.cache_params,
             };
 
@@ -776,7 +777,8 @@ int main(int argc, const char* argv[]) {
                 gen_params.seed,
                 gen_params.video_frames,
                 gen_params.vace_strength,
-                ctx_params.vae_tiling_params,
+                ctx_params.get_tiling_params(gen_params.get_resolved_width(),
+                                             gen_params.get_resolved_height()),
                 gen_params.cache_params,
             };
 
--- a/examples/common/common.hpp
+++ b/examples/common/common.hpp
@@ -475,6 +475,7 @@ struct SDContextParams {
     prediction_t prediction           = PREDICTION_COUNT;
     lora_apply_mode_t lora_apply_mode = LORA_APPLY_AUTO;
 
+    int vae_tiling_threshold             = 0;
     sd_tiling_params_t vae_tiling_params = {false, 0, 0, 0.5f, 0.0f, 0.0f};
     bool force_sdxl_vae_conv_scale       = false;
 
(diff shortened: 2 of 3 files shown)


stduhpf:
You could already do something like that by playing with `--vae-tile-size 48` (48 is 768/16) with vae tiling enabled. I guess it can be nice in some situations to have a way to change the threshold independantly from the tile size.

wbruna:
> You could already do somrthink like that by playing with `--vae-tile-size 48` (48 is 768/16) with vae tiling enabled. I guess it can be nice in some situations to have a way to change the threshold independantly from the tile size.

The main issue with this approach is for images right above the tile size. In this case, a 784x784 image, for instance, uses four 48x48 tiles with a huge overlap, where 34x34 would be enough to keep the overlap above 0.5 (and twice as fast).

But maybe reducing the tile size for dimensions with too few tiles (perhaps only 2, since above that the algorithm would probably just reduce the number of tiles anyway) would be a better approach. I can try to come up with a heuristic for that.

wbruna:
@stduhpf , maybe something like this?

```diff
diff --git a/src/stable-diffusion.cpp b/src/stable-diffusion.cpp
index c0ee118..1300b47 100644
--- a/src/stable-diffusion.cpp
+++ b/src/stable-diffusion.cpp
@@ -2494,8 +2494,17 @@ public:
             } else if (requested_size >= min_tile_dimension) {
                 tile_size = requested_size;
             }
-            tile_size = static_cast<int>(tile_size * encoding_factor);
-            return std::max(std::min(tile_size, static_cast<int>(latent_size)), min_tile_dimension);
+            tile_size      = static_cast<int>(tile_size * encoding_factor);
+            int max_tile_1 = static_cast<int>(latent_size);
+            if (tile_size >= max_tile_1) {
+                // single tile
+                tile_size = max_tile_1;
+            } else {
+                // avoid a high overlap factor with just two tiles
+                int max_tile_2 = static_cast<int>(std::ceil(max_tile_1 / (2 - tile_overlap)));
+                tile_size      = std::max(std::min(tile_size, max_tile_2), min_tile_dimension);
+            }
+            return tile_size;
         };
 
         tile_size_x = get_tile_size(params.tile_size_x, params.rel_size_x, latent_x);
```

With the default tile size 32 and overlap 0.5, a 512x528 image would use 32x22 tiles.

wbruna:
> With the default tile size 32 and overlap 0.5, a 512x528 image would use 32x22 tiles.

Tested it a bit further; at least on Klein 4b, the reduced tiles can cause noticeable color bands.

wbruna:
Closing; #1621 has a better approach.

PR #1276, opened 2026-02-12
fix: correct sdapi LoRA file handling

wbruna:
- open LoRA files through their discovered absolute paths
- do not resolve symlinks when building the LoRA list


leejet:

PR #1277, opened 2026-02-12
feat: add OptionFlags for ArgOptions; complete SDGenerationParams to_json/from_json conversion

donington:
This PR adds the functionality necessary to convert generic option param types to and from json using the ArgOptions.  To achieve this, it adds a number of features:

- adds `OptionFlags` to all of the Option types.  This implements the ability to mark options as assigned during command line parsing, making it easier to iterate through options manually set by the caller.  This also adds a no network tag to mark options that are not meant to be transferred over the network.
- Utilizing this, I added `options_to_json` and `options_from_json`.  These are designed to be generic, able to take an option parameter class and it's corresponding parsed ArgOptions.  The `to_json` variant makes use of the assigned flag to only focus on parameters that were specifically declared by the caller.  These functions operate on all of the basic ArgOptions types dynamically, converting the long_name into a json friendly expression matching the old `to_json_str` formatting (underscores instead of dashes).  To use this function with an option parameter class, it will need to implement `manual_options_to_json(std::vector<ManualOptions> const&, json&)` and `manual_options_from_json(std::vector<ManualOptions>&, json const&)` to finalize the conversion for the manual options.
- Finally, I added the manual options conversions into `SDGenerationParams`, as these are the parameters that are most relevant for network transfer to/from the client and server.  I also added `json to_json(ArgOptions const&)` and `void from_json(json const&)` convenience methods.

### DIFF
--- a/examples/common/common.hpp
+++ b/examples/common/common.hpp
@@ -7,6 +7,7 @@
 #include <sstream>
 #include <string>
 #include <vector>
+#include <unordered_set>
 
 #include <json.hpp>
 using json   = nlohmann::json;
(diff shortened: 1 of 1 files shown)


leejet:
Thanks for the PR. I understand the goal, but this adds a fairly large generic option/JSON serialization layer into `common.hpp` and mixes CLI parsing, assignment tracking, JSON conversion, and network-transfer policy together. For this project, I’d prefer a smaller and more explicit `SDGenerationParams` conversion only for the client/server parameters we actually need, so I’m going to close this for now.

PR #1280, opened 2026-02-14
Switch over ROCm builds to artifacts both for stable and preview rele…

superm1:
…ases

AMD is in the process of moving over to a brand new build system (TheRock). They have a preview release stream for it, and it includes some significant performance uplifts.

To let people experiment with both stacks generate artifacts both for the stable (7.2) and preview (7.11) stacks.

In both cases - ROCm must be installed first to use this artifact. It's intentionally not bundled because so many ISAs are included and the size of the artifact balloons to an untenable size.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -486,80 +486,71 @@ jobs:
             sd-${{ env.BRANCH_NAME }}-${{ steps.commit.outputs.short }}-bin-win-rocm-x64.zip
 
   ubuntu-latest-rocm:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


superm1:
The new artifacts are a much more [reasonable ~145MB.](https://github.com/superm1/stable-diffusion.cpp/releases/tag/master-505-e212912)  And a user won't end up with mutliple ROCm stacks on the machine now if they use other software like llama.cpp.

superm1:
I'm going to do other changes on my master branch so we shouldn't merge from here in case I break something, I'll split this into it's own branch and new PR.

PR #1303, opened 2026-03-01
Fix : gguf download link in anima.md

Juste-Leo2:
The GGUF models currently linked in the documentation were quantized for ComfyUI-GGUF. They are incompatible with this repository and cause the following error:

`[ERROR] stable-diffusion.cpp:335 - get sd version from file failed: ''`

I have updated the links to a repository containing GGUF files specifically quantized for `stable-diffusion.cpp`. I have tested them on the `main` branch and they work correctly.

I can add more quantization levels if there is a need for them :)

### DIFF
--- a/docs/anima.md
+++ b/docs/anima.md
@@ -4,7 +4,7 @@
 
 - Download Anima
     - safetensors: https://huggingface.co/circlestone-labs/Anima/tree/main/split_files/diffusion_models
-    - gguf: https://huggingface.co/Bedovyy/Anima-GGUF/tree/main
+    - gguf: https://huggingface.co/JusteLeo/Anima-GGUF/tree/main
 - Download vae
     - safetensors: https://huggingface.co/circlestone-labs/Anima/tree/main/split_files/vae
 - Download Qwen3-0.6B-Base


leejet:
Thanks! However, I think it would be better to implement compatibility for this kind of naming directly in the code. That way, users of ComfyUI won’t need to keep two copies of the weights locally. You can take a look at this PR https://github.com/leejet/stable-diffusion.cpp/pull/1305.

Juste-Leo2:
> Thanks! However, I think it would be better to implement compatibility for this kind of naming directly in the code. That way, users of ComfyUI won’t need to keep two copies of the weights locally. You can take a look at this PR #1305.

Glad I could help :) !
Quick question: is it normal that, for instance, the Q6_K quantization for ComfyUI GGUF doesn't have the same file size as the Q6_K quantization for stable-diffusion.cpp? I don't know enough about the internals yet, so I'm curious.

Juste-Leo2:
Since the compatibility fix is already merged, I'll close this PR as it's no longer needed.

leejet:
> is it normal that, for instance, the Q6_K quantization for ComfyUI GGUF doesn't have the same file size as the Q6_K quantization for stable-diffusion.cpp?

Because some weight types are different — for example, certain weights were converted to `f32` by Bedovyy, while yours are using `bf16`.

PR #1307, opened 2026-03-02
reset weight adapter for models if no loras in request (fix 'sticky loras')

Korsar13:
Currently, weight_adapter remains unchanged if there are no loras in the query.
Therefore, after a calculation with a given loras, all subsequent queries without a loras specified will use the last specified loras.


leejet:

PR #1310, opened 2026-03-02
feat: add ACE-Step-1.5 support 

rmatif:
This PR add initial support for [ACE-Step-1.5](https://github.com/ace-step/ACE-Step-1.5) 

CLI eg: 

```shell
PROMPT="$(cat prompt.txt)"; LYRICS="$(cat lyrics.txt)";
./build/bin/sd-cli \
  -M audio_gen \
  -m /models/ace_step_1.5_turbo_aio.safetensors \
  -p "$PROMPT" \
  --lyrics "$LYRICS" \
  --sampling-method euler --scheduler simple --cfg-scale 1 --flow-shift 3 --steps 8 \
  --duration 120 --timesignature 4 \
  --language en --keyscale "G major" --bpm 120 \
  --lm-seed 32 -s 32 --fa \
  -o output.wav
``` 

Generates 120s of audio in under 12s on an RTX 4090

Download the model: https://huggingface.co/Comfy-Org/ace_step_1.5_ComfyUI_files/blob/main/checkpoints/ace_step_1.5_turbo_aio.safetensors

Required ggml submodule for now: https://github.com/rmatif/ggml/tree/ace

Since this is a text-to-music model and it introduces significant changes and new code, I’m wondering whether it’s preferable to integrate it into stable-diffusion.cpp, or if it would be better for me to create a separate repo (e.g ace.cpp) dedicated to this. @leejet any thoughts?

Output example: 

https://github.com/user-attachments/assets/bced4897-f5e6-470e-8d31-b3ec41cf180a

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,4 @@
 [submodule "ggml"]
     path = ggml
-	url = https://github.com/ggml-org/ggml.git
+	url = https://github.com/rmatif/ggml.git
+	branch = ace


daniandtheweb:
First of all, great job, that's an amazing addition.
If I may, I'd like to share my thoughts about this.
Given the fact that the project already handles both image and video generation I think that adding audio generation could be an appropriate choice. Moreover the current biggest generation project, ComfyUI, already supports it along with the other standard generation modes.

wbruna:
@LostRuins has been adding support to Koboldcpp for the past few days, via [ServeurpersoCom/acestep.cpp](https://github.com/ServeurpersoCom/acestep.cpp).

LostRuins:
My 2c is that ace step is probably out of scope for this project being that it also requires an LM pass for the planner phase, and deals with audio rather than images/video. But I might be biased seeing that I've already got ace step working via ace-step.cpp, seems unnecessary to reinvent the wheel.

leejet:
Great job! However, I think this goes a bit beyond the intended scope of sd.cpp. It might be better to make it a separate repository, perhaps called ace-step.cpp.

rmatif:
Alright then I’ll close this PR. Since https://github.com/ServeurpersoCom/acestep.cpp exists (which I wasn’t aware of when I started working on this), I think I’ll drop this and may contribute there if I have something relevant to add

PR #1314, opened 2026-03-04
CUDA Dockerfile

Nauscar:
This PR adds a cuda variant to the github workflows `build.yml`.
Version `12.6.3` was used as it is still currently maintained and supports more devices.

This was tested locally using `podman` on Fedora Silverblue with the following commands:

```sh
podman build -f Dockerfile.cuda -t stable-diffusion.cpp:master-cuda  .
podman run --rm --userns=keep-id --security-opt=label=type=nvidia_container_t --device nvidia.com/gpu=all localhost/stable-diffusion.cpp:master-cuda --version
```

Thanks for all your efforts on stable-diffusion.cpp!

### DIFF
--- a/Dockerfile.cuda
+++ b/Dockerfile.cuda
@@ -1,0 +1,25 @@
+ARG CUDA_VERSION=12.6.3
+ARG UBUNTU_VERSION=24.04
+
+FROM nvidia/cuda:${CUDA_VERSION}-cudnn-devel-ubuntu${UBUNTU_VERSION} AS build
+
+RUN apt-get update && apt-get install -y --no-install-recommends build-essential git ccache cmake
+
+WORKDIR /sd.cpp
+
+COPY . .
+
+ARG CUDACXX=/usr/local/cuda/bin/nvcc
+RUN cmake . -B ./build -DSD_CUDA=ON
+RUN cmake --build ./build --config Release --parallel
+
+FROM nvidia/cuda:${CUDA_VERSION}-cudnn-runtime-ubuntu${UBUNTU_VERSION} AS runtime
+
+RUN apt-get update && \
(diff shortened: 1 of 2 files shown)


Dockerfile.cuda
```diff
@@ -0,0 +1,25 @@
...
+
+COPY . .
+
+ARG CUDACXX=/usr/local/cuda/bin/nvcc
+RUN cmake . -B ./build -DSD_CUDA=ON
+RUN cmake --build ./build --config Release --parallel
+
+FROM nvidia/cuda:${CUDA_VERSION}-cudnn-runtime-ubuntu${UBUNTU_VERSION} AS runtime
+
+RUN apt-get update && \
```
Green-Sky, line 18: I think you can drop the second update here.

Nauscar: Thanks for the comment, let me know if I misunderstood.

Without installing `libgomp1` I get the following:

```sh
/sd-cli: error while loading shared libraries: libgomp.so.1: cannot open shared object file: No such file or directory
```

If you were referring to removing only the update, I believe it is necessary to ensure the right package for `libgomp1` is selected.

Note that this is a new stage from `cudnn-runtime` and not a continuation of the above build stage.  This is also the same as the existing functionality in `Dockerfile.vulkan`.

Nauscar:
Something to note, none of the existing Dockerfiles have `ARG DEBIAN_FRONTEND=noninteractive`.  I didn't include this in case its ommision was intentional.

Dockerfile.cuda
Green-Sky: > Note that this is a new stage from cudnn-runtime and not a continuation of the above build stage.

My mistake, I have never seen this before, but I also don't work much with docker. This makes mow sense now. :)

PR #1318, opened 2026-03-04
chore: replace rand and srand at the library level

wbruna:
These functions have global state, so they could interfere with application behavior.

It would arguably be more correct to use `std::default_random_device`, but that seemed a bit overkill for this.


leejet:
In fact, the randomness of the seed is not actually important.

leejet:
Thanks for the PR, but I don’t think the benefit is strong enough for this project right now, so I’m going to close it.

wbruna:
No problem. I was keeping this around with the intention of extending it to report back all parameters set by the library (the final seed, resolved sampler, etc), but never reached something worth submitting.

PR #1321, opened 2026-03-05
🧪 Add tests for binary reading functions

torusvektor:
- Refactored `read_u64`, `read_int`, and `read_short` from `model.cpp` to `util.cpp` and `util.h` to make them testable.
- Improved the implementation of these functions with explicit types and safer casts.
- Added a new test suite in `tests/test_binary_reader.cpp` covering various scenarios and edge cases.
- Integrated the tests into the CMake build system.


leejet:
Thanks for the PR, but I don’t think this is necessary for the project right now, and it also includes some unrelated AI-generated test code, so I’m going to close it.

PR #1323, opened 2026-03-05
ci: change workflow owner of "actions-commit-hash" from "pr-mpt" to "prompt" (org renamed)

WinkelCode:
The organization changed their GitHub name from "pr-mpt" to "prompt", this commit ensures the intended workflow is called directly instead of relying on GitHub's automatic redirection.

**Note:** They registered a placeholder account with the old name, so there is no immediate takeover risk.


leejet:

PR #1327, opened 2026-03-06
refactor: move all cache parameter defaults to the library

wbruna:
Simplifies cache initialization for library users.


wbruna:
@rmatif and @leejet , the `--cache-preset` handling from 186038ec3add50731232a6c49360dcee3b58e87a apparently got discarded before #1066 got merged as 298b11069fbf33f8f471ffebc4b6ab3917fdf800 . The option is validated and written to the `cache_preset` field, but the value isn't used anywhere.

Fixing it isn't straightforward because the needed functions (`get_scm_preset`, `get_preset_threshold`, etc) on `cache-dit.hpp` are not part of the library API. So my idea is, on top of this PR here:
- add a new field for the preset to the `sd_cache_params_t` struct
- initialize the four relevant fields to specific invalid values to distinguish them from values explicitly set through `--cache-options` (I believe negatives would work);
- during inference, replace the negatives (and an empty scm array) with the appropriate values according to the preset.

But... since this bug went off the radar for over two months, I'm wondering if it's really worth the effort, or if we should just remove the flag instead. What do you think?

rmatif:
> But... since this bug went off the radar for over two months, I'm wondering if it's really worth the effort, or if we should just remove the flag instead. What do you think?

I'm not against dropping it

leejet:
If the flag hasn't been doing anything for a while and nobody noticed, I'm leaning toward just dropping it instead of adding more complexity to support it.

PR #1328, opened 2026-03-07
fix: sd-server uses lora_model_dir correctly

mostlygeek:
Pass ctx_params.lora_model_dir instead of "" for /v1/images/generations and /v1/images/edits handlers. The empty string caused extract_and_remove_lora() to return immediately.

Tested this locally and the DEBUG output now correctly has the lora

```
SDGenerationParams {
  loras: "{
    "/mnt/nvme/models/sd.cpp/flux/realism_lora_comfy_converted.safetensors": "1"
  }",
  high_noise_loras: "{
  }"
}
```

Whereas current `master` the loras would be an empty object.

### DIFF
--- a/examples/common/common.hpp
+++ b/examples/common/common.hpp
@@ -1657,12 +1657,18 @@ struct SDGenerationParams {
                 is_high_noise = true;
             }
 
-            fs::path final_path;
+            // Security: reject absolute paths from user input to prevent
+            // arbitrary file access.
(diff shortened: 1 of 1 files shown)


candrews:
This is a subset of https://github.com/leejet/stable-diffusion.cpp/pull/1156/ but without the path sensitization.

leejet:
Closing this PR since it duplicates #1156 .

PR #1336, opened 2026-03-12
feat: add generic DiT support to spectrum cache

rmatif:
Implementing the [paper](https://arxiv.org/abs/2603.01623) literally is heavier because it depends on architecture-specific internal feature hooks and final-block forecasting inside each DiT family, which would require invasive model-side changes and separate validation per model. This patch keeps Spectrum generic at the sampler level

Comparaison with easycache with 20 steps: 

| | Baseline | EasyCache | Spectrum |
|---|---|---|---|
| **Image** | Baseline | EasyCache | Spectrum |
| **Speedup** | — | x1.82 | x1.82 |
| **Config** | — | 0.2, 0.15, 0.95 | default |


JohnLoveJoy:
I'd say it's better than baseline. That ear looked like an artifact.

PR #1344, opened 2026-03-15
feat: add CUDA Dockerfile

DaveBben:
Adds a Dockerfile for CUDA builds. I needed it for my llama-swap setup. Follows same pattern as other Dockerfiles but uses nvidia base images.


leejet:
Superseded by https://github.com/leejet/stable-diffusion.cpp/pull/1314.

PR #1254, opened 2026-02-05
feat: add support for SDXS-09

akleine:
Even though the name "SDXS-09" is similar to "SDXS", it is completely different from SDXS. 
For this reason that "SDXS" was renamed to "SDXS (DS)", where DS stands for the DreamShaper edition by IDKiro.

PR #1356, opened 2026-03-18
feat: SDXS-09 support and update doc

akleine:
1. Download links have been provided instead of the conversion process.
2. The ckpt references have been replaced with those on SafeTensors. 

These changes were made because this file is intended for less experienced users. They are not required to use *.ckpt and can download the models instead of creating them themselves.


akleine:
After some updates to the document file (committed yesterday)
here comes again some code for supporting SDXS-09 .

We have some users who are eagerly awaiting this SDXS-09 patch, e.g. @ProgerXP .
But since February 5th, a lot of code has been changed, **leading to merge conflicts.** 

**Therefore, I have updated my old SDXS-09 code to reflect the current state.**
For this reason the old PR #1254 will be closed soon.

Also to better distinguish between both SDXS versions,  the "old" VERSION_SDXS is now 
called VERSION_SDXS_512_DS, where DS stands for the DreamShaper edition by IDKiro.

A general note: Although the name "SDXS-09" is similar to SDXS-512-DS, the model is completely different, 
but also similarly ultra-fast. 

This makes it interesting for many users. And unlike the SDXS-512-DS,  SDXS-09 generates more different images
for the same prompt, depending from used seed value.

walletkadens:
also interested in this feature
this version of sdxs has the best balance of quality / resources
allowing even free public image generation without much pretense
i can build this from fork but dont want to miss on upstream commits
please merge
thx

leejet:

ProgerXP:
Just tested it. @leejet, @akleine, you are my heroes! Thanks a lot for implementing this.

---

I'm curious. I'm comparing generation from @akleine's original branch and current master and notice some tiny differences (not only with this image, practically with any run):

diff

Full size [akleine](https://github.com/user-attachments/assets/f6c758c9-d19f-46fa-83a1-19b3f6db096c), [master](https://github.com/user-attachments/assets/121b59ca-4545-4a2d-8e41-3f120fcbfc52).

What's the cause? I'm using the same model file and parameters.

```
sd-cli -m sdxs.safetensors --cfg-scale 1 --steps 1 --seed 42 -p 'flock of colorful piegons'
```

PR #1361, opened 2026-03-23
Cli: Fix flf2v always making loops 

stduhpf:
It was loading the first frame again instead of the last one


leejet:
Superseded by https://github.com/leejet/stable-diffusion.cpp/pull/1373.

PR #1363, opened 2026-03-24
sampler improvements

wbruna:
This started as an attempt to make Euler Ancestral work with flow models, and kind of got out of control 🙈 In order, we have:
- the eta parameter now controls the noise levels for ancestral samplers (on which it defaults to 1.0, so I had to change its initialization);
- a giant refactor replacing raw loops with iteration functions, like this:
```diff
     float dt = sigma_down - sigma;
-    // x = x + d * dt
-    {
-        float* vec_d = (float*)d->data;
-        float* vec_x = (float*)x->data;
-        for (int i = 0; i < ggml_nelements(x); i++) {
-            vec_x[i] = vec_x[i] + vec_d[i] * dt;
-        }
-    }
+    denoiser_tensor_iter(x, d, [dt](float& x, const float& d) {
+        x = x + d * dt;
+    });
```
- found what looks like a bug in the iPNDM_v implementation, thanks to the refactor above (I'll make a separate PR for it later)
- _another_ giant refactor moving each sampler implementation to its own function, replacing the thousand-lines switch
- the actual implementations for Euler Ancestral and DPM++(2s) Ancestral for flow models.

Now I'm looking at #1354, and concluding we should have implementation objects instead (like the denoiser code, turning the switch inside `sample_k_diffusion` into a factory function), so that each sampler would be able to report back things like "does this work with unconditioned model outputs"? It would also make it easier to share code between related samplers.

Also, it may make sense to move everything into a denoiser.cpp, to help with build times...

So, before I reach an even larger and harder to review diff, I'd appreciate some feedback from people who have dealt with the sampler code before 🙂 Are these sensible changes? Shall I instead do all refactors first and all new implementations afterwards? Separate PRs?

### DIFF
--- a/src/denoiser.hpp
+++ b/src/denoiser.hpp
@@ -781,6 +781,27 @@ static void generate_ancestral_step(float& sigma_up, float& sigma_down, float si
     }
 }
 
+static void generate_rf_ancestral_step(float& sigma_down, float& scale_factor, float& renoise_coeff, float sigma_from, float sigma_to, float eta) {
+    // downstep_ratio = 1 + (sigma_to / sigma_from - 1) × eta
+    // sigma_down     = sigma_to × downstep_ratio
+    // scale_factor   = (1 - sigma_to) / (1 - sigma_down)
+    // renoise_coeff  = √(sigma_to² - sigma_down² × scale_factor²)
+
+    float downstep_ratio = 1.0f + (sigma_to / sigma_from - 1.0f) * eta;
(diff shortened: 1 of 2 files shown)


wbruna:
Closing this PR: similar refactors were done by leejet in the master branch, and I've separated the ancestral samplers implementation into other PRs: #1364 and #1374 . If I end up reorganizing the samplers into implementation objects, I'll reopen this or create another one.

PR #1364, opened 2026-03-24
feat: add support for the eta parameter to ancestral samplers

wbruna:
Applies the `eta` parameter to the DPM++(2s) and Euler ancestral implementations, so the amount of injected noise can be adjusted (e.g. Euler A with `eta=0` should be the same as Euler). It reuses the calculation from the RES samplers, so it's mostly a refactor and interface/UI adjustments.

#1363 already includes this, but since this is self-contained and useful on its own, I believe it's worth including directly.


wbruna:
Superseded by #1374 .

PR #1366, opened 2026-03-25
chore: clean up unused variables in ipndm_v implementation

wbruna:
Since the calculations in the reference implementation are _way_ more complex, I believe the use of those constant factors is intentional. So, just clean up the `h_n_2` and `h_n_3` unused variables.


wbruna:
Superseded by #1373

PR #1383, opened 2026-03-31
refactor: centralize CompVisDenoiser table calculations

wbruna:


src/stable-diffusion.cpp
```diff
@@ -73,19 +73,9 @@ const char* sampling_methods_str[] = {
...
-    float ls_sqrt = sqrtf(linear_start);
-    float le_sqrt = sqrtf(linear_end);
-    float amount  = le_sqrt - ls_sqrt;
-    float product = 1.0f;
-    for (int i = 0; i < timesteps; i++) {
-        float beta = ls_sqrt + amount * ((float)i / (timesteps - 1));
-        product *= 1.0f - powf(beta, 2.0f);
-        alphas_cumprod[i] = product;
-    }
+void calculate_alphas_cumprod(float* alphas_cumprod) {
```
wbruna, line 79: I've preserved this, but I'm not sure the `alphas_cumprod` tensor is really being used.

src/stable-diffusion.cpp
```diff
@@ -999,13 +989,6 @@ class StableDiffusionGGML {
                 }
             }
 
-            auto comp_vis_denoiser = std::dynamic_pointer_cast<CompVisDenoiser>(denoiser);
```
leejet, line 1002: This change removes the initialization path from `alphas_cumprod_tensor`.

Previously, `CompVisDenoiser` could be initialized using the model-provided `alphas_cumprod`, but now it relies entirely on a hardcoded schedule.

This means we can no longer use `alphas_cumprod_tensor` to initialize the denoiser, which changes the original behavior.

leejet (changes requested):

src/stable-diffusion.cpp
wbruna: The diff doesn't show this, but the tensor is always filled by `calculate_alphas_cumprod` unconditionally at line ~802:

```c++
calculate_alphas_cumprod((float*)alphas_cumprod_tensor->data);
```

and _then_ propagated to the `comp_vis_denoiser` at this point. Unless the idea is that the tensor could be _overwritten_ by the model when loading?

leejet: Yes, in the original code path, if the model contains `alphas_cumprod`, it would override the default values.

PR #1400, opened 2026-04-07
fix: Use PkgConfig for WebP and WebM

### LINKED ISSUE #1399: Don't require cmake files when using system webp and webm, use PkgConfig instead
libwebp and libwebm system libraries can used instead of the bundled/vendor copies, which is great.

However, the CMake files must be installed on the system to use the system libraries, and those files are not always installed. Specifically, libwebp can be built using autotools or CMake, and since the CMake files are only installed when using CMake , if libwebp is built using autotools, no cmake files are installed.

For example, Gentoo doesn't install the CMake files for webp because it uses autotools for this package. See https://bugs.gentoo.org/937031 for details.

The standard way to use libraries is with PkgConfig - this approach will always work. For libwebp: `pkg_check_modules(WebP REQUIRED libwebp)`

FYI @wbruna

candrews:
PkgConfig is the standard approach to locating and using system libraries. It does not rely on the library having been built with CMake, improving compatibility.

webpmux is the library name; it should not be prefixed with lib. See: https://github.com/webmproject/libwebp/commit/bfad7ab589a997a69ba568a7c5d9696474f917fc

Closes: https://github.com/leejet/stable-diffusion.cpp/issues/1399


wbruna (changes requested):
Pretty sure we shouldn't assume `pkg-config` will always be available, e.g. on Windows. I believe it would be best to call `pkg_check_modules` as a fallback, in case the CMake module isn't found.

candrews:
> Pretty sure we shouldn't assume `pkg-config` will always be available, e.g. on Windows. I believe it would be best to call `pkg_check_modules` as a fallback, in case the CMake module isn't found.

That's a good point. I've done as you suggested, only use PkgConfig if it's available.

leejet:
This change appears to have broken CI (link error with `-lwebpmux`). Please fix the build regression first.

Also, I think there are two issues in the current implementation:

1. **Fallback logic is still incorrect**
   The code only falls back when `find_package(PkgConfig)` is not found. However, a much more common failure case is when `pkg-config` exists but the required `.pc` files are missing or mismatched.
   Using `pkg_check_modules(... REQUIRED ...)` will fail configuration immediately in that case, so the fallback path is never reached.

2. **Missing `libwebpmux` in pkg-config path**
   The pkg-config branch only resolves `libwebp`, but `webpmux` is a separate module upstream (`libwebpmux.pc`).
   So `PkgConfig::WebP` may only provide `-lwebp`, while the examples still require `webpmux` to be resolved and linked separately.

candrews:
> This change appears to have broken CI (link error with `-lwebpmux`). Please fix the build regression first.

Done! I added an alias so libwebpmux and webpmux both work in all cases. I undid the modifications to examples/{cli,server}.

> Also, I think there are two issues in the current implementation:
> 
>     1. **Fallback logic is still incorrect**
>        The code only falls back when `find_package(PkgConfig)` is not found. However, a much more common failure case is when `pkg-config` exists but the required `.pc` files are missing or mismatched.
>        Using `pkg_check_modules(... REQUIRED ...)` will fail configuration immediately in that case, so the fallback path is never reached.

I've addressed that by trying the CMake approach first, then the PkgConfig approach. I also used [`cmake_pkg_config`](https://cmake.org/cmake/help/latest/command/cmake_pkg_config.html) to eliminate the use of the PkgConfig executable and simplify the logic.

>     2. **Missing `libwebpmux` in pkg-config path**
>        The pkg-config branch only resolves `libwebp`, but `webpmux` is a separate module upstream (`libwebpmux.pc`).
>        So `PkgConfig::WebP` may only provide `-lwebp`, while the examples still require `webpmux` to be resolved and linked separately.

I added webpmux. This wasn't necessary on my system, but it makes sense to have it, and it still works.

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.12)
+cmake_minimum_required(VERSION 4.1)
```
wbruna: CMake 4.1 seems to be [younger than a year](https://gitlab.kitware.com/cmake/cmake/-/commits/v4.1.0?ref_type=tags); this would exclude most current Linux distros (e.g. Debian Stable has 3.31.6).

wbruna (changes requested):

CMakeLists.txt
candrews: I changed back to using pkg_check_modules instead of cmake_pkg_config, reducing the CMake requirement back to the original 3.12.

CMakeLists.txt
```diff
@@ -108,19 +108,28 @@ if(SD_WEBP)
...
-            message(FATAL_ERROR
-                "Could not find a compatible webpmux target in system WebP package. "
-                "Expected WebP::libwebpmux or WebP::webpmux."
-            )
+            find_package(PkgConfig REQUIRED)
+            pkg_check_modules(WebP REQUIRED IMPORTED_TARGET GLOBAL libwebp)
+            pkg_check_modules(WebPMux REQUIRED IMPORTED_TARGET GLOBAL libwebpmux)
+            link_libraries(PkgConfig::WebP)
+            link_libraries(PkgConfig::WebPMux)
+            add_library(libwebpmux ALIAS PkgConfig::WebP)
```
wbruna: This looks a bit funny: is libwebpmux really supposed to alias webp non-mux, just for this case?

wbruna (commented):
Apart from the libwebpmux alias, I don't see any obvious issues (but I didn't test it).

CMakeLists.txt
candrews: I fixed it.
This mismatch didn't cause any problems in my testing, so I didn't catch it. Thank you for catching it 👍

candrews:
@leejet could you please take a look?

I've been using this PR for a while without issue.

candrews:
@wbruna @leejet can you please take a peek at this PR?

leejet:

PR #1403, opened 2026-04-08
feat: add er_sde sampler

rmatif:
Add support for the er_sde sampler 

Reference: https://github.com/Comfy-Org/ComfyUI/blob/master/comfy/k_diffusion/sampling.py#L1500


src/denoiser.hpp
```diff
@@ -1295,6 +1322,112 @@ static sd::Tensor<float> sample_res_2s(denoise_cb_t model,
...
 
+static sd::Tensor<float> sample_er_sde(denoise_cb_t model,
+                                       sd::Tensor<float> x,
+                                       std::vector<float> sigmas,
+                                       std::shared_ptr<RNG> rng,
+                                       bool is_flow_denoiser) {
+    constexpr int max_stage                  = 3;
+    constexpr int num_integration_points     = 200;
+    constexpr float num_integration_points_f = 200.0f;
+    constexpr float s_noise                  = 1.0f;
```
wbruna: Since we don't have a separate `s_noise` parameter, maybe this could reuse `eta`?

src/denoiser.hpp
```diff
@@ -818,6 +818,33 @@ static std::tuple<float, float, float> get_ancestral_step_flow(float sigma_from,
...
+        return 1.0f - sigma;
+    }
+    return 1.0f;
+}
+
+static float er_sde_noise_scaler(float x) {
+    x = std::max(x, 0.0f);
+    return x * (std::exp(std::pow(x, 0.3f)) + 10.0f);
+}
+
```
wbruna: If these won't be reused for other samplers, I'd suggest keeping them as lambdas inside `sample_er_sde`, or at least close to that function.

PR #1418, opened 2026-04-14
move generate + image format and base64 encoding into new function  generate_and_encode()

Korsar13:


leejet:
Thanks for the PR. The safe part to extract is mostly just encoding/base64, which would only reduce a small amount of duplication. The current PR reduces more code by also moving generation into the helper, but that changes route/job control flow and error handling. Given that tradeoff, I’d prefer to keep the current code explicit and close this for now.

PR #1422, opened 2026-04-15
fix: update space handling in token_split and improve ltrim/rtrim fun…

maikebing:


wbruna (changes requested):
This PR shouldn't change the ggml submodule.

The change itself look reasonable, though, but I had to dig up a bit to understand it. A better explanation on the commit message would have been appreciated, because it's not an 'update': we have what looks like the wrong function being called on the tokenizer. Even more so, because the fix for the standard `isspace` theoretically behaving badly _looks_ related, but it's actually a totally separate issue.

PR #1436, opened 2026-04-18
feat: adapt res samplers for flow models for eta > 0

wbruna:
This adjusts the noise addition steps to follow the flow math. I believe the deterministic steps do not need further changes, apart from the sigma_up and sigma_down values.

@rmatif , could you please take a look?

### DIFF
--- a/src/denoiser.hpp
+++ b/src/denoiser.hpp
@@ -1259,6 +1259,7 @@ static sd::Tensor<float> sample_res_multistep(denoise_cb_t model,
                                               sd::Tensor<float> x,
                                               const std::vector<float>& sigmas,
                                               std::shared_ptr<RNG> rng,
+                                              bool is_flow_denoiser,
                                               float eta) {
     sd::Tensor<float> old_denoised = x;
     bool have_old_sigma            = false;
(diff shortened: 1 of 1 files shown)


clort81:
I'm confused: existing get_ancestral_step is this:
```
    static std::pair<float, float> get_ancestral_step(float sigma_from, float sigma_to, float eta = 1.0f) {
    float sigma_up   = 0.0f;
    float sigma_down = sigma_to;
            
    if (eta <= 0.0f) {
        return {sigma_down, sigma_up};
    }   
    
    float sigma_from_sq = sigma_from * sigma_from;
    float sigma_to_sq   = sigma_to * sigma_to;
    if (sigma_from_sq > 0.0f) {
        float term = sigma_to_sq * (sigma_from_sq - sigma_to_sq) / sigma_from_sq;
        sigma_up   = std::min(sigma_to, eta * std::sqrt(std::max(term, 0.0f)));
    }   
    
    float sigma_down_sq = sigma_to_sq - sigma_up * sigma_up;
    sigma_down          = sigma_down_sq > 0.0f ? std::sqrt(sigma_down_sq) : 0.0f;
    return {sigma_down, sigma_up};
}
```
Does your pull request upgrade this function?

wbruna:
> Does your pull request upgrade this function?

Kind of: when running a flow model, it switches the samplers to use the similar `get_ancestral_step_flow`, and apply the additional signal scale. It's the same change that was done in #1374 for Euler Ancestral.

Note that `eta=0` produces the same results on the two variants, and it's the default for these samplers, so you'll only notice a difference if you change it:

| eta | master | PR |
|:---:|:---:|:---:|
| 0.0 | test_1776809473 |test_1776809935 |
| 1.0 | test_1776809502 | test_1776809976|

PR #1446, opened 2026-04-20
fix: build with GGML_BACKEND_DL=ON

Korsar13:
GGML_BACKEND_API is unavailable with GGML_BACKEND_DL=ON
changed to use GGML_API only

### DIFF
--- a/src/convert.cpp
+++ b/src/convert.cpp
@@ -103,7 +103,7 @@ bool convert(const char* input_path,
     bool output_is_safetensors = ends_with(output_path, ".safetensors");
     TensorTypeRules type_rules = parse_tensor_type_rules(tensor_type_rules);
 
-    auto backend    = ggml_backend_cpu_init();
+    auto backend    = ggml_backend_init_by_type(GGML_BACKEND_DEVICE_TYPE_CPU, nullptr);
     size_t mem_size = 1 * 1024 * 1024;  // for padding
     mem_size += model_loader.get_tensor_storage_map().size() * ggml_tensor_overhead();
     mem_size += model_loader.get_params_mem_size(backend, type);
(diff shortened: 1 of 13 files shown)


wbruna (changes requested):
Some backends currently need `#ifdef` tests.

Anyway, this will conflict with #1184 / #1368 , so I'd recommend waiting a bit for that effort.

Cyberhan123:
This PR should ideally avoid forcibly offloading weights to the CPU; this approach is premature and crudely assigning APIs to the CPU won't solve the problems associated with using shared ggml.

This PR https://github.com/leejet/stable-diffusion.cpp/pull/1368 includes that part. I've already released the SDK via https://github.com/seasonjs/stable-diffusion.cpp-build.

You can combine them via https://github.com/seasonjs/ggml.cpp-build.

By the way, let me promote my project. It has already integrated with llama.cpp, whisper.cpp, and diffusion.cpp. Yes, they all share a single ggml library and automatically load the backend.

See: https://github.com/Cyberhan123/slab.rs

The reason I haven't advertised it is because I don't think it meets my standards yet, but the models tested can generally be configured using `bun run models:pack` and directly imported into the `.slab` configuration file for loading and inference.

https://github.com/Cyberhan123/slab.rs/releases/download/v0.1.0-beta3/Slab_0.1.0_x64-offline-setup.exe This is the Windows link; it's previewable, but many features haven't been incorporated.

wbruna:
@Korsar13 , could you check if #1448 would be enough?

Korsar13:
> @Korsar13 , could you check if #1448 would be enough?

IMHO, https://github.com/leejet/stable-diffusion.cpp/pull/1448 is completely superior to this PR

PR #1464, opened 2026-04-28
Add Nvidia Spark GB10 Docker image

ich777:
- Add Dockerfile `Dockerfile.cuda-spark` with optimized build flags
- Modify build/publish matrix to support arm64 build for Nvidia Spark GB10

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -177,7 +177,7 @@ jobs:
   build-and-push-docker-images:
     name: Build and push container images
     if: ${{ github.event_name != 'pull_request' }}
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.runner }}
 
     permissions:
       contents: read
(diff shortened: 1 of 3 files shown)


ich777:
@leejet is that something that can be implemented?
It would be super cool to also have an image for the Nvidia Spark GB10

leejet:
I think the idea makes sense, but I don’t think this needs a separate `Dockerfile.cuda-spark` in its current form. From the diff, the Spark variant is still using the same CUDA backend and build flow as the existing CUDA image; the main differences appear to be build parameters:

- `platform=linux/arm64`
- `CUDA_VERSION=13.0.0`
- `UBUNTU_VERSION=24.04`
- `CMAKE_CUDA_ARCHITECTURES=121`
- `GGML_CUDA_FA_ALL_QUANTS=ON`

So this looks more like a CUDA build variant for DGX Spark / GB10 rather than a distinct Dockerfile target that needs to be maintained separately.

I would suggest parameterizing the existing `Dockerfile.cuda` instead of duplicating most of it into `Dockerfile.cuda-spark`. For example, `Dockerfile.cuda` could expose build args such as:

```dockerfile
ARG CUDA_VERSION=...
ARG UBUNTU_VERSION=...
ARG CUDA_ARCHITECTURES=...
ARG GGML_CUDA_FA_ALL_QUANTS=...

ich777:
> So this looks more like a CUDA build variant for DGX Spark / GB10 rather than a distinct Dockerfile target that needs to be maintained separately.

Oh okay, I can change that.
My initial thought for a dedicated file was that the Spark uses a somewhat unique combination (SM 121 + CUDA 13.0, where other targets use either SM 120 or SM 121 with CUDA 13.1+) but you are probably right, I'll change the PR if that would the repo more nicely.

PR #1470, opened 2026-05-01
Add tensor splitting (row + tensor), lazy loading and autofitting logic

pwilkin:
Add logic + commandline arguments to handle automatic fitting, lazy loading (only load DiT after conditioner finishes) and row + tensor splits (row splits are actually, unlike in core llama.cpp, recommended here since they save much more VRAM). Also does automatic calculation of VAE tiling.

Requires a GGML patch (attached in `ggml.patch`).

Allowed me to run @leejet 's branch of LTX-2 on my 16 GB 5060 + 10 GB 3080 fully on VRAM with Gemma 3 12B Q8_K_XL + LTX-2 DiT Q6_K.

Supersedes #1184 (besides the RPC support).


examples/common/common.cpp
```diff
@@ -409,18 +474,6 @@ ArgOptions SDContextParams::get_options() {
          "--mmap",
          "whether to memory-map model",
          true, &enable_mmap},
-        {"",
-         "--control-net-cpu",
```
wbruna, line 470: As mentioned in https://github.com/leejet/stable-diffusion.cpp/pull/1184#issuecomment-4207278746 , we should keep old flags as aliases.

Maybe they could be hidden from the normal help to avoid too much clutter, and be shown only with `-v/--verbose`?

examples/common/common.cpp
```diff
@@ -461,6 +514,30 @@ ArgOptions SDContextParams::get_options() {
          "--chroma-enable-t5-mask",
          "enable t5 mask for chroma",
          true, &chroma_use_t5_mask},
+        {"",
+         "--auto-fit",
+         "automatically pick DiT/VAE/Conditioner device placements based on "
+         "free GPU memory (default ON)",
+         true, &auto_fit},
```
wbruna, line 541: Would this include Vulkan backends with a main dGPU and an iGPU? If so, auto-fit by default may not be a good idea.

src/ggml_extend.hpp
```diff
@@ -2018,6 +2146,45 @@ struct GGMLRunner {
...
+        if (pending != nullptr) {
+            g_pending_multi_backend_spec() = nullptr;
+            multi_backend_mode             = true;
+            multi_backend_kind             = pending->mode;
+            additional_backends            = pending->additional_backends;
+            tensor_backend_fn              = pending->tensor_backend_fn;
+            cpu_fallback_backend           = pending->cpu_fallback;
+            row_split_ratios               = pending->tensor_split_ratios;
+            row_main_device                = pending->main_device;
+#ifdef SD_USE_CUDA
```
wbruna: We have _just_ removed all these ifdefs. Please switch to `sd_backend_is`.

src/util.cpp
```diff
@@ -174,12 +174,33 @@ bool is_directory(const std::string& path) {
...
 
     ~MmapWrapperImpl() override {
+#ifdef __linux__
+        // Drop the kernel pagecache pages for this file. madvise(DONTNEED)
+        // alone only unmaps from the process address space; pagecache
+        // entries persist (`free` reports them as buff/cache and the OOM
+        // killer doesn't touch them, but they ARE counted against
+        // overcommit and can starve other allocations on tight-RAM
+        // systems). posix_fadvise(POSIX_FADV_DONTNEED) is the documented
+        // way to evict pagecache for a specific fd's pages.
```
wbruna, line 225: I could add this to #1414 . In fact, maybe that PR could even simplify the lazy-loading? I could adapt it if needed (or at least so we don't step in each other's toes).

src/version.cpp
```diff
@@ -18,3 +21,12 @@ const char* sd_commit(void) {
 const char* sd_version(void) {
     return STRINGIZE(SDCPP_BUILD_VERSION);
 }
+
+void sd_list_devices(void) {
```
wbruna: This is the wrong place to do this: `version.cpp` is supposed to _only_ bundle binary version information, so that it can be easily replaced if the library is built without it. I'd move it to `util.cpp` (there is also an overlap with a log over there)

CMakeLists.txt
```diff
@@ -72,6 +72,7 @@ option(SD_USE_SYSTEM_GGML            "sd: use system-installed GGML library" OFF
 if(SD_CUDA)
     message("-- Use CUDA as backend stable-diffusion")
     set(GGML_CUDA ON)
+    add_definitions(-DSD_USE_CUDA)
```
wbruna: See https://github.com/leejet/stable-diffusion.cpp/pull/1470/changes#r3174514043 .

src/stable-diffusion.cpp
```diff
@@ -373,19 +576,353 @@ class StableDiffusionGGML {
...
+        // - share_devices/share_bytes: per-device share order from auto-fit
+        //   (largest first by descending share). The first device is the
+        //   "main" CUDA device, where the compute buffer lives.
+        // Returns true on success; populates out_spec.tensor_split_ratios
+        // with a vector of length total CUDA device count.
+        auto prepare_row_split_spec = [&](const std::vector<std::string>&     share_devices,
+                                          const std::vector<int64_t>&         share_bytes,
+                                          std::vector<ggml_backend_t>&        out_extra_backends,
+                                          MultiBackendSpec&                   out_spec) -> bool {
+#ifdef SD_USE_CUDA
```
wbruna: See https://github.com/leejet/stable-diffusion.cpp/pull/1470/changes#r3174514043 .

wbruna (changes requested):

pwilkin:
Yeah, that's a problem with CC, it tramples code style consistency. I'll refactor.

examples/common/common.cpp
pwilkin: To be honest, I'd still have to check how well this works on Vulkan (I don't think I have a setup with an iGPU to test though), but the idea is the algorithm is supposed to take those types of quirks into account. For now I only tested on my CUDA setup (also have to fix row-split to be supported on Vulkan if possible).

pwilkin: I do have a Vulkan SDK setup FWIW and I've build llama.cpp with Vulkan already, so I'll setup the Vulkan build and test next.

examples/common/common.cpp
```diff
@@ -380,6 +380,54 @@ ArgOptions SDContextParams::get_options() {
          "--upscale-model",
          "path to esrgan model.",
          &esrgan_path},
+        {"",
```
leejet, line 422: I’d prefer not to add one CLI/API field per model component. This may become hard to maintain as more components are added.

Could we use a generic placement syntax instead, e.g.:

  --backend 'clip=CPU,vae=CUDA0,diffusion=Vulkan0'

and parse it in `new_sd_ctx` into a map from component name to backend/device?

That should scale better, keep the API surface smaller, and make future components easier to support without adding new flags every time.

examples/common/common.cpp
```diff
@@ -392,6 +440,23 @@ ArgOptions SDContextParams::get_options() {
          "--chroma-t5-mask-pad",
          "t5 mask pad size of chroma",
          &chroma_t5_mask_pad},
+        {"",
```
leejet, line 462: Similar concern here: this introduces multiple --fit-* flags per component (dit/vae/cond).

Could we consider a generic form instead, for example:

  --fit-compute-reserve dit=...,vae=...,cond=...

and parse it into a component -> value map?

This would keep the CLI/API more scalable and consistent, and avoid growing a large number of per-component flags as new modules are added.

leejet (changes requested):

pwilkin:
Just wanna say I'm sorry I've been slacking on this, going to try to rebase this on top of @wbruna 's optimizations with mmap some time this week.

pwilkin:
@leejet @wbruna Okay, so I *finally* managed to get this mess resolved and integrated with the solutions in master.

This was tougher than expected, fortunately Anthropic releasing Fable helped ;) got the entire pipeline working again with the entire LTX-2 fully on GPU on my 26GB VRAM setup (Q8 Gemma, Q6 LTX-2.5). Managed to make it work this time without any hacks in the GGML code, so that's an extra plus.

OtherCrashOverride:
Would you mind posting an example command line of this working? Monitoring with 'nvtop', it never uses more than one GPU for DIT.  It also crashes out at the end.

```
[INFO ] stable-diffusion.cpp:783  - Version: LTXAV 
[INFO ] backend_fit.hpp:651  - auto-fit plan (margin=512 MiB per GPU):
[INFO ] backend_fit.hpp:652  -   available devices:
[INFO ] backend_fit.hpp:657  -     CUDA0        NVIDIA GeForce RTX 5060 Ti       free  15740 /  15888 MiB
[INFO ] backend_fit.hpp:657  -     CUDA1        NVIDIA GeForce RTX 5060 Ti       free  15576 /  15886 MiB
[INFO ] backend_fit.hpp:662  -   components:
[INFO ] backend_fit.hpp:664  -     DiT          params  12127 MiB, compute reserve   2048 MiB
[INFO ] backend_fit.hpp:664  -     VAE          params   1384 MiB, compute reserve   1024 MiB
[INFO ] backend_fit.hpp:664  -     Conditioner  params   9286 MiB, compute reserve    512 MiB
[INFO ] backend_fit.hpp:669  -   decisions:
[INFO ] backend_fit.hpp:675  -     DiT          -> GPU 1              (VRAM 14175 MiB)
[INFO ] backend_fit.hpp:675  -     VAE          -> GPU 0              (VRAM 2408 MiB)
[INFO ] backend_fit.hpp:675  -     Conditioner  -> GPU 0              (VRAM 9798 MiB)
[INFO ] backend_fit.hpp:696  -   projected per-device peak:
[INFO ] backend_fit.hpp:701  -     CUDA0        peak   9798 /  15740 MiB free  (remaining 5941 MiB)
[INFO ] backend_fit.hpp:701  -     CUDA1        peak  14175 /  15576 MiB free  (remaining 1401 MiB)
[INFO ] backend_fit.hpp:707  -     CPU          host RAM additional 0 MiB
[INFO ] stable-diffusion.cpp:956  - auto-fit: backend spec 'diffusion=CUDA1,te=CUDA0,vae=CUDA0', params backend spec '(none)'
```
I manually used this backend spec in master prior to this patch set and it works.  I can see CUDA0 and CUDA1 being used at different stages, but never at the same time just as before.

The new error that doesnt happen in master:
```
[DEBUG] stable-diffusion.cpp:6183 - decode audio latent 16x52x8x1
[DEBUG] ggml_extend.hpp:2131 - ltx_audio_vae compute buffer size: 126.52 MB(VRAM)
[INFO ] ggml_extend.hpp:2369 - ltx_audio_vae offload params (339.92 MB, 1285 tensors) to runtime backend (CUDA0), taking 0.13s
[ERROR] ggml_extend.hpp:70   - ggml_cuda_compute_forward: IM2COL failed
[ERROR] ggml_extend.hpp:70   - CUDA error: invalid argument
```

pwilkin:
@OtherCrashOverride could you please post your command that triggers this crash? I'm wondering if this is a real problem with IM2COL or some variant of OOM.

leejet (changes requested):
Thanks a lot for the contribution and for continuing to rebase this on top of the latest master.

Since `master` has recently moved to the new weight / model management module, there are now quite a few related branches and overlapping changes. I tried doing a merge locally to make the diff easier to compare against the current codebase. However, I have **not tested that merge**, so I am not fully sure whether it introduces any issue. If you run into problems caused by that merge, feel free to revert the related merge commits.

At the moment, this PR is still very large and touches several major areas at once, which makes it difficult for me to review and merge safely. Also, since the current `ModelManager` implementation already supports lazy loading, I think the duplicated lazy-loading parts in this PR can be removed or greatly simplified.

I would suggest splitting the remaining functionality into three smaller PRs:

### PR 1: ModelManager-based layer split + ggml runner scheduling

This PR should focus on implementing layer split and `ggml` runner scheduling on top of the existing `ModelManager`.

For backend selection, I would prefer using the generic `--backend` syntax instead of adding per-component CLI/API fields. For example:

```bash
--backend "dit=cuda0&cuda1"
```

This means the DiT component should use both `cuda0` and `cuda1`.

This first PR should only support explicit/manual placement. No auto-fit logic is needed here. The goal is to make the lower-level multi-device execution mechanism easy to review and test first.

### PR 2: split buffer support on top of PR 1

After PR 1 is merged, a second PR can add split buffer support. I think this should be built around the current backend / model manager design instead of adding another parallel allocation path.

A possible implementation direction:

1. **Create split buffer type at the backend layer**

   Around `src/core/ggml_extend_backend.cpp:257`, add a helper to `SDBackendManager`:

   * get the device from `ggml_backend_get_device(backend)`

   * get the registry from `ggml_backend_dev_backend_reg(dev)`

   * for CUDA, get the proc with:

     ```cpp
     ggml_backend_reg_get_proc_address(reg, "ggml_backend_split_buffer_type")
     ```

   * find the current device index inside that registry and use it as `main_device`

   * call:

     ```cpp
     fn(main_device, tensor_split.data())
     ```

   I would start with **CUDA only**. SYCL appears to expose a different proc signature, so I would avoid trying to generalize this in the first version.

2. **Let `ModelManager` own the split buffer type**

   In `src/model_manager.h`, add something like:

   ```cpp
   void set_split_buffer_type(ggml_backend_t backend,
                              ggml_backend_buffer_type_t split_buft);
   ```

   Internally this can be stored as:

   ```cpp
   std::map<ggml_backend_t, ggml_backend_buffer_type_t>
   ```

3. **Update buffer type selection**

   The current entry point seems to be around `src/model_manager.cpp:680`.

   Current behavior is roughly:

   * if params backend != compute backend, use the compute device host buffer
   * otherwise use the params backend default buffer

   I think this should become:

   * if the tensor is eligible for split and the target is the compute backend, return the split buffer type
   * otherwise keep the existing behavior

4. **Handle allocation during params-backend CPU / streaming load**

   Around `src/model_manager.cpp:228`, the code currently uses:

   ```cpp
   ggml_backend_alloc_ctx_tensors(..., compute_backend)
   ```

   which always uses the default buffer type.

   If we want split buffer to work when `--params-backend cpu` is used, or during streaming load, tensors should probably be grouped by buffer type. The split group can then be allocated with:

   ```cpp
   ggml_backend_alloc_ctx_tensors_from_buft(staging_ctx, split_buft)
   ```

   while the rest keeps the default path.

5. **LoRA handling**

   `src/model_manager.cpp:319` currently applies LoRA by building `add_inplace` / `cpy` graphs over model tensors. Split tensors participating in non-matmul ops may cause issues.

   For the MVP, I suggest either disabling this path when split buffer is enabled, or forcing LoRA to use runtime apply mode instead of modifying split model tensors directly.

### PR 3: auto-fit / automatic placement

I also think the auto-fit logic should be split into a separate PR.

Auto-fit is a policy layer built on top of the lower-level mechanisms, while layer split, runner scheduling, and split buffer support are mechanism-level changes. Keeping auto-fit separate would make the core multi-device implementation easier to review and test first, and would also make it easier to evaluate the fitting heuristics independently later.

For this PR, I would prefer not to introduce more auto-fit-specific options such as `--fit-compute-reserve` / `--fit-target`. Instead, it should reuse a more general mechanism such as `--max-vram`, so the user can specify the available memory budget and the auto-fit logic can derive a placement plan from that.

Overall, I think the core ideas are useful, but I would like to reduce the scope first. A smaller first PR for manual layer split / scheduler integration, followed by split-buffer support, and finally a separate auto-fit PR, would make this much easier to review, test, and merge safely.

pwilkin:
Yup, feels good. Will divide it into separate PRs and submit separately, with lazy loading I should be able to get a low enough quant to test everything on my machine :)

pwilkin:
@leejet aight, lazy me has taken his sweet time with the split, but I finally whipped Fable up to do the job and the split PRs have landed :)

leejet:
Superseded by #1734 #1735 #1736.

PR #1474, opened 2026-05-05
refactor: unify Euler, Euler Ancestral and DDIM implementations

wbruna:
This started as an attempt to simplify the DDIM sampler, and ended up removing it entirely 🙂 It turns out it is equivalent to Euler Ancestral. I've kept the algebraic demonstrations on each commit message.

I've also joined the original Euler Ancestral with the flow variant, with the helper function from #1436 . The same approach could probably be used for other ancestral implementations. 

The Euler merge is a little less clear-cut, since it loses its original simplicity, but I believe the unified code path to be overall simpler to follow and maintain.

### DIFF
--- a/src/denoiser.hpp
+++ b/src/denoiser.hpp
@@ -1631,28 +1631,24 @@ static sd::Tensor<float> sample_ddim_trailing(denoise_cb_t model,
         float sigma    = sigmas[i];
         float sigma_to = sigmas[i + 1];
 
-        auto model_output_opt = model(x, sigma, i + 1, nullptr);
-        if (model_output_opt.empty()) {
+        auto denoised_opt = model(x, sigma, i + 1, nullptr);
+        if (denoised_opt.empty()) {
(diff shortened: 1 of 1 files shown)


src/denoiser.hpp
```diff
@@ -823,67 +823,38 @@ static std::tuple<float, float, float> get_ancestral_step(float sigma_from,
...
+                    x *= alpha_scale;
+                }
+                x += sd::Tensor<float>::randn_like(x, rng) * sigma_up;
+            }
         }
     }
     return x;
 }
 
-static sd::Tensor<float> sample_euler(denoise_cb_t model,
```
leejet, line 871: I think `sample_euler` should be kept. It is the simplest and most elegant implementation, making it easier for beginners to understand without being distracted by additional parameters.

leejet (changes requested):

leejet:

PR #1477, opened 2026-05-06
feat: cross-stage offload modes and layer-streaming for low-VRAM GPUs

fszontagh:
# feat: cross-stage offload modes and layer-streaming for low-VRAM GPUs

## Why

Two problems that come up on small GPUs running large diffusion models:

1. **Cross-stage component placement.** Where does the text encoder live while diffusion runs? Where does diffusion go while the VAE decodes? On a 12 GB card running an 11.5 GB diffusion model, we need to move components in and out between stages or VAE decode hits OOM.
2. **Models that don't fit at all.** When the diffusion weights themselves exceed VRAM, we need to stream them in per-layer rather than load all at once.

This PR adds a single new flag, `--offload-mode`, that handles cross-stage placement, plus a per-layer streaming path (`--offload-mode layer_streaming`) for the doesn't-fit-at-all case.

## New CLI flags

| Flag | Description |
|---|---|
| `--offload-mode <mode>` | One of `none`, `cond_only`, `cond_diffusion`, `aggressive`, `layer_streaming`. Default `none`. |
| `--offload-cond-stage` / `--no-offload-cond-stage` | Override the cond-stage offload decision. |
| `--offload-diffusion` / `--no-offload-diffusion` | Override the diffusion-model offload decision. |
| `--offload-log` / `--no-offload-log` | Log offload events to stderr. |
| `--vram-estimation <method>` | `dryrun` (probe graph) or `formula` (analytic). |
| `--streaming-prefetch <N>` | Layers to prefetch ahead during streaming. Default 1. |
| `--streaming-min-vram <MB>` | Minimum free VRAM kept during streaming. Default 512. |

## What each mode does

| Mode | What it does | Use case |
|---|---|---|
| `none` (default) | No offload. Identical to current master behaviour. | Default; everything fits on GPU. |
| `cond_only` | Move text encoder to CPU after conditioning, keep diffusion on GPU. | Tight VRAM during diffusion. |
| `cond_diffusion` | Move both text encoder and diffusion model out between stages, swap them in for their stage. | VAE decode needs room; diffusion is too big to coexist with VAE compute buffer. |
| `aggressive` | Evict every component as soon as it's not actively used; reload on demand. | Lowest VRAM footprint at any moment; pays reload costs each transition. |
| `layer_streaming` | Diffusion weights live in pinned host RAM; each transformer block uploads to GPU just before it runs and is evicted afterwards. Async prefetch keeps PCIe full. | Models that don't fit at all (Z-Image bf16 11.5 GB on 12 GB card). |

## How layer streaming works

Three pieces, each a known-but-effective optimization at a different layer of the stack:

1. **Pinned host buffer** for streamed weights, so `cudaMemcpyAsync` actually goes async (a pageable source falls through to a synchronous bounce-buffer copy in the driver).
2. **Per-layer prefetch** overlapped with the previous layer's compute - the next layer's H2D starts on a separate stream while the current kernel is still running.
3. **Chunk graph for the resident block** - layers that fit on GPU stay there across sampling steps and run as one combined ggml graph dispatch instead of one mini-graph per layer.

A unified VRAM heuristic decides automatically which layers stay resident and which stream, based on actual free VRAM. Users don't have to pick a budget manually.

## Benchmarks - RTX 3060 (12 GB), PCIe 3.0 x16

Hardware: RTX 3060 12 GB. The card itself supports PCIe 4.0, but the board is DDR3-era so the slot is capped at PCIe 3.0 x16 (8.0 GT/s). PCIe bandwidth is the dominant cost during streaming, so faster boards (PCIe 4.0 x16, ~24 GB/s practical) should reduce these numbers materially.

All numbers below: **batch_count=4, steps=12, resolution=688x1024, LoRA applied at runtime**, same prompt/seed across configs.

### Z-Image-Turbo bf16 (11.5 GB diffusion model — does NOT fit in 12 GB)

Workload: 4 images per generation, 12 sampling steps each, batch=4. This is where streaming matters most — without offload of some kind, the model can't even load.

| Config | generate_image | Notes |
|---|---|---|
| `--offload-mode layer_streaming` | **175 s** | This PR. GPU utilization steady >90%; effective PCIe TX ~3.5 GB/s during streaming windows. |
| `--offload-to-cpu --max-vram 9` | 335 s | Existing graph-cut path. ~2× slower. |

### Z-Image-Turbo Q8 (6.7 GB diffusion model — fits in VRAM, but VAE compute buffer doesn't)

Workload: 4 images per generation, 12 sampling steps each, batch=4. When the model fits, streaming gives up most of its advantage and the simpler existing offload paths are slightly faster. Listed for completeness.

| Config | generate_image | Notes |
|---|---|---|
| `--offload-to-cpu` | **115 s** | Fastest when model fits. |
| `--vae-tiling` | 118 s | Tile VAE compute on GPU. |
| `--offload-mode layer_streaming` | 122 s | Auto-picks coarse-stage; still goes through streaming bookkeeping (~6% overhead). |
| `--offload-to-cpu --max-vram 6` | 152 s | Graph-cut adds dispatch overhead even when params fit. |
| `--vae-on-cpu` | 602 s | Reference; VAE on CPU is brutal. |

So the recommendation in the docs is: pick `--offload-mode layer_streaming` when the model doesn't fit (where it's ~2× faster than alternatives), and stick with the existing `--offload-to-cpu` (or no offload) when it does. `--offload-mode none` (default) keeps current master behaviour.

## Architectures

The streaming runtime is shared via `tensor_registry.hpp`, `layer_streaming.hpp`, `memory_budget.hpp`. Verified end-to-end on RTX 3060:

- Z-Image / Z-Image-Turbo (bf16 + Q8) - primary target
- Flux schnell
- Anima
- Qwen Image

Implemented and built but **not personally verified** by me - appreciate someone with the hardware/models confirming:

- MMDiT / SD3
- UNet (SD1.x / SDXL)
- WAN

## Known issues

- `--lora-apply-mode immediately` + `--offload-mode layer_streaming` crashes - the immediate folder reaches into weight buffers that haven't been uploaded to GPU yet under streaming. Use `at_runtime` (default `auto` already picks this in streaming mode). Pre-existing class of issue surfaced by streaming.
- VRAM estimation isn't perfect; `dryrun` is more accurate but adds a small startup cost. Switch to `dryrun` if you hit OOM during the first step.

## Backwards compatibility

Default behaviour is unchanged. `--offload-mode none` matches current master byte-for-byte. All new flags are opt-in.

## Bug fixes folded in

While exercising the offload paths I found and fixed a small set of pre-existing bugs. They're independent of the new offload modes and benefit users who never set `--offload-mode`. Happy to split these into a separate small PR if preferred.

- **`GGMLRunner` destructor leaked `runtime_params_buffer` and `partial_runtime_params_buffer`.** `free_params_buffer()` only released the CPU-side `params_buffer`. When the runner had been staged onto the runtime backend (any offload mode active, including the segmented offload from #1476), the GPU-side weight buffer(s) leaked on destruction. Real leak under LoRA + offload — many short-lived runners are created during LoRA application. Two-line additions to the destructor.
- CFG causing redundant model reloads under streaming.
- t_emb buffer aliasing in Z-Image's per-layer path.
- GGMLRunner scratch-buffer reuse.
- VAE-encode OOM in `aggressive` mode.
- Includes #1469's empty-`MultiLoraAdapter` fix (already merged into master); will rebase to drop that commit at PR time.

## Documentation

`docs/vram_offloading.md` covers the modes, decision tree, and example commands.


candrews:
This effort is really exciting!

Is there a way to automatically select the appropriate configuration so users don't need to manually do so? It would be nice if these performance improvements Just Worked :tm: (similar to how llama.cpp's `--fit` solves this tuning problem) instead of requiring extensive testing by each user to determine the best settings for each model (which inevitably results in cargo cult behaviors, confusion, and misunderstanding).

fszontagh:
> This effort is really exciting!
> 
> Is there a way to automatically select the appropriate configuration so users don't need to manually do so? It would be nice if these performance improvements Just Worked ™️ (similar to how llama.cpp's `--fit` solves this tuning problem) instead of requiring extensive testing by each user to determine the best settings for each model (which inevitably results in cargo cult behaviors, confusion, and misunderstanding).

Some pieces are already auto: e.g. our `layer_streaming` planner probes free VRAM at start-of-step and picks coarse-stage vs per-layer streaming on its own; the `--max-vram` graph-cut from #1476 segments based on the cap. What's still manual is the *mode* itself plus per-stage knobs (`--offload-to-cpu`, `keep_clip_on_cpu`, etc.).

A `--auto` (or default behaviour) that probes free VRAM + model size at load time and picks the most aggressive mode that fits seems straightforward to layer on top of what's here.

Roughly:
- model + cond + VAE all fit in headroom → `none`
- model fits but cond+model don't → `cond_only` / `cond_diffusion`
- model doesn't fit → `layer_streaming` (or graph-cut with auto `max_vram`)

leejet:
Nice job! This looks like a solid direction overall. I have a few review comments, mostly around keeping the execution path unified.

Conceptually, layer streaming is still segment execution: mark boundaries, split the graph, load the segment's params, execute, then evict or keep them resident. If we build pinned host buffers, next-segment/layer prefetch, and resident-block caching on top of the existing graph split path, we can keep one generic execution framework and avoid carrying complex per-model manual split code for Flux, Z-Image, Qwen, Wan, UNet, etc. Ideally the actual split/execution machinery should be shared with the current graph split implementation.

For cross-stage component placement, I may be missing some details, but it looks like this overlaps with the newer `--backend` / `--params-backend` options. It would be good to clarify what cases `--offload-mode cond_only/cond_diffusion/aggressive` cover that cannot be represented by those backend placement options, and whether the two systems can be unified.

AndriiParf:
Good evening (or good day), thanks for this awesome PR!! I tried it on my own system (Vulkan, AMD RX 580 8GB, Arch Linux, flux-2-klein-9b-Q8_0.gguf) and got a segmentation fault (core dumped). Using the AI (I don't know much about ML or C++), the AI suggested the following:
### Cause of the bug:
For **Flux.2 and Flux.2 Klein** models, `share_modulation` is set to `true`. This means the individual transformer blocks (`DoubleStreamBlock` and `SingleStreamBlock`) do NOT instantiate their own local modulation blocks (`img_mod`, `txt_mod`, `modulation`). Instead, they share global modulation blocks initialized in the parent `Flux` class.

In the original non-streaming path, these global modulations are precalculated and passed to `block->forward(...)`.
However, in the streaming path (`compute_streaming_true`), `ds_img_mods`, `ds_txt_mods`, and `ss_mods` are passed as empty vectors. This forces `DoubleStreamBlock::forward` and `SingleStreamBlock::forward` to fall back to their local `img_mod` objects, which are `nullptr` under shared modulation, triggering an instant null-pointer dereference (Segfault).

### Suggested Fix:
To resolve this, we can make `forward_double_block` and `forward_single_block` inside `struct Flux` automatically compute these shared modulations on-the-fly when `share_modulation` is active and the incoming vectors are empty.

In **`src/flux.hpp`**, modify `forward_double_block`:

```cpp
        std::pair<ggml_tensor*, ggml_tensor*> forward_double_block(GGMLRunnerContext* ctx,
                                                                    int block_idx,
                                                                    ggml_tensor* img,
                                                                    ggml_tensor* txt,
                                                                    ggml_tensor* vec,
                                                                    ggml_tensor* pe,
                                                                    ggml_tensor* txt_img_mask,
                                                                    std::vector<ModulationOut>& ds_img_mods,
                                                                    std::vector<ModulationOut>& ds_txt_mods) {
            if (params.share_modulation && ds_img_mods.empty()) {
                auto double_stream_modulation_img = std::dynamic_pointer_cast<Modulation>(blocks["double_stream_modulation_img"]);
                auto double_stream_modulation_txt = std::dynamic_pointer_cast<Modulation>(blocks["double_stream_modulation_txt"]);
                ds_img_mods = double_stream_modulation_img->forward(ctx, vec);
                ds_txt_mods = double_stream_modulation_txt->forward(ctx, vec);
            }
            auto block = std::dynamic_pointer_cast<DoubleStreamBlock>(blocks["double_blocks." + std::to_string(block_idx)]);
            auto img_txt = block->forward(ctx, img, txt, vec, pe, txt_img_mask, ds_img_mods, ds_txt_mods);
            return img_txt;
        }
```

And modify `forward_single_block`:

```cpp
        ggml_tensor* forward_single_block(GGMLRunnerContext* ctx,
                                           int block_idx,
                                           ggml_tensor* txt_img,
                                           ggml_tensor* vec,
                                           ggml_tensor* pe,
                                           ggml_tensor* txt_img_mask,
                                           std::vector<ModulationOut>& ss_mods) {
            if (params.share_modulation && ss_mods.empty()) {
                auto single_stream_modulation = std::dynamic_pointer_cast<Modulation>(blocks["single_stream_modulation"]);
                ss_mods = single_stream_modulation->forward(ctx, vec);
            }
            auto block = std::dynamic_pointer_cast<SingleStreamBlock>(blocks["single_blocks." + std::to_string(block_idx)]);
            return block->forward(ctx, txt_img, vec, pe, txt_img_mask, ss_mods);
        }
```
After following his advice, everything went smoothly and very quickly. Sorry for using AI

nArn0:
I was just hit by the same problem while testing Wan2.2 with the exact same card on the exact same board (PCIe 3). 

I can make WAN2.2 work with WAN2GP but i would really love to make it work with sd.ccp instead in order to use GGUF version of the model.

Pix13:
Do you think you could make this feature work directly from the disk instead of RAM <-> VRAM ? I know it's a niche but there is a powerful HW that has 16Gb shared ram/vram ( a mini stryx halo basically) named the bc-250. I already vibe-coded an offload-to-disk feature that works ( could make work qwen-image-edit-2511-Q4_0.gguf - 12Gb on the board for example). Of course the nvme is the caveat there but if your architecture permits it without too much hassle.. it would be incredible :) ( i guess i'll maintain my branch otherwise, rebasing on your work)

GreenShadows:
That would quickly destroy the SSD, unless you have something like Optane.

Pix13:
> That would quickly destroy the SSD, unless you have something like Optane.

How would reading destroy the ssd ? Unless i'm mistaken, this streams part of the model to the vram, no data generated goes "back" so it would only be read operations ?

(i can live with killing nvme, if it's not the case, they're consumables)

fszontagh:
> Do you think you could make this feature work directly from the disk instead of RAM <-> VRAM ? I know it's a niche but there is a powerful HW that has 16Gb shared ram/vram ( a mini stryx halo basically) named the bc-250. I already vibe-coded an offload-to-disk feature that works ( could make work qwen-image-edit-2511-Q4_0.gguf - 12Gb on the board for example). Of course the nvme is the caveat there but if your architecture permits it without too much hassle.. it would be incredible :) ( i guess i'll maintain my branch otherwise, rebasing on your work)

im already working on similar thing. This feature is coming from the #1414. 
this can be enabled / disabled with the parameters in the cli  `--mmap` / `--no-mmap`. By default this is enabled. 
When `--mmap` is enabled and you select the layer streaming, its can slow down the entire process if you use HDD. I only tested on HDD (worst case). In theory in tmpfs / memfs this can be speed up if you have enough memory for it. 

In my [REST API project](https://github.com/fszontagh/stable-diffusion.cpp-restapi) it can be disabled / enabled at model load time.
`POST /models/load │ options.enable_mmap: true/false. Default: true.`

In theory this is not killing the SSD its just normal reading.

examples/common/common.cpp
```diff
@@ -554,6 +554,78 @@ ArgOptions SDContextParams::get_options() {
         return 1;
     };
 
+    auto on_offload_mode_arg = [&](int argc, const char** argv, int index) {
```
leejet, line 557: I don't think this feature should introduce a new offload execution mode or expose this many new CLI flags.

We already have the necessary user-facing control surface via:

* backend
* params-backend
* max-vram

The runtime should derive the streaming/offload policy automatically from those constraints.

Most of the newly added flags expose internal scheduling policy rather than user intent, which significantly increases API complexity and creates another execution path that can diverge from existing graph-cut execution semantics.

I would strongly recommend redesigning this as a scheduling/prefetch policy layered on top of the existing graph-cut execution plan instead of introducing a parallel “layer streaming” execution mode.

src/flux.hpp
```diff
@@ -846,6 +848,162 @@ namespace Flux {
             }
         }
 
+        struct StreamingInputResult {
```
leejet, line 851: This is where I think the design diverges from the existing execution model.

The graph has already been segmented by graph cut marks and the max-vram planner. Rebuilding model-specific “layer subgraphs” here duplicates graph partitioning logic and introduces a second segmentation mechanism that will likely drift over time.

Layer streaming should operate on top of existing graph-cut segments and only add:

* prefetch scheduling
* residency decisions
* eviction/reload policy

It should not manually reinterpret model structure and create a parallel partitioning path.

leejet (changes requested):
The overall direction makes sense, but I don't think this should merge in its current architecture.

The implementation currently introduces:

* a parallel layer-streaming execution path,
* additional graph partitioning logic,
* and many new user-facing policy flags.

I believe this should instead build on top of the existing graph-cut/max-vram execution model and implement streaming purely as a residency/prefetch policy over existing graph segments.

Otherwise we risk maintaining two separate execution/partitioning systems that may diverge over time.

AndriiParf:
I tried wan 2.2 on my PC (RX 580 8GB, 32GB DDR4, Arch Linux, q4_k_m low and high noise models) and initially got a black screen, but after following the AI's recommendations, the generation process started and everything worked fine. AI report: 
### Bug 1: `high_noise_diffusion_model` bypasses layer streaming config
When running a dual-model setup (using both `--diffusion-model` and `--high-noise-diffusion-model`), layer-by-layer streaming is only enabled for the primary `diffusion_model` [3]. The `high_noise` model falls back to the default non-streaming path [3]. This causes an immediate OOM/GPU hang as soon as the high-noise sampling phase begins.

#### Fix 1:
In `src/stable-diffusion.cpp` (around line 725), enable streaming for the high-noise model right after its parameters are bound:
```cpp
            if (high_noise_diffusion_model) {
                high_noise_diffusion_model->set_max_graph_vram_bytes(max_graph_vram_bytes);
                get_param_tensors(high_noise_diffusion_model, module_can_mmap(SDBackendModule::DIFFUSION));
                
                // Enable streaming for high_noise_diffusion_model
                if (offload_config.mode == SD_OFFLOAD_LAYER_STREAMING) {
                    if (high_noise_diffusion_model->supports_layer_streaming()) {
                        LOG_INFO("Enabling layer-by-layer streaming for high_noise_diffusion_model");
                        high_noise_diffusion_model->enable_layer_streaming(
                            offload_config.streaming_prefetch_layers,
                            offload_config.streaming_min_free_vram);
                    }
                }
            }
```

Also, update the initial CPU-offloading block (around line 1058) to offload the high-noise layers immediately too:
```cpp
        if (offload_config.mode == SD_OFFLOAD_LAYER_STREAMING) {
            if (diffusion_model && diffusion_model->is_layer_streaming_enabled()) {
                LOG_INFO("Offloading diffusion model layers to CPU for layer streaming");
                diffusion_model->offload_streaming_layers();
            }
            if (high_noise_diffusion_model && high_noise_diffusion_model->is_layer_streaming_enabled()) {
                LOG_INFO("Offloading high_noise_diffusion_model layers to CPU for layer streaming");
                high_noise_diffusion_model->offload_streaming_layers();
            }
        }
```

---

### Bug 2: Out-of-bounds vector access inside `WanRunner`
This is what caused the immediate segmentation fault on systems with strict bounds checking.

When running `Wan2.2-I2V-A14B` (which doesn't have VACE blocks), `stage1_c_out_` is `nullptr` and `c_n = 0`.
Therefore, the `sizes` vector passed to `ensure_pinned_act_buffers` has only **5 elements**, and the returned `ptrs` vector also has a size of **5** (valid indices are 0 to 4).

However, the line:
```cpp
persistent_c_ = (c_n > 0) ? ptrs[5] : nullptr;
```
still evaluates the `ptrs[5]` index expression. On toolchains with strict assertions (like Arch Linux's default GCC package which has `-D_GLIBCXX_ASSERTIONS` enabled by default), accessing index `5` of a size 5 vector triggers an immediate abort, resulting in a segmentation fault.

#### Fix 2:
In `src/wan.hpp` (around line 765), change:
```cpp
persistent_c_ = (c_n > 0) ? ptrs[5] : nullptr;
```
to:
```cpp
persistent_c_ = (c_n > 0 && ptrs.size() > 5) ? ptrs[5] : nullptr;
```

---

### Tip for other low-VRAM/RAM testers:
Standard VAE decoding for 33 frames of 832x480 WAN video requests a massive `15.5 GB` RAM compute buffer (`wan_vae compute buffer size: 15501.51 MB(RAM)`), which easily triggers the Linux OOM killer on 32GB RAM systems. 
Replacing the heavy VAE with TAEHV (`--tae taew2_1.safetensors`) brings the VAE compute buffer down to < 0.5 GB, allowing the entire pipeline to fit comfortably within RAM while streaming on an 8GB GPU.

fszontagh:
Superseded by #1576, which rebuilds this on top of the graph-cut planner (#1476) with one new flag (`--stream-layers`) instead of the parallel system here. Closing.

Issue #1475, opened 2026-05-06
[Bug] Build failure on Fedora 44 (GCC 16)


cphlipot:
### Git commit

$ git rev-parse HEAD
3d6064b37ef4607917f8acf2ca8c8906d5087413

### Operating System & Version

Fedora 44

### GGML backends

Vulkan

### Command-line arguments used

N/A (project did not build)

### Steps to reproduce

attempt build on Fedora 44 using instructions in build.md

```
mkdir build && cd build
cmake .. -DSD_VULKAN=ON
cmake --build . --config Release
```

### What you expected to happen

project gets built successfully.

### What actually happened

build failed

### Logs / error messages / stack trace

```
error: no match for 'operator=' (operand types are 'std::string' and 'std::u8string')
  267 | lora_entry.name = p.stem().u8string();
```

### Additional context / environment details

Fedora 44 Ships with GCC 16. GCC 16 will compile projects with C++20 by default unless the project overides the behavior which stable-diffusion.cpp does not.

https://gcc.gnu.org/gcc-16/changes.html#cxx

> C++20 by default: GCC 16 changes the default language version for C++ compilation from [-std=gnu++17](https://gcc.gnu.org/onlinedocs/gcc-16.1.0/gcc/C-Dialect-Options.html#index-std-1) to [-std=gnu++20](https://gcc.gnu.org/onlinedocs/gcc-16.1.0/gcc/C-Dialect-Options.html#index-std-1).

in C++20 it does not allow direct assignment from std::u8string to std::string, which stable diffusion cpp currently does and results in compiler errors.

potential fixes include:
1. use std::string instead of u8string
2. convert project to use u8string instead of std::string
3. compile with --std=c++17 to get old compiler behavior with the newer GCC so we don't need to change the source code.

see also https://stackoverflow.com/questions/56833000/c20-with-u8-char8-t-and-stdstring

wbruna:
> unless the project overides the behavior which stable-diffusion.cpp does not.

Actually, we do:

https://github.com/leejet/stable-diffusion.cpp/blob/master-593-3d6064b/CMakeLists.txt#L256
https://github.com/leejet/stable-diffusion.cpp/blob/master-593-3d6064b/examples/cli/CMakeLists.txt#L20
https://github.com/leejet/stable-diffusion.cpp/blob/master-593-3d6064b/examples/server/CMakeLists.txt#L104

```cmake
[...]

PR #1478, opened 2026-05-06
fix: Fix broken GCC 16 build (enforce C11/C++17 compile )

cphlipot:
## Summary

Currently the project requests the C11 and C++17 feature sets using
target_compile_features, however this only gaurentees us a minimum
C++ version of C++17, and allows compile to proceed with C++20
(the GCC 16 default) if it is already set.

Building Stable-diffusion.cpp with C++20 results in build failures
as std::u8string can no longer be assigned to std::string due to
C++20 standards changes.

Explicitly set the required language standard to C++17 in order to
ensure we pass/fail the build consistently across compiler toolchains
even if compilers by default try to expose a superset of C++17
features.

This fix copies the approach used by https://github.com/ggml-org/ggml/blob/master/CMakeLists.txt#L284-L288

A side effect of this change is that is also now enforces the CONTRIBUTING.md's requirement
to use C++17 at the CMake level.

## Related Issue / Discussion

#1475

## Additional Information

Tested build now succeeds on Fedora 44 with GCC 16 using the
following commands for vulkan build from build.md

Verify GCC 16+ is being used:
```
gcc --version

gcc (GCC) 16.1.1 20260501 (Red Hat 16.1.1-1)
```

then verify build:
```
mkdir build && cd build
cmake .. -DSD_VULKAN=ON
```

## Checklist


candrews:
Tested this patch and it fixed the compilation problem for me.

```
$ gcc --version
gcc (Gentoo 16.1.0 p1) 16.1.0
```

Nindaleth:
Thanks for unblocking the GCC 16 build!

cphlipot:
Updated the PR. Now the CmakeLists.txt should enforce C11/C++17 at the project level instead of target level, aligning itself with how other projects like GGML do this. this also simplified the change a bit.

Also i noticed @leejet just added a CONTRIBUTING.md and PR template, so updated the PR to comply with those.

leejet:

Issue #1447, opened 2026-04-20
[Bug] ERNIE - white blank (diffusion-fa and gen-size affected)


SolicTous:
### Git commit

https://github.com/leejet/stable-diffusion.cpp/commit/44cca3d626d301e2215d5e243277e8f0e65bfa78

### Operating System & Version

Win 10

### GGML backends

CUDA

### Command-line arguments used

chcp 65001 "../sd-cli.exe" ^  --diffusion-model ../../Models/Diffusers/ERNIE-Image/ernie-image-turbo-UD-Q4_K_M.gguf ^  --vae ../../Models/Diffusers/Vae/vae_ernie_image_turbo.safetensors ^  --llm ../../Models/Diffusers/Ministral/ministral-3-3b.safetensors ^  -p "a lovely cat holding a sign says 'ernie.cpp'" ^  --cfg-scale 1.0 ^  --steps 8 ^  --diffusion-fa ^  -H 1024 ^  -W 1024 ^  -v ^  -o ../output/cli_out.jpg pause

### Steps to reproduce

Run any model of Ernie-Image

### What you expected to happen

Images as on example

### What actually happened

Getting white blank

<img width="256" height="256" alt="Image" src="https://github.com/user-attachments/assets/914c27a8-408f-4834-af55-8cd8b17a1f99" />

### Logs / error messages / stack trace

"../sd-cli.exe"  --diffusion-model ../../Models/Diffusers/ERNIE-Image/ernie-image-turbo-Q8_0.gguf  --vae ../../Models/Diffusers/Vae/vae_ernie_image_turbo.safetensors  --llm ../../Models/Diffusers/Ministral/Ministral-3-3B-Instruct-2512-UD-Q8_K_XL.gguf  -p "a lovely cat holding a sign says 'ernie.cpp'"  --cfg-scale 1.0  --steps 8  --diffusion-fa  -H 1024  -W 1024  -v  -o ../output/cli_out.jpg
[DEBUG] main.cpp:547  - version: stable-diffusion.cpp version unknown, commit 44cca3d
[DEBUG] main.cpp:548  - System Info:
[...]

PR #1481, opened 2026-05-08
fix: Update ernie_image.md (avoid white screen)

SolicTous:
A Ernie Image's way to avoid the white screen issue https://github.com/leejet/stable-diffusion.cpp/issues/1447


SolicTous:
looks like last commits now fixed diffusion-fa

PR #1484, opened 2026-05-10
feat: add Gradient Estimation sampler

wbruna:
Implements the Gradient Estimation sampler, from https://github.com/ToyotaResearchInstitute/gradient-estimation-sampler :

> sd-cli --diffusion-model ./z_image_turbo-Q8_0.gguf --llm ./Qwen3-4B-UD-Q4_K_XL.gguf --vae ./ae_bf16.safetensors --cfg-scale 1 --steps 8 --sampling-method $sampler --scheduler simple --eta $eta -p 'a pretty flower' -W 768 -H 768

| eta | Euler | Euler GE |
|:---:|:---:|:---:|
| 0 | test euler_a 0 | test euler_ge 0 |
| 1 | test euler_a 1 | test euler_ge 1 |

I've called it 'euler_ge' because it's less of a mouthful than 'gradient_estimation', but arguably the latter would be more 'standard'?

For now, `gamma` is fixed at 2. We should probably implement a generic key=value option for this kind of sampling parameter. 

The noise injection could be wrong: it works well with Simple and SGM Uniform, but other samplers tend to make noisy images.


leejet:
You can use --extra-sample-args to pass additional parameters.

wbruna:
Added support for setting the `gamma` parameter; lower values seem to help with other schedulers:

> sd-cli --diffusion-model ./z_image_turbo-Q8_0.gguf --llm ./Qwen3-4B-UD-Q4_K_XL.gguf --vae ./ae_bf16.safetensors --cfg-scale 1 --steps 8 --sampling-method euler_ge --scheduler smoothstep --eta 0.5 --extra-sample-args gamma=1.5 -p 'a pretty flower' -W 768 -H 768

smoothstep_gamma1 5_eta0 5

PR #1485, opened 2026-05-10
feat: add hidream o1 image support

leejet:
```
.\bin\Release\sd-cli.exe -m  ..\models\hidream_o1_image_dev_bf16.safetensors -p "a lovely cat holding a sign says 
'hidream o1 cpp'" --cfg-scale 1.0 -v -H 1024 -W 1024
```

output


src/denoiser.hpp
```diff
@@ -867,6 +867,31 @@ static sd::Tensor<float> sample_euler_flow(denoise_cb_t model,
...
+        if (denoised_opt.empty()) {
+            return {};
+        }
+        sd::Tensor<float> denoised = std::move(denoised_opt);
+        if (sigma_next == 0.0f) {
+            x = std::move(denoised);
+            continue;
+        }
+        auto noise = sd::Tensor<float>::randn_like(x, rng);
+        x          = sigma_next * noise * s_noise + (1.0f - sigma_next) * denoised;
```
wbruna: Interesting. This is basically the same as the LCM sampler, but using eta as an additional noise multiplier. Maybe the two implementations could be merged?

leejet: Yes, it has currently been merged into a single LCM sample method.

Issue #1425, opened 2026-04-15
[Bug] WAN 2.2 tensor naming weirdness.


lastrosade:
### Git commit

6b675a5ede9b0edf0a0f44191e8b79d7ef27615a

### Operating System & Version

Windows 11

### GGML backends

HIP

### Command-line arguments used

```
$models = "a:\AI_MODELS\Stable Diffusion\Models"
& 'a:\AI_MODELS\Stable Diffusion\stable-diffusion.cpp.hip.bin\sd-cli.exe' -M vid_gen `
--diffusion-model "$($models)\checkpoints\wan\Remix_v3_low.q5_K.gguf" `
--high-noise-diffusion-model "$($models)\checkpoints\wan\Remix_v3_high.q5_K.gguf" `
--t5xxl "$($models)\text_encoders\NSFW_WAN_umt5-xxl-encoder.q8_0.gguf" `
--tae "$($models)\tae\taew2_1.safetensors" `
--lora-model-dir "$($models)\LoRa\wan" `
-n "" --seed 0 --offload-to-cpu --clip-on-cpu --fa --diffusion-fa --vae-tiling `
--sampling-method euler --cfg-scale 1 --steps 4 --scheduler simple `
--high-noise-sampling-method euler --high-noise-cfg-scale 1 `
--flow-shift 3.0 -W 432 -H 256 --fps 15 --video-frames 61 `
-i "a:\AI_MODELS\Stable Diffusion\stable-diffusion.cpp\assets\with_lcm.png" `
-p "a lovely cat" --verbose
```

### Issue

Using the WAN 2.2 Remix Models

There seems to be an issue with how the tensor names are parsed.
While the low noise model loads correctly, the high noise model does not.

```
[INFO ] model.cpp:1614 - unknown tensor 'model.high_noise_diffusion_model.model.diffusion_model.blocks.0.cross_attn.norm_q.weight | q5_K | 1 [5120, 1, 1, 1, 1]' in model file
[ERROR] model.cpp:1655 - tensor 'model.high_noise_diffusion_model.blocks.0.cross_attn.norm_q.weight' not in model file
```
I will spare you the full list.

<img width="973" height="82" alt="Image" src="https://github.com/user-attachments/assets/b1859222-6b92-45e1-b7ba-e12490cb2f67" />

[fulllog.txt](https://github.com/user-attachments/files/26756399/fulllog.txt)

lastrosade:
[...]

Issue #1064, opened 2025-12-08
[Bug] cannot open z-image.


bendyfan11:
### Git commit

n/a, downloaded vulkan build

### Operating System & Version

Windows 10

### GGML backends

Vulkan

### Command-line arguments used

sd.exe -m D:\models\img\z_img_turbo-Q4_k.gguf --vae D:\models\img\vae\ae.safetensors --llm D:\models\img\Qwen3-4B-Instruct-2507-Q4_K_M.gguf --diffusion-fa

### Steps to reproduce

attempt to load Z image

### What you expected to happen

the model loads

### What actually happened

model does not load.

### Logs / error messages / stack trace

[INFO ] stable-diffusion.cpp:228  - loading model from 'D:\models\img\z_img_turbo-Q4_k.gguf'
[WARN ] model.cpp:379  - unknown format D:\models\img\z_img_turbo-Q4_k.gguf
[ERROR] stable-diffusion.cpp:230  - init model loader from file failed: 'D:\models\img\z_img_turbo-Q4_k.gguf'
[INFO ] stable-diffusion.cpp:282  - loading llm from 'D:\models\img\Qwen3-4B-Instruct-2507-Q4_K_M.gguf'
[INFO ] model.cpp:370  - load D:\models\img\Qwen3-4B-Instruct-2507-Q4_K_M.gguf using gguf format
[INFO ] stable-diffusion.cpp:296  - loading vae from 'D:\models\img\vae\ae.safetensors'
[INFO ] model.cpp:373  - load D:\models\img\vae\ae.safetensors using safetensors format
[ERROR] stable-diffusion.cpp:306  - get sd version from file failed: 'D:\models\img\z_img_turbo-Q4_k.gguf'
new_sd_ctx_t failed

wbruna:
You need to use the flag `--diffusion-model` instead of `-m`; see https://github.com/leejet/stable-diffusion.cpp/blob/master/docs/z_image.md for an example.

bendyfan11:
> You need to use the flag `--diffusion-model` instead of `-m`; see https://github.com/leejet/stable-diffusion.cpp/blob/master/docs/z_image.md for an example.

I actually ended up loading the model, but now it dosnt want to load the sampler. figuring that out now

akleine:
_Git commit_
**n/a, downloaded vulkan build**

Somewhat off topic:
that's the reason why I suggested providing the current commit number at start of sd, see #1005
So printing a commit number would help here.

Issue #887, opened 2025-10-12
[Enhancement]: Simplify model flags


MrDrMcCoy:
To improve the user experience, would it be feasible to combine the `--diffusion-model` and `--model` flags by having the application perform an initial examination of the model file to see what it needs and adapt to supplied flags or error with a useful message?

Issue #884, opened 2025-10-12
Chroma: get sd version from file failed: new_sd_ctx_t failed


MrDrMcCoy:
Hello,

I have been trying to get any version of [Chroma](https://github.com/leejet/stable-diffusion.cpp/blob/master/docs/chroma.md) to work with `stable-diffusion.cpp`, but to no avail. I only get the following error:

```
[ERROR] stable-diffusion.cpp:273  - get sd version from file failed: <FILE_PATH> new_sd_ctx_t failed
```

I tried with the following combinations of...

Model:
- `chroma-unlocked-v50.safetensors`
- `chroma-unlocked-v50-flash-heun.safetensors`
- `Chroma1-HD.safetensors`
- `Chroma1-HD-Flash.safetensors`
- `chroma-unlocked-v50-BF16.gguf`
- `chroma-unlocked-v47-flash-heun-BF16.gguf`
- `chroma-unlocked-v47-flash-heun-Q6_K.gguf`

I also attempted to use my own quantizations of the above down to `q4_1` in `gguf` format. 

VAEs:
- `ae.safetensors` from flux.1-dev
- `vae/diffusion_pytorch_model.safetensors` from flux.1-dev

clip_l:
- `clip_l.safetensors`

T5XXL:
- `t5xxl_fp16.safetensors`

App builds:
- `CPU-only`
- `ROCm`
- `Vulkan`

Platform: Linux

GPUs:
- `Radeon 7900XTX`
- `Radeon 7600M XT`
- `Radeon 880M`

I have access to an `Arc B580` for `SYCL`/`Vulkan` testing, but don't know that it will be worth the effort when CPU-only also fails.

Example command:

```shell
stable-diffusion --model flux.1/base/chroma-unlocked-v47-flash-heun-Q6_K.gguf --clip_l flux.1/clip/clip_l.safetensors --t5xxl flux.1/clip/t5xxl_fp16.safetensors --vae flux.1/vae/ae.safetensors --vae-on-cpu --clip-on-cpu --control-net-cpu --color --height 768 --width 512 --steps 8 --batch-count 1 --cfg-scale 1.0 --sampling-method euler --scheduler discrete --prompt "close-up photo of a woman in her early 30s" --output out/chroma-test
```

stduhpf:
I believe using `--diffusion-model` instead of `--model` is required to make it work. 

[...]

PR #1487, opened 2026-05-12
fix: detect QWEN_IMAGE version from bare GGUF tensor names

honey1000001001:
## Problem

Some GGUF files (e.g. from [unsloth](https://huggingface.co/unsloth/Qwen-Image-2512-GGUF)) store diffusion model tensors **without** the `model.diffusion_model.` prefix. For example, tensor names look like:

- `transformer_blocks.0.img_mod.1.weight`
- `img_in.bias`
- `proj_out.weight`

Instead of:
- `model.diffusion_model.transformer_blocks.0.img_mod.1.weight`

This causes `get_sd_version()` to fail with `get sd version from file failed`, even though the binary has full Qwen-Image support (the `--llm` flag, `QwenImageModel` class, etc. are all present).

Reference: [lemonade-sdk/lemonade#1874](https://github.com/lemonade-sdk/lemonade/issues/1874) — users hitting this when trying to run `Qwen-Image-2512-GGUF` through `sd-cpp`.

## Fix

Two changes:

### 1. `src/model.cpp` - Version detection

In `get_sd_version()`, match `transformer_blocks.0.img_mod.1.weight` with or without the `model.diffusion_model.` prefix.

### 2. `src/name_conversion.cpp` - Tensor name conversion

After `replace_with_prefix_map()`, add `model.diffusion_model.` prefix to bare Qwen-Image tensor names that don't already have a recognized prefix. This ensures `QwenImageModel` (which uses prefix `model.diffusion_model`) can find the tensors.

## Testing

Verified locally that `sd-server.exe` can now detect Qwen-Image GGUF version and starts loading the model (text encoder + VAE load successfully, main GGUF passes version detection).

### DIFF
--- a/src/model.cpp
+++ b/src/model.cpp
@@ -437,7 +437,8 @@ SDVersion ModelLoader::get_sd_version() {
         if (tensor_storage.name.find("model.diffusion_model.joint_blocks.") != std::string::npos) {
             return VERSION_SD3;
         }
-        if (tensor_storage.name.find("model.diffusion_model.transformer_blocks.0.img_mod.1.weight") != std::string::npos) {
+        if (tensor_storage.name.find("model.diffusion_model.transformer_blocks.0.img_mod.1.weight") != std::string::npos ||
+            tensor_storage.name.find("transformer_blocks.0.img_mod.1.weight") != std::string::npos) {
             return VERSION_QWEN_IMAGE;
         }
         if (tensor_storage.name.find("llm_adapter.blocks.0.cross_attn.q_proj.weight") != std::string::npos) {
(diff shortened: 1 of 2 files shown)


wbruna:
Are you loading it through the field `model_path` (flag `--model`) instead of `diffusion_model_path` (flag `--diffusion-model`)?

I'm probably not finding all relevant discussions right now, but this is related to _at least_ : #403 #434 #435 #884 #887 #1064 #1425

leejet:
For files that contain only diffusion weights, please use `--diffusion-model` instead of `--model`.

PR #1463, opened 2026-04-27
feat: add ltx2.3 support

leejet:
### LTX-2.3 dev T2V

```
.\bin\Release\sd-cli.exe -M vid_gen --diffusion-model  ..\..\ComfyUI\models\diffusion_models\ltx-2.3-22b-dev-UD-Q4_K_M.gguf --vae ..\..\ComfyUI\models\vae\ltx-2.3-22b-dev_video_vae.safetensors --audio-vae ..\..\ComfyUI\models\vae\ltx-2.3-22b-dev_audio_vae.safetensors --llm ..\..\ComfyUI\models\text_encoders\gemma-3-12b-it-qat-UD-Q4_K_XL.gguf --embeddings-connectors ..\..\ComfyUI\models\text_encoders\ltx-2.3-22b-dev_embeddings_connectors.safetensors  -p "a lovely cat" --cfg-scale 6.0 --sampling-method euler -v -n "worst quality, low quality, blurry, distorted, artifacts" -W 1280 -H 720 --diffusion-fa --offload-to-cpu --video-frames 33 --fps 24 -o t2v.webm
```
[t2v.webm](https://github.com/user-attachments/assets/a5212ad4-4338-4c60-82b9-05d4797035eb)

### LTX-2.3 dev I2V

```
.\bin\Release\sd-cli.exe -M vid_gen --diffusion-model  ..\..\ComfyUI\models\diffusion_models\ltx-2.3-22b-dev-UD-Q4_K_M.gguf --vae ..\..\ComfyUI\models\vae\ltx-2.3-22b-dev_video_vae.safetensors --audio-vae ..\..\ComfyUI\models\vae\ltx-2.3-22b-dev_audio_vae.safetensors --llm ..\..\ComfyUI\models\text_encoders\gemma-3-12b-it-qat-UD-Q4_K_XL.gguf --embeddings-connectors ..\..\ComfyUI\models\text_encoders\ltx-2.3-22b-dev_embeddings_connectors.safetensors  -p "a lovely cat" --cfg-scale 6.0 --sampling-method euler -v  -W 1280 -H 720 --diffusion-fa --offload-to-cpu --video-frames 33 -i ..\assets\ernie_image\turbo_example.png -o i2v.webm
```

[i2v.webm](https://github.com/user-attachments/assets/6efb1579-72ef-4e0c-8ea3-b18fdabcbc09)

PR #1489, opened 2026-05-12
ltx2.3: fix sd_audio stereo format

stduhpf:
targets https://github.com/leejet/stable-diffusion.cpp/pull/1463

https://github.com/leejet/stable-diffusion.cpp/pull/1463#issuecomment-4426081199

If changing the format of sd_audio's data is undesirable, we could alternatively interleave the audio samples just before saving to webm or avi, but it was easier for me to implement it like that.


leejet:
I didn’t notice that this PR was targeting the `ltx2.3` branch. I’ve cherry-picked it to the `master` branch now.

PR #1491, opened 2026-05-12
fix: missing ggml_backend_free on the upscaler

wbruna:
Included a small refactor to make it easier to manage the `ggml_backend_t` instances, and deduplicate the *-on-cpu backends.

Maybe it would be better to replace the backend raw pointers with `shared_ptr`s everywhere instead, but I didn't want a huge diff just for this fix.


wbruna:
Will likely be superseded by #1500 .

PR #1503, opened 2026-05-16
feat: make negative max_vram control the amount of spare vram

wbruna:
## Summary

Repurpose the `-1` sentinel value to mean "spare this amount of free VRAM".

1 GiB can be a lot in certain circumstances: for instance, a small card, or a dedicated one which doesn't need to keep that much free VRAM available. The user could choose a positive value specific for their case, but that would miss the ability to calculate a good value from the available VRAM.

I've kept the documentation and help texts as-is for now to avoid a big diff, but I'll adapt them if this feature is considered useful.

## Checklist


leejet:
Please update the document.

PR #1490, opened 2026-05-12
Temporal tile size + overlap

stduhpf:
Targets https://github.com/leejet/stable-diffusion.cpp/pull/1463

 This PR improves the temporal tiled decoding for the LTX2.3 Video VAE by adding support for custom tile sizes (processing multiple latent frames per batch) and temporal overlap/padding to ensure smoother transitions between "tiles" as the VAE decoder has access to the near future. 

For now it ses env variables `VAE_TILE_FRAMES` and `VAE_TILE_PAD` to control the effect. In my experience, setting `VAE_TILE_FRAMES=4` and `VAE_TILE_PAD=1` seems to give very decent results (see comments).

How it works: 
When using temporal tiling, instead of processing latent frames sequencially, each being able to access only its predecessors, the latent frames are batched together in chunks/tiles of size `VAE_TILE_FRAMES` to be decoded at once, still with access to the previous batches. The last `VAE_TILE_PAD` latent frames of each batch are discarded (and removed from the cache), and then re-processed as part of the next batch.

Quick comparison of compute buffer VRAM usage for 512x512x129 (17 latent frames) for various temporal tile sizes with 0 padding (padding seems to slightly decrease the compute buffer size, but might make decoding a bit longer)

| Tile Size (frames) | Compute Buffer Size (VRAM) |
| :--- | :--- |
| 1 | 3,271.25 MB |
| 2 | 5,525.38 MB |
| 4 | 10,365.62 MB |
| 8 | 19,744.12 MB |
| No Tiling | 19,118.13 MB |

PR #1510, opened 2026-05-17
Feat: Temporal tile custom size with overlap

stduhpf:
## Summary

This PR improves the temporal tiled decoding for the LTX2.3 Video VAE by adding support for custom tile sizes (processing multiple latent frames per batch) and temporal overlap/padding to ensure smoother transitions between "tiles" as the VAE decoder has access to the near future.

For now it ses env variables `VAE_TILE_FRAMES` and `VAE_TILE_PAD` to control the effect. In my experience, setting `VAE_TILE_FRAMES=4` and `VAE_TILE_PAD=1` seems to give very decent results (see [comment](https://github.com/leejet/stable-diffusion.cpp/pull/1490#issuecomment-4436357040)).

## Related Issue / Discussion

https://github.com/leejet/stable-diffusion.cpp/pull/1490

### How it works:

Currently, when temporal tiling is enabled, it switches from decoding all latent frames at once to decoding them sequentially one-by-one. However, this causes choppy transitions every 8 frames (corresponding to 1 latent frame).

To fix this, instead of processing one frame at a time, latent frames are now decoded in small, overlapping batches (tiles). Each batch has access to the cached context of all the previous ones (like it is already the case when decoding frames sequentially). To ensure seamless transitions, the final frame(s) of a batch act as an overlap; they are discarded from the current output and re-processed at the start of the next batch.

### Quick comparison

| | No tiling | frame-by frame |  tile = 4 ,  pad = 1 |
| --- | --- | --- | --- |
| result | <video src="https://github.com/user-attachments/assets/67e9d02d-502f-4838-8994-4a142db43a9e" /> | <video src="https://github.com/user-attachments/assets/c1cba3b0-f7d2-4d33-8691-2f86f09d8fbf" /> | <video src="https://github.com/user-attachments/assets/8cc4b6e4-e58b-45fc-9662-fee423c486cf" /> | 
| compute buffer size | 14381.63 MB | 2933.75 MB | 10143.13 MB | 
| decode_first_stage time | 28.77s | 13.64s | 18.09s |
| tiles processed | 1 (or 0?) | 13 | 5 | 

## Checklist

- [x] Use args instead of env variables


src/ltx_vae.hpp
```diff
@@ -1183,7 +1224,22 @@ struct LTXVideoVAE : public VAE {
         bool use_tiled = decode_graph && temporal_tiling_enabled &&
                          z_tensor.dim() == 5 && z_tensor.shape()[2] > 1;
         if (use_tiled) {
-            out = vae.decode_tiled(&runner_ctx, z, timestep);
+            // TODO: pass as args
+            int tile_frames = 1;
+            int tile_pad    = 0;
+            // use env variables for now
+            const char* env_tile_frames = std::getenv("VAE_TILE_FRAMES");
```
leejet: Maybe we can add something like `--extra-sample-args`, but for tiling, such as `--extra-tiling-args`, to avoid bloating the tiling parameters.

theaerotoad:
This is a massive improvement, especially for those on Vulkan who are vae-size limited!

leejet:

stduhpf:
Thank you for completing it! I was planning to but got lost in other side-quests before.

PR #1511, opened 2026-05-17
ci: add RDNA1 + RDNA2 targets for ROCm 7.13

sofiageo:
## Summary

ROCm 7.13 adds support for RDNA1 and RDNA2 targets. This PR also adds missing gfx1152 Kraken Point target.

## Checklist

- [ X] I have read and confirmed this PR follows the [contribution guidelines](https://github.com/leejet/stable-diffusion.cpp/blob/master/CONTRIBUTING.md).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -449,7 +449,7 @@ jobs:
 
     env:
       ROCM_VERSION: "7.13.0"
-      GPU_TARGETS: "gfx906;gfx908;gfx90a;gfx942;gfx950;gfx1100;gfx1101;gfx1102;gfx1150;gfx1151;gfx1200;gfx1201"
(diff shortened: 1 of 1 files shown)


superm1 (approved):

leejet:
@codex review

leejet:
It looks like adding these targets has made the CI runtime much longer.

superm1:
> It looks like adding these targets has made the CI runtime much longer.

Hopefully for the most part they should be cached in ccache though right?

leejet:
ccache can help for repeated builds, but the added GPU targets will still create separate cache entries. So the first CI run, or any run with a cold cache, will still take longer.

superm1:
> ccache can help for repeated builds, but the added GPU targets will still create separate cache entries. So the first CI run, or any run with a cold cache, will still take longer.

If it turns out to be a problem with CI taking too long overall; I would suggest we can drop the ROCm 7.2.x builds and only fixate on 7.13.

PR #1514, opened 2026-05-17
ltx-audio: use radix-2 fft to speed up compute_log_mel_spectrogram

stduhpf:
## Summary

This PR optimises one part of the LTX2.3 audio VAE by implementing Cooley-Tukey Radix-2 FFT algorithm to replace inlined DFT. Output sound quality is indistinguishable in my experience (as it should be).

(I also changed the intermediate `mel_value` to be float instead of double, this technically increases the footprint of this PR a bit, I don't think it affect performance or quality much either way, I kinda included it in this PR by mistake, but the code is a bit cleaner like that, so I'm leaving it in)

### Performace comparison, generating a 32x32 24fps video

- 121 frames (5s):

|  | compute_log_mel_spectrogram() time | speedup | Total audio decode time | Total speedup |
| DFT (master) | 21.315s | 1x | 209.086s | 1x |
| FFT (PR) | 1.284s | 16.6x | 188.718s | 1.1x |

- 9 frames (0.33s):

|  | compute_log_mel_spectrogram() time | speedup | Total audio decode time | Total speedup |
| DFT (master) | 1.566s | 1x | 4.577s | 1x |
| FFT (PR) | 119ms | 13.2x | 2.873s | 1.6x |

## Checklist

### DIFF
--- a/src/ltx_audio_vae.h
+++ b/src/ltx_audio_vae.h
@@ -5,6 +5,7 @@
 #include <numeric>
 #include <string>
 #include <vector>
+#include <complex>
 
 #include "ggml_extend.hpp"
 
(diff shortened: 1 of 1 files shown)


stduhpf:
Even with this change, the audio decoding remains very slow, especially for longer sequences. Maybe it would be worth investigating for using temporal tiling there too?

leejet:
I plan to add the missing operator to ggml, which should make it much faster. If this work takes too long or there are issues with the implementation, I’ll consider merging this PR.

leejet:
The current computation time of the audio VAE should be much faster than before. https://github.com/leejet/stable-diffusion.cpp/pull/1538

PR #1566, opened 2026-05-26
fix: prevent crash in case of a mem alloc error and graceful exit

akleine:
## Summary

Added a check against  nullptr  to avoid crash + core dump  in **model.cpp**

## Related Issue / Discussion

This test script 
`./sd-cli --diffusion-model ~/SD_models/flux/LongCat-Image-Q4_K_M.gguf --llm ~/SD_models/flux/Qwen2.5-VL-7B-Instruct.Q3_K_M.gguf  -p 'a lovely cat' --mmap`
ends  with  **Segmentation fault (core dumped)**  because there is not enough VRAM.

Here follows  some output for `version: stable-diffusion.cpp version master-647-72e512a-1-ga397e03, commit a397e03`

```
[INFO ] stable-diffusion.cpp:363  - Version: Longcat-Image                                                                                                                                    
[INFO ] stable-diffusion.cpp:391  - Weight type stat:                      f32: 387  |    q3_K: 113  |    q4_K: 221  |    q5_K: 23   |    bf16: 6                                             
[INFO ] stable-diffusion.cpp:392  - Conditioner weight type stat:          f32: 141  |    q3_K: 113  |    q4_K: 81   |    q5_K: 3                                                             
[INFO ] stable-diffusion.cpp:393  - Diffusion model weight type stat:      f32: 246  |    q4_K: 140  |    q5_K: 20   |    bf16: 6                                                             
[INFO ] stable-diffusion.cpp:394  - VAE weight type stat:                                                                                                                                     
[INFO ] flux.hpp:1291 - flux: depth = 10, depth_single_blocks = 20, guidance_embed = false, context_in_dim = 3584, hidden_size = 3072, num_heads = 24                                         
[INFO ] stable-diffusion.cpp:771  - using VAE for encoding / decoding                                                                                                                         
[ERROR] ggml_extend.hpp:69   - ggml_backend_cuda_buffer_type_alloc_buffer: allocating 3492.20 MiB on device 0: cudaMalloc failed: out of memory                                               
[ERROR] ggml_extend.hpp:69   - alloc_tensor_range: failed to allocate CUDA0 buffer of size 3661840640                                                                                         
[ERROR] ggml_extend.hpp:2699 - flux alloc params backend buffer failed, num_tensors = 412                                                                                                     
[INFO ] model.cpp:806  - NOT using mmap for '/home/xxx/SD_models/flux/LongCat-Image-Q4_K_M.gguf' (mmap disabled by caller)                                                                    
[INFO ] model.cpp:806  - NOT using mmap for '/home/xxx/SD_models/flux/Qwen2.5-VL-7B-Instruct.Q3_K_M.gguf' (mmap disabled by caller)                                                           
[INFO ] model.cpp:818  - model files processing completed in 0.00s                                                                                                                            
./0_test.sh: Zeile 1:  7259 Segmentation fault (core dumped) ./sd-cli --diffusion-model ~/SD_models/flux/LongCat-Image-Q4_K_M.gguf --llm ~/SD_models/flux/Qwen2.5-VL-7B-Ins
```
        
## Additional Information

After this patch the user  gets some information and we run into a clean exit: 
```
...

[INFO ] stable-diffusion.cpp:391  - Weight type stat:                      f32: 387  |    q3_K: 113  |    q4_K: 221  |    q5_K: 23   |    bf16: 6    
[INFO ] stable-diffusion.cpp:392  - Conditioner weight type stat:          f32: 141  |    q3_K: 113  |    q4_K: 81   |    q5_K: 3    
[INFO ] stable-diffusion.cpp:393  - Diffusion model weight type stat:      f32: 246  |    q4_K: 140  |    q5_K: 20   |    bf16: 6    
[INFO ] stable-diffusion.cpp:394  - VAE weight type stat:             
[INFO ] flux.hpp:1291 - flux: depth = 10, depth_single_blocks = 20, guidance_embed = false, context_in_dim = 3584, hidden_size = 3072, num_heads = 24
[INFO ] stable-diffusion.cpp:771  - using VAE for encoding / decoding
[ERROR] ggml_extend.hpp:69   - ggml_backend_cuda_buffer_type_alloc_buffer: allocating 3492.20 MiB on device 0: cudaMalloc failed: out of memory
[ERROR] ggml_extend.hpp:69   - alloc_tensor_range: failed to allocate CUDA0 buffer of size 3661840640
[ERROR] ggml_extend.hpp:2699 - flux alloc params backend buffer failed, num_tensors = 412
[ERROR] stable-diffusion.cpp:965  - Diffusion model params buffer allocation failed
[INFO ] main.cpp:761  - new_sd_ctx_t failed
```
 [message  **UPDATED** after more commits ]

## Checklist


src/model.cpp
```diff
@@ -995,6 +995,12 @@ bool ModelLoader::load_tensors(on_new_tensor_cb_t on_new_tensor_cb, int n_thread
                         continue;
                     }
 
+                    if (dst_tensor->data == nullptr) {
```
wbruna, line 1002: I'm not sure this is the best place to check for this, because the allocation can't fail for individual tensors; it's all-or-nothing.

Take a look at `StableDiffusionGGML::init`, in `stable-diffusion.cpp`, a bit before `load_tensors`: some components do report allocation failure, but not all. We should probably propagate the underlying `alloc_params_buffer` error in all cases.

akleine: I have seen that code in stable-diffusion.cpp and wondered about the reasoning behind it. I am just a newbie, and I guess **there is a reason** for writing the code this way.
So I decided to keep that as is - and better add this little quick fix.
And you’re right, the allocation problem is an all-or-nothing issue. So the message refers to the **first** tensor and then the program exits.

leejet, line 1002: I don't think `load_tensors()` is the right primary place to handle this.

`dst_tensor->data == nullptr` is only a downstream symptom. The actual allocation failure happens earlier in `StableDiffusionGGML::init()`: some components already check `alloc_params_buffer()` and return false on failure, but `cond_stage_model`, `diffusion_model`, and `high_noise_diffusion_model` currently ignore the return value.

Could we instead propagate those failures there, e.g. check:

- `cond_stage_model->alloc_params_buffer()`
- `diffusion_model->alloc_params_buffer()`
- `high_noise_diffusion_model->alloc_params_buffer()`

and return false with a component-level error message?

The null-data check in `load_tensors()` can still be kept as a defensive guard, but the current message `memory allocation failed '<tensor>'` is misleading because params buffer allocation is all-or-nothing, not per tensor.

leejet (changes requested):

leejet:
Please resolve the conflict.

akleine:
> Please resolve the conflict.

Just done.  
At first I didn't notice your refactoring  yesterday in commit #55c2aed.  
Sorry about this.

@leejet:   _Thank you for your work !_

wbruna (changes requested):
Missing this change:
> the current message memory allocation failed '<tensor>' is misleading because params buffer allocation is all-or-nothing, not per tensor.

And a minor thing: the cond_stage_model / diffusion_model / high_noise_diffusion_model tests should probably follow the same pattern as the others (`if (model && !alloc)` instead of `if (model) { if (!alloc) }`).

Otherwise LGTM.

akleine:
> Otherwise LGTM.

@wbruna : Thank you for your work!

leejet (approved):

leejet:

PR #1576, opened 2026-05-29
feat: `--stream-layers` for streaming weights from CPU during generation

fszontagh:
## Summary

This is the successor of #1477. That earlier PR did the same thing (stream model weights from CPU to GPU so larger models fit), but it ran as a parallel system alongside the existing graph-cut planner (#1476) and exposed many user-facing flags. Both points came up in the review.

This PR rebuilds the feature on top of the graph-cut planner instead of running alongside it. There is one new boolean flag, `--stream-layers`. When it is off, behavior is byte-identical to upstream master.

The change is split into two commits:

1. Foundation: adds the `--stream-layers` flag, a `SegmentResidency` annotation pass on the existing planner, and a small layer registry used by the runner. No behavior change when the flag is off.
2. Operational: chunk-K residency (keep the first K base segments on GPU across sampling steps), multi-runner safety (per-call free-VRAM clamp, restrict streaming to the diffusion runner, release residency before VAE decode), and runtime-LoRA correctness fallbacks (skip chunk-K when a `weight_adapter` is attached, prefer the runtime LoRA mode when streaming is on because immediate mode OOMs).

## Usage

Streaming kicks in when both `--max-vram` (or its auto sentinel `-1`) and `--stream-layers` are set. Weights need a place to stream from, so `--offload-to-cpu` is implicitly enabled if you forget it (with a log line).

```
./sd-cli \
  --diffusion-model models/z_image_turbo_bf16.safetensors \
  --llm models/Qwen3-4b-Z-Engineer-V2.gguf \
  --vae models/ae.safetensors \
  -p "a cat" --cfg-scale 1.0 -H 1024 -W 1024 --steps 8 \
  --max-vram -1 --offload-to-cpu --stream-layers \
  -o out.png
```

`--max-vram -1` auto-detects free VRAM and reserves 1 GiB headroom. Pass a positive value (e.g. `--max-vram 9`) to set the budget explicitly.

example output of the command above

## Tested

- 7 architecture smoke matrix at `--max-vram 4 --offload-to-cpu --stream-layers`: Z-Image Q8 and bf16, HiDream, Flux schnell, SD3.5 large, SDXL, WAN 2.2 5B. All generated valid output, no `cudaMalloc` failures.
- Z-Image Q8 with `--stream-layers off`: byte-identical PNG to upstream walker.
- Multi-LoRA correctness check: Z-Image bf16 1024x688, batch_count=2, runtime LoRA mode with two LoRAs stacked, `--cfg-scale 1.0 --guidance 3.5 --flow-shift 3.0`. Both batch images come out correct.

## Performance

The async-prefetch path that was in earlier iterations of this branch is intentionally disabled in this PR because of correctness regressions it caused with runtime LoRA and with `batch_count > 1` combined with non-default `--guidance` / `--flow-shift`. The implementation stays in the source for the follow-up to build on, but the engagement gate is `false`.

chunk-K residency is still active and saves H2D for the resident segments across all sampling steps. The wallclock benefit varies by model and `--max-vram` budget.

## Future work

The headline perf win, keeping the GPU near 100 percent utilization during streaming, needs PR #1477's `chunk_graph.hpp` helper ported on top of this foundation. That caches a fused cgraph for K base layers so the host does not need to issue per-layer kernel launches between H2D copies. It is the planned next PR, written specifically against this branch.

Other items I have queued for follow-ups:

- Per-module flag form: `--stream-layers diffusion,llm,vae`, once cross-runner residency eviction is in place. The current PR scopes streaming to the diffusion runner only.
- `SDCPP_STREAM_PROFILE` env-var-gated per-stage timing breakdown, useful for tuning.

## Checklist


nArn0:
Loved the previous PR, madly in love with this new one. It's going for a spin on my server asap 😍

fszontagh:
> Loved the previous PR, madly in love with this new one. It's going for a spin on my server asap 😍

I'm happy for it, thank you. 

Sorry, but now the preformance is slower than the prev. PR, and missing some optimizations yet which was implemented earlier. See Future work in the PR description. 
For example Z-Image turbo with batch 2, 9 steps at 688x1024px was ~1m27s on my PC, now its 2m17s with this PR. But at least large sized models can fit into the GPU with enough system RAM.

src/stable-diffusion.cpp
```diff
@@ -239,8 +245,25 @@ class StableDiffusionGGML {
...
+            stream_layers = false;
+        }
+        if (stream_layers && !offload_params_to_cpu && params_backend_spec.empty()) {
+            // Streaming needs weights somewhere they can stream FROM — i.e.
+            // params on CPU. Without --offload-to-cpu or an explicit
+            // --params-backend, weights would live on GPU, which makes
+            // streaming a silent no-op (params_backend == runtime_backend
+            // short-circuits should_use_graph_cut_segmented_compute).
+            LOG_INFO("--stream-layers requires CPU-resident weights; enabling --offload-to-cpu implicitly");
+            offload_params_to_cpu = true;
```
wbruna: This is a bit inconsistent, isn't it? I would expect that stream_layers either turned on both offload_params_to_cpu and max_vram, or got disabled with warnings in both cases. IMHO, at the library level a warning would be more consistent with how we handle other settings.

src/layer_registry.h
```diff
@@ -0,0 +1,49 @@
...
+struct LayerInfo {
+    std::vector<ggml_tensor*> tensors;
+    std::vector<ggml_tensor*> gpu_twins;
+    ggml_context*             twin_ctx   = nullptr;
+    ggml_backend_buffer_t     gpu_buffer = nullptr;
+    bool                      on_gpu     = false;
+    size_t                    bytes      = 0;
+};
+
+class Registry {
```
wbruna: Nit: `LayerRegistry` would be clearer for someone unfamiliar with the code.

src/stable-diffusion.cpp
fszontagh: Good point, swapped to warn-and-disable in `94be2a7` so it matches the `--max-vram` case.

src/layer_registry.h
fszontagh: Renamed to `LayerRegistry` in `fc9ddee`. Thanks!

src/ggml_extend.hpp
```diff
@@ -2334,17 +2622,120 @@ struct GGMLRunner {
...
+
+        // Split the effective budget so chunk-K residency and the prefetch
+        // window (2 segments + compute buffer + safety) coexist:
+        //   effective = chunk_K + 2*segment + reserve, with chunk_K == 2*segment
+        //   => segment = (effective - reserve) / 4
+        //   => chunk_K = 2 * segment
+        constexpr size_t reserve = 512ull * 1024 * 1024 + 768ull * 1024 * 1024;
+        size_t plan_budget       = effective_budget;
+        size_t residency_budget  = effective_budget;
+        if (stream_layers_enabled && effective_budget > 0) {
```
leejet: This if block is currently a no-op. Both plan_budget and  residency_budget are initialized to effective_budget before the branch, and the branch assigns them back to the same  value.

src/ggml_extend.hpp
```diff
@@ -1725,7 +1726,31 @@ struct GGMLRunner {
     ggml_context* partial_offload_ctx                   = nullptr;
     ggml_backend_buffer_t partial_runtime_params_buffer = nullptr;
     std::vector<std::pair<ggml_tensor*, ggml_tensor*>> partial_offload_pairs;
+
+    // Async-prefetch staging: a parallel partial-offload track used by the
```
leejet: There are quite a few comments in this change that read more like implementation notes or historical narration than necessary code comments. I’d suggest trimming them down and keeping only the comments that explain non-obvious behavior, invariants, or correctness constraints.

In general, if the code already expresses what is happening, the comment can be removed. Comments should be short and focused on why the logic exists, especially for the more complex state-management paths. For example, the async prefetch history and detailed rationale could be summarized more briefly, or moved to the PR description/design notes rather than kept inline.

src/ggml_extend.hpp
```diff
@@ -2334,17 +2622,120 @@ struct GGMLRunner {
...
+            for (const auto& seg : plan_out->segments) {
+                if (seg.input_param_bytes > largest_segment_params) {
+                    largest_segment_params = seg.input_param_bytes;
+                }
+            }
+            const size_t residency_for_annotate =
+                (residency_budget > largest_segment_params)
+                    ? (residency_budget - largest_segment_params)
+                    : 0;
+            sd::ggml_graph_cut::annotate_residency(*plan_out, residency_for_annotate);
```
leejet: There seem to be two separate residency annotations here: `resolve_graph_cut_plan()` annotates the resolved/merged plan_out, but `compute_streaming_segments()` later re-annotates `graph_cut_plan_cache_.graph_cut_plan` and builds the actual resident tensor set from that base plan.

  As a result, the residency marks on plan_out do not appear to be the source of truth for chunk-K residency today; they mostly affect logging / future registry behavior. This makes the budgeting path harder to reason about and can lead to the merged plan and base plan carrying different residency state.

  Could we make the ownership clearer by annotating only the plan that is actually used for residency, or by explicitly separating “execution residency labels” from “resident tensor selection”? That would avoid two independent annotations drifting apart.

src/ggml_extend.hpp
```diff
@@ -2630,6 +3038,371 @@ struct GGMLRunner {
...
+        //     prefetched buffers + compute buffer in --max-vram accumulate
+        //     enough FP error across the extra boundary-cache roundtrips to
+        //     collapse the diffusion output to pure white frames.
+        //
+        // Until both are properly fixed (most likely by porting PR #1477's
+        // chunk_graph.hpp — see unified_streaming_future_optimizations memory
+        // note), the synchronous segment loop below is the only validated
+        // streaming path. chunk-K residency still engages, which preserves
+        // most of the per-step H2D savings.
+        const bool prefetch_enabled = false;
```
leejet: Since async prefetch is currently disabled (`prefetch_enabled = false`) and the PR description notes known correctness regressions, I’d prefer to keep the prefetch implementation out of this PR and bring it back in the follow-up that actually enables/fixes it.

The pending offload path adds a fair amount of extra state (`pending_offload_ctx`, pending buffer/pairs, promote/discard logic, and the prefetch executor), but it does not affect runtime behavior today. Removing it from this PR would make the chunk-K residency change much easier to review and reduce the risk of carrying untested cleanup/swap logic in the main code path.

src/ggml_graph_cut.cpp
```diff
@@ -753,4 +753,58 @@ namespace sd::ggml_graph_cut {
...
+        if (max_graph_vram_bytes <= reserved) {
+            return;
+        }
+        const size_t available = max_graph_vram_bytes - reserved;
+
+        // Greedily mark segments RESIDENT from the start until we exceed the
+        // available budget. Honors heterogeneous segment sizes (small prelude +
+        // larger transformer layers) better than dividing by a single estimate.
+        size_t cumulative = 0;
+        for (auto& seg : plan.segments) {
```
leejet, line 797: The residency budget currently only accumulates `segment.input_param_bytes` and relies on a fixed 768 MB compute-buffer reserve. Since the plan already tracks `compute_buffer_size`, `output_bytes`, and cut/cache input sizes, it would be safer to account for the actual active segment memory footprint instead of using a fixed reserve.

 At minimum, the residency check should model resident params plus the largest active streamed segment params and compute buffer. Otherwise the number of resident segments can be over- or under-estimated depending on resolution/model/backend, and the fixed 768 MB reserve is hard to validate.

leejet (changes requested):
My main suggestion is to keep this PR focused on the chunk-K residency behavior, make the active path easier to reason about, and move the disabled async prefetch implementation to the follow-up PR where it will actually be fixed/enabled.

src/ggml_extend.hpp
fszontagh: Removed in `8c337ab`, both branches were assigning the same value.

src/ggml_extend.hpp
fszontagh: Trimmed in `8c337ab`, the history and disabled-feature notes are gone and what's left only covers the chunk-K invariants.

src/ggml_extend.hpp
fszontagh: Consolidated in `8c337ab`, only the base_plan annotation drives the resident set now; the merged plan no longer gets re-annotated.

src/ggml_extend.hpp
fszontagh: Ripped out in `8c337ab` (pending_offload track, prefetch executor, the gated dispatch, ~430 lines). Will come back in the follow-up that actually fixes the LoRA race and the small-segment FP drift.

src/ggml_graph_cut.cpp
fszontagh: Done in `06b1994`, the 768 MB constant is gone and the headroom now uses the plan-tracked footprint (params + compute buffer + output + cut/external inputs) of the worst-case segment.

leejet:

PR #1595, opened 2026-06-01
feat: PuLID-Flux identity-injection support

RapidMark:
Adds **PuLID-Flux** identity injection to the Flux denoise loop (works on CUDA / Vulkan / HIP / Metal). Given a single source portrait, generated images preserve the source person's face across arbitrary scenes and prompts. Pure-ggml implementation (every op has a backend kernel), so it's cross-vendor by construction.

This PR has three stages, all folded in below: **(1)** the original feature (#1542), **(2)** the gguf id-embedding rework (this PR, supersedes #1542), and **(3)** a rebase onto current master after the #1615 src-layout refactor.

## 1. The feature (original submission, #1542)

Mirrors the reference [ToTheBeginning/PuLID](https://github.com/ToTheBeginning/PuLID) (`encoders_transformer.py` + `flux/model.py`) and the PuLID v0.9.1 hook schedule (every 2nd of the double blocks, every 4th of the single blocks).

### What's included
- **`src/model/adapter/pulid.hpp`** — `PuLIDPerceiverAttentionCA`, the cross-attention module (Q from image tokens, K/V from the ID embedding). Pure-ggml graph; runs on CPU / CUDA / Vulkan / Metal without backend-specific code.
- **`src/model/diffusion/flux.hpp`** — adds the `pulid_ca.<i>` child blocks to `Flux` (constructed conditionally when PuLID weights are present), inserts the cross-attention between transformer blocks at the reference intervals (every 2nd double, every 4th single), and threads the identity embedding + weight through `forward` / `forward_orig` / `compute` / `build_graph`. `skip_layers` + PuLID is explicitly refused (would misalign the hook schedule).
- **`src/stable-diffusion.cpp`** — loads the `pulid_ca.*` weights via `model_loader` under the existing `model.diffusion_model.` prefix so they bind to the new blocks naturally, and loads the id-embedding, wrapping it as a `sd::Tensor<float>` passed via `DiffusionParams`.
- **`include/stable-diffusion.h`** — public API: `sd_pulid_params_t` (per-generation embedding path + weight), `pulid_weights_path` on `sd_ctx_params_t`, `pulid_params` on `sd_img_gen_params_t`.
- **`examples/common/common.{cpp,h}`** — three CLI flags: `--pulid-weights`, `--pulid-id-embedding`, `--pulid-id-weight`.
- **`src/model/diffusion/model.hpp`** — extends `DiffusionParams` to carry the embedding + weight; `FluxModel::compute` forwards both.
- **`docs/pulid.md`** — usage, embedding format, supported PuLID versions (v0.9.0 / v0.9.1; v1.1 deferred), memory-budget notes, and the three-way SHA-256 falsification recipe.
- **`scripts/pulid_extract_id.py`** — reference precompute tool that produces the id-embedding from a source portrait.

### Why split extraction from injection
PuLID-Flux's identity extractor is a stack of three large PyTorch models (ArcFace + EVA-CLIP-L + IDFormer). Porting all three to C++/ggml would add thousands of lines for code that runs once per source person. By making sd.cpp consume a precomputed embedding, the C++ surface stays small (~600 lines), the heavy ML stack runs once on any PyTorch backend, and PuLID is decoupled from active development on insightface / EVA-CLIP / IDFormer.

### Verification
The three-way SHA-256 falsification recipe in `docs/pulid.md` distinguishes "wired but inert" from "actively altering the trajectory":

| Run | Expected hash relation |
|-----|------------------------|
| A: no `--pulid-*` flags | baseline |
| B: PuLID flags, `--pulid-id-weight 0.0` | byte-identical to A |
| C: PuLID flags, `--pulid-id-weight 1.0` | differs from A, preserves source identity |

Verified on three backends from the same source:
- **Vulkan-AMD** (RX 6700 XT, `-DSD_VULKAN=ON`): A == B byte-identical, A != C, C preserves identity.
- **Vulkan-NVIDIA** (RTX 3060, same binary, `--backend "diffusion=vulkan1"`): A == B, A != C, C visually equivalent to the AMD output at the same seed (different bytes per the usual cross-backend nondeterminism).
- **CUDA-NVIDIA** (RTX 3060, separate `-DSD_CUDA=ON -DCMAKE_CUDA_ARCHITECTURES=86` build, CUDA 13.2): A == B byte-identical, A != C, C preserves identity. The PerceiverAttentionCA pure-ggml graph runs unchanged across all three — no backend-specific conditionals.

### Measurements
Per-image sampling at 512×512 / 4 steps / Flux Schnell Q4 + PuLID:

| Backend | Sampling (s) | Notes |
|---------|-------------:|-------|
| AMD 6700 XT (Vulkan) | 22 | 12 GB consumer card |
| NVIDIA 3060 (Vulkan) | 11 | same binary as AMD |
| NVIDIA 3060 (CUDA) | 9.6 | separate `-DSD_CUDA=ON` build |

`batch_count=3` confirms long-lived-process amortization: per-image sampling drops from 19.6 s (cold) to ~11 s (warm) as the model stays resident across iterations. Tested with Flux Schnell Q4_K_S at 512²/4 and Flux Dev Q4_K_S at 768²/20. At 1024² the VAE decode needs a large single compute buffer that can exceed a consumer card's max-allocation limit → use `--vae-tiling` (or route the VAE to a roomier backend); this is existing sd.cpp behavior, not PuLID-specific, but documented in `docs/pulid.md` since PuLID users hit it.

### Not yet supported (in docs/pulid.md)
- PuLID v1.1 (renamed key layout `id_adapter_attn_layers.*`). Follow-up.
- Multiple ID images fused into one embedding (reference pipeline supports it; the precompute tool takes one portrait per run).
- The `--true-cfg` negative branch — PuLID only injects on the positive conditioning path in the reference; this matches.

### Backward compatibility
Non-PuLID generations are unaffected: PuLID is only constructed when the loader sees a `pulid_ca.*` tensor. A regression run without `--pulid-*` flags is byte-identical to pre-patch.

## 2. Update — gguf id-embedding (this PR, supersedes #1542)

In the original submission the id-embedding was a **bespoke `PULIDV01` 32-byte-header binary** with a hand-rolled parser. Per @Green-Sky's review on #1542, this PR reworks it into a **standard GGUF container**: a single fp16 tensor `pulid_id` of shape `[2048, 32]`, loaded via `gguf_init_from_file` exactly like the `pulid_ca.*` weights. The custom header and parser are gone — one fewer on-disk format, and the embedding loads through the same path as every other tensor. `scripts/pulid_extract_id.py` writes the gguf; `docs/pulid.md` documents the gguf layout.

Opened fresh rather than force-pushing #1542 because that branch had drifted onto an old master and its history was tangled; a clean rebase was easier to review.

## 3. Follow-up — rebased onto the src/ layout refactor (#1615)

Master landed the **#1615 "reorganize src model layout"** refactor while this PR was open. This rebase re-homes PuLID onto the new tree, **no functional change**:
- `src/pulid.hpp` → `src/model/adapter/pulid.hpp` (includes updated to `core/ggml_extend.hpp` + `model/common/block.hpp`).
- The Flux injection moved with `src/flux.hpp` → `src/model/diffusion/flux.hpp`.
- Detection follows the refactor's factory: `FluxParams` renamed `FluxConfig`, weight-based config now in `FluxConfig::detect_from_weights`; PuLID auto-detection (`pulid_ca.` → `pulid_enabled`) moved into it.
- `DiffusionParams` plumbing follows `src/diffusion_model.hpp` → `src/model/diffusion/model.hpp`; includes updated to the new `src/{core,model}/...` paths.

The feature, hook schedule, and CLI/API surface are unchanged from stages 1–2.

## Validation (rebased branch)
- **Build: OK** — clean Vulkan build (`sd-cli` + `sd-server` link).
- **Identity: OK on RDNA4** (Radeon RX 9070 XT, Vulkan): Flux Krea-Dev Q4 + PuLID v0.9.1 + a freshly extracted gguf id-embedding, 1024² / 20 steps / dpm++2mv2 / cfg-scale 2.0 / guidance 3.5 / id-weight 1.0 / `--diffusion-fa` / `--vae-tiling`. Embedding loaded through `gguf_init_from_file` (`PuLID id-embedding: loaded [2048, 32] type=f16`), identity injected across all steps, source identity preserved in a scene unrelated to the reference photo.

## Usage
```
sd ... --diffusion-model flux1-dev.gguf --pulid-weights pulid_flux_v0.9.1.safetensors \
       --pulid-id-embedding face.pulidembd --pulid-id-weight 1.0
```
The `.pulidembd` is a gguf with one fp16 tensor `pulid_id [2048,32]` (produce it with `scripts/pulid_extract_id.py`).

### DIFF
--- a/docs/pulid.md
+++ b/docs/pulid.md
@@ -1,0 +1,195 @@
+# PuLID-Flux face-identity preservation
+
+stable-diffusion.cpp supports the [PuLID-Flux](https://github.com/ToTheBeginning/PuLID)
+identity-injection technique on top of Flux.1 (schnell or dev) models.
+Given a single source portrait, PuLID-Flux produces new generations that
+preserve the source person's face across arbitrary scenes, poses, and
+prompts.
+
+Unlike PhotoMaker (which extracts the identity inside the inference
+process from a directory of images), PuLID-Flux's identity extractor is
+a heavy stack (insightface ArcFace + EVA-CLIP-L + IDFormer encoder) that
+is impractical to port to C++/ggml. To keep this implementation small and
+cross-vendor, **stable-diffusion.cpp consumes a precomputed identity
+embedding** produced by an external Python tool that runs once per source
+portrait. Everything downstream of that one-shot extraction is C++ and
+runs on any backend (Vulkan, CUDA, Metal, ROCm, CPU).
+
+## Architecture summary
+
+The PuLID-Flux contribution to the Flux denoise loop is a stack of 20
+small cross-attention modules (`PerceiverAttentionCA`) inserted between
+the Flux transformer blocks:
+
+- After every 2nd of the 19 double-stream blocks (10 hook points)
+- After every 4th of the 38 single-stream blocks (10 hook points)
+
+Each cross-attention layer takes the current image tokens as query, the
+32-token / 2048-dim identity embedding as key+value, and adds its output
+(scaled by `id_weight`, typically 1.0) back to the image tokens.
+
+## Required weights
+
+Three files in addition to the standard Flux weight set:
+
+1. **Flux base** (transformer + VAE + clip_l + t5xxl) -- exactly as
+   [docs/flux.md](flux.md) describes.
+2. **PuLID weights** -- download from
+   [guozinan/PuLID](https://huggingface.co/guozinan/PuLID):
+   - `pulid_flux_v0.9.0.safetensors` or `pulid_flux_v0.9.1.safetensors`
+     (recommended; this implementation is verified against v0.9.1)
+   - **v1.1 (`pulid_v1.1.safetensors`) is NOT yet supported** -- it uses
+     renamed keys (`id_adapter_attn_layers.*` instead of `pulid_ca.*`)
+     and possibly different module structure. Future PR.
+3. **Identity embedding (.pulidembd)** -- produced by the precompute
+   tool below.
+
+## Precompute the identity embedding
+
+The precompute tool runs the PyTorch identity-extraction stack on a
+single portrait image and writes the resulting `(32, 2048)` embedding
+to a `.pulidembd` binary file (about 131 KB). Run it once per source
+person; the same file is reused for any number of generations.
+
+A reference Python script is provided alongside this docs file at
+[`scripts/pulid_extract_id.py`](../scripts/pulid_extract_id.py). It
+requires:
+- A working CUDA / CPU PyTorch + diffusers stack
+- `insightface`, `facexlib`, `eva-clip`, `torchvision`
+- The PuLID weights file (same one stable-diffusion.cpp will load below)
+- The ToTheBeginning/PuLID repo's `pulid/pipeline_flux.py` (and its
+  dependencies under `pulid/` and `flux/`) -- recommended to vendor
+  rather than pip-install due to upstream packaging quirks
+
+Run it as:
+
+```
+python pulid_extract_id.py \
+  --portrait /path/to/source-photo.jpg \
+  --pulid-weights /path/to/pulid_flux_v0.9.1.safetensors \
+  --out /path/to/source.pulidembd
+```
+
+## Format (gguf)
+
+The embedding is a standard **gguf** container holding a single tensor:
+
+```
+tensor name : "pulid_id"
+shape       : [token_dim, num_tokens]   (ggml order; typically [2048, 32])
+type        : F16 (also accepts F32 / BF16)
+metadata    : general.architecture = "pulid", pulid.version = 1
+```
+
+stable-diffusion.cpp loads it with the normal gguf reader
+(`gguf_init_from_file`) and converts to fp32 at load time -- no bespoke
+parser. Total file size for the typical (32, 2048, fp16) case is ~131 KB.
+
+## Command-line usage
+
+```
+.\bin\Release\sd-cli.exe \
+  --diffusion-model     models\flux1-schnell-Q4_K_S.gguf \
+  --vae                 models\ae.safetensors \
+  --clip_l              models\clip_l.safetensors \
+  --t5xxl               models\t5xxl_fp16.safetensors \
+  --pulid-weights       models\pulid_flux_v0.9.1.safetensors \
+  --pulid-id-embedding  source.pulidembd \
+  --pulid-id-weight     1.0 \
+  -p "candid photograph of a young woman on a beach at sunset" \
+  --cfg-scale 1.0 --sampling-method euler --steps 4 -W 512 -H 512 \
+  --seed 42 --clip-on-cpu \
+  -o out.png
+```
+
+For Flux Dev (instead of Schnell), add `--guidance 3.5` and `--steps 20`.
+
+## Flags
+
+| Flag                       | Purpose                                                           |
+|----------------------------|-------------------------------------------------------------------|
+| `--pulid-weights <path>`   | Path to `pulid_flux_v0.9.x.safetensors`. Loaded with the model.   |
+| `--pulid-id-embedding <p>` | Path to a `.pulidembd` binary produced by the precompute tool.    |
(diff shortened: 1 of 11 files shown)


wbruna:
I believe this should be adapted to use the extension mechanism introduced in 2a07540c2a2d06e8b5ba6cc82cb90c26f4d414c1 , in a similar way that was done with Photomaker.

RapidMark:
Thanks — agreed in intent: PuLID should follow the extension pattern where it can.

The catch is it doesn't fit as literally framed. PhotoMaker works entirely by modifying the conditioning (`prepare_condition` / `before_condition` cover it). PuLID injects identity via cross-attention *inside the diffusion forward* — the `pulid_ca` blocks, fed the precomputed id-embedding through `DiffusionParams` — so the conditioning-side hooks don't reach it. Adapting PuLID to the extension mechanism would mean **extending the interface** (a `before_diffusion` / contribute-to-`DiffusionParams` hook), and whether to add that is really a maintainer call — not something I'd want to build unilaterally.

Concretely, I'd move PuLID's lifecycle — weight loading, the id-embedding, enable/disable, param registration — into a `GenerationExtension` (mirroring `photomaker_extension`). But the cross-attention injection has to stay in the Flux forward, which is why the extension needs that small hook to feed the embedding into the diffusion pass.

@leejet — flagging you, since this hinges on that small addition to the `GenerationExtension` interface (the diffusion-forward hook), which is your design call. Does that direction sound right, or would you prefer PuLID route its embedding a different way? Happy to do the refactor once it's clear.

leejet:
I think for this case, we could add a new extension hook to modify `DiffusionParams`.

RapidMark:
Would you like me to take a GO at it?

RapidMark:
Done — refactored onto the extension mechanism per your suggestion. Added one general hook, `before_diffusion(DiffusionParams&, int step)`, the diffusion-forward analog of `before_condition`: called for each enabled extension after the per-step `DiffusionParams` (and its `extra`) is built, before `compute()`. Any extension/architecture can set fields on `params` (typically the matching `extra` variant) — it's not PuLID-specific.

PuLID's runtime now lives in `src/extensions/pulid_extension.cpp` (enable-detect, per-gen gguf id-embedding load, and the injection into `FluxDiffusionExtra`). The only PuLID code left outside the extension is the `pulid_ca.*` weight merge in the ctor — those are Flux model weights and have to be in the tensor map before the model is built (before any extension `init()` runs). The old per-sample threading through `sample()` is gone.

Built + validated on Flux Krea Q4 + PuLID v0.9.1 (Vulkan): embed loads `[2048, 32]`, and off-vs-on with identical seed/prompt produces a coherent identity-shifted portrait.

One idiom call: `pulid_params` rides in the condition context next to the existing `pm_params` — happy to generalize that into a per-extension params carrier if you'd prefer.

leejet:
Thanks for the PR!

I tried to push the conflict-resolution changes directly to this PR branch, since “Allow edits by maintainers” is enabled. However, GitHub denied the push:

```text
ERROR: Permission to CloudhandsAI/stable-diffusion.cpp.git denied to leejet.
fatal: Could not read from remote repository.
```

It looks like this is because the PR branch comes from an organization-owned fork (`CloudhandsAI/stable-diffusion.cpp`), and GitHub does not grant maintainers push access to organization-owned forks via the “Allow edits by maintainers” option.

I resolved the conflicts on my side and am ready to merge the PR. For future PRs, would it be possible to open them from a branch in your personal fork instead? That would allow maintainers to push small fixes or conflict-resolution changes directly when needed.

RapidMark:
Thanks @leejet — and thanks for resolving the conflicts on your side. To make the merge easy straight from the PR, I've rebased this branch onto current master, resolved the conflicts, and pushed — so it should merge cleanly from the button now.

On the org-fork push limitation: noted — we're looking into a personal-fork setup for future PRs so you can push small fixes directly, and I'll confirm once that's sorted.

Thanks again for the `before_diffusion` hook suggestion.

leejet:

PR #1615, opened 2026-06-06
refactor: reorganize src model layout

leejet:
## Summary

- Reorganize internal `src` files into `core`, `runtime`, `conditioning`, and `model/*` directories.

## Related Issue / Discussion

N/A

## Additional Information

N/A

## Checklist


wbruna:
@leejet , while we're at it: could we rename `gits_noise.inl` to `gits_noise.h`? We keep having to special-case the extension everywhere (CMakeLists.txt, formatting, synchronizing source trees, ...); and IMHO it's not a useful distinction for that single file, since we already have several other implementation-only .h and .hpp files.

leejet:
That’s fine. I kept `.inl` only because the original contributor used the `.inl` file extension, and I didn’t change it.

PR #1621, opened 2026-06-08
feat: automatically fall back to VAE tiling when an untiled decode exceeds the backend buffer limit

RapidMark:
On memory-constrained backends — integrated GPUs especially — a full-image VAE decode allocates a single compute buffer larger than the backend's maximum single-buffer/allocation size, and sd.cpp **hard-fails** instead of falling back to the tiling it already supports. The user has to know to pass `--vae-tiling` up front; otherwise the run crashes at the very end, after sampling has already completed.

### Repro

AMD Radeon 8060S (Strix Halo, RDNA3.5 iGPU, 128 GB unified memory), Vulkan backend, Flux Krea-dev Q4 at 1024×1024, with no tiling flag:

```
[INFO ] stable-diffusion.cpp - sampling completed, taking 10.24s
ggml_vulkan: Requested buffer size exceeds device buffer size limit: ErrorOutOfDeviceMemory
[ERROR] ggml_extend.hpp - vae: failed to allocate the compute buffer
[ERROR] vae.hpp - vae decode compute failed
[ERROR] main.cpp - generate failed
```

The ~8.5 GB single-shot VAE decode buffer exceeds the iGPU's Vulkan per-buffer limit. The card has ample *total* memory (it shares 128 GB system RAM) — the failure is the **per-buffer** ceiling, not capacity. The whole gen is lost after a successful sampling pass.

### Change

Add an **automatic fallback to tiling**, on by default, and keep it non-breaking:

- **`--vae-tiling` stays exactly as it was** — a boolean flag that forces tiling on.
- **The auto-fallback is the new default.** Before allocating the untiled decode buffer, the planned size is measured from the graph with `ggml_gallocr_reserve_n_size` (no-alloc planning, zero allocation) and compared against `ggml_backend_buft_get_max_size()`; if it won't fit, the decode goes straight to tiling. This is non-breaking — a decode that previously fit behaves identically, and one that previously OOM'd now recovers — and strictly safer. On CPU `get_max_size()` is `SIZE_MAX`, so it no-ops there.
- **`--no-vae-tiling-fallback`** disables the fallback for anyone who wants the old hard-fail behavior.
- A reactive backstop remains: if the untiled `_compute` still returns empty at runtime (e.g. the planned size fit the max but the device is genuinely full), it frees the buffer and retries once tiled — so a true OOM is also covered.

Implemented with a `bool auto_tile` appended to the end of `sd_tiling_params_t` (kept at the end so the C ABI stays backward-compatible; default `true`), the proactive probe in `GGMLRunner::alloc_compute_buffer`, and the fallback branch in `VAE::decode`.

Choosing the **real graph-planned size** (not a hardcoded bytes-per-pixel estimate) keeps it correct across every VAE architecture (SD/SDXL/Flux/Wan/LTX) and backend with no tuning.

### Validation (AMD Radeon 8060S iGPU, Krea Q4, 1024²)

- default (no flag) → `vae: untiled decode buffer exceeded the backend limit; retrying with tiling`, completes, **exit 0**
- `--no-vae-tiling-fallback` → fails at decode, **exit 1** (the old behavior, opt-in)
- `--vae-tiling` → tiles from the start, **exit 0**

The tiled GPU decode (~6.9 s) is also far faster than the usual workaround of routing the VAE to CPU (~29.5 s) to dodge the OOM, and is visually equivalent at 0.5 tile overlap.

Helps any constrained device, not just iGPUs — an 8 GB discrete card at high resolution hits the same per-buffer wall. Scoped to `decode` (where the failure occurs); `encode` has the same shape and could get the identical treatment later.

---

Thanks to @wbruna for pushing toward the proactive graph-planned size, and @stduhpf for catching that the original tristate would have broken the `--vae-tiling` syntax (this revision keeps it a plain flag + auto-by-default + opt-out).

### DIFF
--- a/examples/common/common.cpp
+++ b/examples/common/common.cpp
@@ -958,7 +958,7 @@ ArgOptions SDGenerationParams::get_options() {
          &extra_sample_args},
         {"",
          "--extra-tiling-args",
-         "extra VAE tiling args, key=value list. LTX video VAE supports temporal_tile_frames (default: 4), temporal_tile_overlap (default: 1)",
+         "extra VAE tiling args, key=value list. max_buffer_size (bytes) forces the auto fallback to tile when an untiled VAE compute buffer would exceed it. LTX video VAE supports temporal_tile_frames (default: 4), temporal_tile_overlap (default: 1)",
          (int)',',
          &extra_tiling_args},
     };
@@ -1148,6 +1148,12 @@ ArgOptions SDGenerationParams::get_options() {
          "process vae in tiles to reduce memory usage",
          true,
          &vae_tiling_params.enabled},
+        {"",
+         "--no-vae-tiling-fallback",
+         "disable the automatic fallback to VAE tiling when an untiled decode would exceed the "
+         "backend's max buffer size (fail instead of tiling)",
+         false,
+         &vae_tiling_params.auto_tile},
         {"",
          "--temporal-tiling",
          "enable temporal tiling for LTX video VAE decode",
@@ -1891,6 +1897,9 @@ bool SDGenerationParams::from_json_str(
         const json& tiling_json = j["vae_tiling_params"];
         if (tiling_json.contains("enabled") && tiling_json["enabled"].is_boolean()) {
             vae_tiling_params.enabled = tiling_json["enabled"];
+        }
+        if (tiling_json.contains("auto_tile") && tiling_json["auto_tile"].is_boolean()) {
+            vae_tiling_params.auto_tile = tiling_json["auto_tile"];
         }
         if (tiling_json.contains("temporal_tiling") && tiling_json["temporal_tiling"].is_boolean()) {
             vae_tiling_params.temporal_tiling = tiling_json["temporal_tiling"];
@@ -2711,10 +2720,12 @@ std::string build_sdcpp_image_metadata_json(const SDContextParams& ctx_params,
     }
 
     if (gen_params.vae_tiling_params.enabled ||
+        !gen_params.vae_tiling_params.auto_tile ||
         gen_params.vae_tiling_params.temporal_tiling ||
         !gen_params.extra_tiling_args.empty()) {
         root["vae_tiling"] = {
             {"enabled", gen_params.vae_tiling_params.enabled},
+            {"auto_tile", gen_params.vae_tiling_params.auto_tile},
             {"temporal_tiling", gen_params.vae_tiling_params.temporal_tiling},
             {"tile_size_x", gen_params.vae_tiling_params.tile_size_x},
             {"tile_size_y", gen_params.vae_tiling_params.tile_size_y},
--- a/include/stable-diffusion.h
+++ b/include/stable-diffusion.h
@@ -153,7 +153,7 @@ enum lora_apply_mode_t {
 };
 
 typedef struct {
-    bool enabled;
+    bool enabled;  // true => always tile (ON)
     bool temporal_tiling;
     int tile_size_x;
     int tile_size_y;
@@ -161,6 +161,7 @@ typedef struct {
     float rel_size_x;
     float rel_size_y;
     const char* extra_tiling_args;
+    bool auto_tile;  // AUTO (default): tile only when an untiled VAE decode would exceed the backend's max buffer size
 } sd_tiling_params_t;
 
 typedef struct {
--- a/src/core/ggml_extend.hpp
+++ b/src/core/ggml_extend.hpp
@@ -1705,10 +1705,17 @@ protected:
 
     ggml_context* compute_ctx    = nullptr;
     ggml_gallocr* compute_allocr = nullptr;
+    // Set when alloc_compute_buffer deferred to tiling on purpose (not a failure).
+    bool compute_buffer_deferred_to_tiling = false;
 
     size_t max_graph_vram_bytes           = 0;
     bool stream_layers_enabled            = false;
     size_t observed_max_effective_budget_ = 0;
+
+    // When set, alloc_compute_buffer declines a too-large untiled decode so VAE AUTO can tile.
+    bool probe_compute_buffer_fits_ = false;
+    // Optional user cap (bytes) to force tiling; 0 = no cap.
+    size_t probe_max_bytes_ = 0;
 
     std::shared_ptr<WeightAdapter> weight_adapter = nullptr;
     std::weak_ptr<RunnerWeightManager> weight_manager;
@@ -1978,10 +1985,77 @@ protected:
     }
 
     bool alloc_compute_buffer(ggml_cgraph* gf) {
+        compute_buffer_deferred_to_tiling = false;
         if (compute_allocr != nullptr) {
             return true;
         }
-        compute_allocr = ggml_gallocr_new(ggml_backend_get_default_buffer_type(runtime_backend));
+        ggml_backend_buffer_type_t buft = ggml_backend_get_default_buffer_type(runtime_backend);
+
+        if (probe_compute_buffer_fits_) {
+            // Measure the planned untiled compute buffer once (no allocation), then defer to
+            // tiling before the real reserve hits a raw backend error.
+            ggml_gallocr* probe = ggml_gallocr_new(buft);
+            size_t sizes[1]     = {0};
+            ggml_gallocr_reserve_n_size(probe, gf, nullptr, nullptr, sizes);
+            ggml_gallocr_free(probe);
+            size_t planned = sizes[0];
+
+            // User cap (extra_tiling_args max_buffer_size), any backend.
+            if (probe_max_bytes_ > 0 && planned > probe_max_bytes_) {
+                LOG_DEBUG("%s: untiled compute buffer %.2f MB exceeds requested max_buffer_size %.2f MB; deferring to tiling",
+                          get_desc().c_str(),
+                          planned / 1024.0 / 1024.0,
+                          probe_max_bytes_ / 1024.0 / 1024.0);
+                compute_buffer_deferred_to_tiling = true;
+                return false;
+            }
+
+            // Free VRAM, any non-CPU backend: a decode can fit every op's per-buffer cap yet
+            // still exceed total free VRAM. Margin covers the scratch pool the reserve omits.
+            ggml_backend_dev_t dev = ggml_backend_get_device(runtime_backend);
+            if (dev != nullptr && ggml_backend_dev_type(dev) != GGML_BACKEND_DEVICE_TYPE_CPU) {
+                size_t free_vram = 0, total_vram = 0;
+                ggml_backend_dev_memory(dev, &free_vram, &total_vram);
+                size_t margin = planned / 3;
+                if (margin < 512ull * 1024 * 1024) {
+                    margin = 512ull * 1024 * 1024;
+                }
+                if (free_vram > 0 && free_vram < planned + margin) {
+                    LOG_DEBUG("%s: untiled compute buffer %.2f MB won't fit free VRAM; deferring to tiling",
+                              get_desc().c_str(),
+                              planned / 1024.0 / 1024.0);
+                    compute_buffer_deferred_to_tiling = true;
+                    return false;
+                }
+            }
+
+            // Per-buffer cap: Vulkan via supports_op (the real limit; buft_get_max_size only
+            // reports the suballocation block there), other backends via buft_get_max_size.
+            if (sd_backend_is(runtime_backend, "Vulkan")) {
+                for (int i = 0; i < ggml_graph_n_nodes(gf); ++i) {
+                    ggml_tensor* op = ggml_graph_node(gf, i);
+                    if (!ggml_backend_supports_op(runtime_backend, op)) {
+                        LOG_DEBUG("%s: untiled compute op %.2f MB exceeds backend support; deferring to tiling",
+                                  get_desc().c_str(),
+                                  ggml_nbytes(op) / 1024.0 / 1024.0);
+                        compute_buffer_deferred_to_tiling = true;
+                        return false;
+                    }
+                }
+            } else {
+                size_t max_size = ggml_backend_buft_get_max_size(buft);
+                if (max_size > 0 && planned > max_size) {
+                    LOG_DEBUG("%s: untiled compute buffer %.2f MB exceeds backend max single buffer %.2f MB; deferring to tiling",
+                              get_desc().c_str(),
+                              planned / 1024.0 / 1024.0,
+                              max_size / 1024.0 / 1024.0);
+                    compute_buffer_deferred_to_tiling = true;
+                    return false;
+                }
+            }
+        }
+
+        compute_allocr = ggml_gallocr_new(buft);
 
         if (!ggml_gallocr_reserve(compute_allocr, gf)) {
             // failed to allocate the compute buffer
@@ -2432,7 +2506,9 @@ protected:
         GraphWeightDoneGuard graph_weight_done_guard(this, &params_to_prepare);
 
         if (!alloc_compute_buffer(gf)) {
-            LOG_ERROR("%s alloc compute buffer failed", get_desc().c_str());
+            if (!compute_buffer_deferred_to_tiling) {
+                LOG_ERROR("%s alloc compute buffer failed", get_desc().c_str());
+            }
             return std::nullopt;
         }
         struct ComputeBufferGuard {
@@ -2821,6 +2897,11 @@ public:
 
     void set_stream_layers_enabled(bool enabled) {
         stream_layers_enabled = enabled;
+    }
+
+    void set_probe_compute_buffer_fits(bool enabled, size_t max_bytes = 0) {
+        probe_compute_buffer_fits_ = enabled;
+        probe_max_bytes_           = enabled ? max_bytes : 0;
     }
 };
 
--- a/examples/common/common.h
+++ b/examples/common/common.h
@@ -229,7 +229,7 @@ struct SDGenerationParams {
     int video_frames                     = 1;
     int fps                              = 16;
     float vace_strength                  = 1.f;
-    sd_tiling_params_t vae_tiling_params = {false, false, 0, 0, 0.5f, 0.0f, 0.0f, nullptr};
+    sd_tiling_params_t vae_tiling_params = {false, false, 0, 0, 0.5f, 0.0f, 0.0f, nullptr, true};
     std::string extra_tiling_args;
 
     std::string pm_id_images_dir;
--- a/examples/server/api.md
+++ b/examples/server/api.md
@@ -518,7 +518,7 @@ Shared default fields used by both `img_gen` and `vid_gen`:
 | `output_format` | `string` |
 | `output_compression` | `integer` |
 
-`vae_tiling_params.extra_tiling_args` accepts a key=value list. For LTX video VAE temporal tiling, `temporal_tile_frames` defaults to `4` and `temporal_tile_overlap` defaults to `1`.
+`vae_tiling_params.extra_tiling_args` accepts a key=value list. `max_buffer_size` (bytes) forces the automatic tiling fallback when an untiled VAE compute buffer would exceed it. For LTX video VAE temporal tiling, `temporal_tile_frames` defaults to `4` and `temporal_tile_overlap` defaults to `1`.
 
 `img_gen`-specific default fields:
 
--- a/src/model/vae/vae.hpp
+++ b/src/model/vae/vae.hpp
@@ -199,7 +199,52 @@ public:
                 "vae decode compute failed while processing a tile",
                 silent);
         } else {
+            // AUTO: probe first so a too-large decode tiles instead of erroring; output.empty() backstops a real OOM.
+            const bool auto_probe = !tiling_params.enabled && tiling_params.auto_tile;
+            if (auto_probe) {
+                size_t max_bytes = 0;
+                if (tiling_params.extra_tiling_args != nullptr) {
+                    for (const auto& [key, value] : parse_key_value_args(tiling_params.extra_tiling_args, "VAE extra tiling arg")) {
+                        if (key == "max_buffer_size") {
+                            max_bytes = strtoull(value.c_str(), nullptr, 10);
+                        }
+                    }
+                }
+                set_probe_compute_buffer_fits(true, max_bytes);
+            }
             output = _compute(n_threads, input, true);
+            if (auto_probe) {
+                set_probe_compute_buffer_fits(false);
+            }
+            if (output.empty() && !tiling_params.enabled && tiling_params.auto_tile) {
+                free_compute_buffer();
+                if (!silent) {
+                    LOG_WARN("vae: untiled decode buffer exceeded the backend limit; retrying with tiling");
+                }
+                sd_tiling_params_t auto_tiling = tiling_params;
+                auto_tiling.enabled            = true;
+                set_tiling_params(auto_tiling);
+                const int scale_factor = get_scale_factor();
+                int64_t W              = input.shape()[0] * scale_factor;
+                int64_t H              = input.shape()[1] * scale_factor;
+                float tile_overlap;
+                int tile_size_x, tile_size_y;
+                get_tile_sizes(tile_size_x, tile_size_y, tile_overlap, auto_tiling, input.shape()[0], input.shape()[1]);
+                output = tiled_compute(
+                    input,
+                    n_threads,
+                    static_cast<int>(W),
+                    static_cast<int>(H),
+                    scale_factor,
+                    tile_size_x,
+                    tile_size_y,
+                    tile_overlap,
+                    circular_x,
+                    circular_y,
+                    true,
+                    "vae decode compute failed while processing a tile",
+                    silent);
+            }
         }
 
         free_compute_buffer();
--- a/src/stable-diffusion.cpp
+++ b/src/stable-diffusion.cpp
@@ -192,7 +192,7 @@ public:
     bool apply_lora_immediately = false;
 
     std::string taesd_path;
-    sd_tiling_params_t vae_tiling_params = {false, false, 0, 0, 0.5f, 0, 0, nullptr};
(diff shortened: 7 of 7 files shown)


wbruna:
> I chose **reactive** (retry on the real allocation failure) over **proactive** (estimate the buffer size and compare against `ggml_backend_buft_get_max_size()`) deliberately: a size estimate is **VAE-architecture-specific** (peak activation differs across SD/SDXL/Flux/Wan/LTX VAEs), so a hardcoded bytes-per-pixel constant would be brittle

Wouldn't be possible to check with the real value, calculated from the graph before the allocation?

RapidMark:
Good call — done (pushed just now).

Instead of retrying on failure, the AUTO path now measures the planned compute-buffer size up front with `ggml_gallocr_reserve_n_size` (no-alloc planning, zero allocation) and compares it to `ggml_backend_buft_get_max_size()` before allocating; if it won't fit, it goes straight to tiling — no bytes-per-pixel estimate. On CPU `get_max_size()` is `SIZE_MAX`, so it no-ops there.

I kept the original retry-on-empty as a backstop for a genuine runtime OOM (planned size fits the max, but the device is actually full). Net effect on the auto path: the backend no longer prints its raw "allocation failed" error — just an INFO line and the tiled decode.

Validated on an AMD Radeon 8060S iGPU (Krea Q4, 1024²): `--vae-tiling auto` now logs only the INFO "retrying with tiling" and completes; `off` still fails; `on` still tiles.

stduhpf:
I think having a fallback to vae tiling is a much welcome addition, but I'm having some small issues with the user experience there. Modifying the syntax of `--vae-tiling` arg from a flag to a tistate option will break previously working commands, and I think we could implement the same feature without breakling anything.

For example we could add a `--vae-tiling-auto` flag.

Alternatively, set "auto" tiling as default and add something like a `--no-auto-vae-tiling`  to disable it.

RapidMark:
Thanks — agreed, changing `--vae-tiling`'s syntax wasn't worth the breakage. Pushed a revision (your alternative B):

- `--vae-tiling` is back to the original boolean flag (force tiling on).
- The auto-fallback is now the **default** — it only tiles when an untiled decode would exceed the backend's max buffer size, so it's non-breaking and strictly safer for existing commands (an untiled decode that previously fit behaves identically; one that previously OOM'd now recovers).
- `--no-vae-tiling-fallback` disables it for the old hard-fail behavior.

Validated on an AMD Radeon 8060S iGPU (Krea Q4, 1024²): default auto-recovers (logs `vae: untiled decode buffer exceeded the backend limit; retrying with tiling`, exit 0); `--no-vae-tiling-fallback` fails (exit 1); `--vae-tiling` tiles from the start (exit 0).

include/stable-diffusion.h
```diff
@@ -153,14 +153,19 @@ enum lora_apply_mode_t {
...
     int tile_size_x;
     int tile_size_y;
     float target_overlap;
     float rel_size_x;
     float rel_size_y;
     const char* extra_tiling_args;
+    // Tristate with `enabled`: enabled => ON (always tile); else auto_tile => AUTO (tile only when
+    // an untiled VAE compute buffer can't be allocated, e.g. it exceeds the backend's max buffer
+    // size on an iGPU); else OFF (never tile, fail if the untiled buffer doesn't fit). Default AUTO.
+    // Appended (rather than folded into an enum) to keep the struct ABI backward-compatible.
```
wbruna: I think these kind of detailed comments would be fine _if_ the file already had detailed comments everywhere else.

Also (just suggestions, I don't know what @leejet  would prefer): maybe we could use an `extra_tiling_args` parameter instead of a separate flag, since it'd be more useful as a workaround or for testing?

And rather than a simple on/off switch, maybe we could receive a threshold override here? Say, -1 for disabling, 0 for auto, > 0 as the new limit. That way, the user could increase it if they know the device can handle it, decrease it to save VRAM for other reasons, etc (working around the Vulkan 1G limit would be an immediate use case, too).

src/core/ggml_extend.hpp
```diff
@@ -1895,10 +1907,39 @@ struct GGMLRunner {
...
+        if (probe_compute_buffer_fits_) {
+            // Measure the planned compute buffer WITHOUT allocating (no_alloc
+            // planning) and bail before the real reserve if it exceeds the
+            // backend's max single-buffer size. This lets the caller (VAE AUTO
+            // tiling) fall back to tiling without the backend ever emitting its
+            // raw "allocation failed" error on the successful auto path. A
+            // genuine runtime OOM (planned size <= max, but the device is full)
+            // is NOT caught here -- it still surfaces from the real reserve
+            // below, so the reactive fallback remains the backstop.
+            size_t max_size = ggml_backend_buft_get_max_size(buft);
```
wbruna: I believe the size calculation is needlessly including the VAE weights here. By default I get:
> [DEBUG] ggml_extend.hpp:1932 - vae: untiled compute buffer 2112.06 MB exceeds backend max single buffer 1024.00 MB; deferring to tiling

Fiddling with `GGML_VK_SUBALLOCATION_BLOCK_SIZE` env var, it shows:
> [DEBUG] ggml_extend.hpp:1953 - vae compute buffer size: 1920.06 MB(VRAM)

But this may be kind of a moot point, at least on Vulkan: as far as I can tell by looking at the ggml code, the limit on Vulkan will by default be capped at 1G anyway (`suballocation_block_size`, which is the value reported by `ggml_backend_buft_get_max_size`; the more useful `max_buffer_size` doesn't seem to be accessible from the API 😕).

wbruna (changes requested):
These are mostly suggestions, but I'm marking as 'changes needed' anyway because Claude is listed as a co-author (see [CONTRIBUTING.md](https://github.com/leejet/stable-diffusion.cpp/blob/master/CONTRIBUTING.md#ai-assisted-contributions)).

pwilkin:
FWIW this is part of my sweeping memory-management / lazy-loading changes in #1470 as well.

RapidMark:
Thanks @pwilkin, good to know. This one's deliberately narrow — it's an OOM safety-net in the VAE decode path: when an untiled decode would exceed the backend's max buffer size, it automatically falls back to tiling instead of erroring out. So it should be complementary to the broader memory-management / lazy-loading in #1470 rather than overlapping it. Happy to rebase on top of #1470, or defer to it entirely if you and leejet would rather land the consolidated approach — whatever keeps things cleanest.

RapidMark:
@leejet — before I make any code changes here, I'd like your call on direction, since @wbruna raised two API options and deferred to your preference:

**1. auto-fallback API.** Currently a tristate on the VAE-tiling flag (off / on / auto — where auto tiles only when an untiled VAE compute buffer would exceed the backend's max single-buffer size). @wbruna suggested a threshold override instead — `-1` disable, `0` auto, `> 0` a user-set byte limit — letting a user raise the cap on a capable device, lower it to save VRAM, or work around Vulkan's ~1 GB suballocation cap. He also floated an `extra_tiling_args`-style parameter. Keep auto/on/off, or move to the threshold override?

**2. size measurement.** @wbruna noted the proactive measurement may include the VAE weights alongside the compute buffer (~2112 MB vs ~1920 MB), though it's likely moot on Vulkan since `ggml_backend_buft_get_max_size` returns the 1 GB suballocation block there. Refine, or leave as-is?

I'll hold the implementation until you point a direction, then make the changes.

wbruna:
> the more useful max_buffer_size doesn't seem to be accessible from the API 😕).

Actually, there _is_ a way to access it: `ggml_backend_vk_device_supports_op` rejects any tensor larger than the max buffer size. The following attempt:

```diff
diff --git a/src/core/ggml_extend.hpp b/src/core/ggml_extend.hpp
index 4ebbc0a..2c14551 100644
--- a/src/core/ggml_extend.hpp
+++ b/src/core/ggml_extend.hpp
@@ -1922,6 +1922,21 @@ protected:
             // genuine runtime OOM (planned size <= max, but the device is full)
             // is NOT caught here -- it still surfaces from the real reserve
             // below, so the reactive fallback remains the backstop.
+            if (sd_backend_is(runtime_backend, "Vulkan")) {
+                size_t max_size = 0;
+                for (int i = 0; i < ggml_graph_n_nodes(gf); ++i) {
+                    ggml_tensor* op = ggml_graph_node(gf, i);
+                    max_size = std::max(ggml_nbytes(op), max_size);
+                    if (!ggml_backend_supports_op(runtime_backend, op)) {
+                        LOG_DEBUG("%s: untiled compute op size %.2f MB exceeds backend support; deferring to tiling",
+                                  get_desc().c_str(),
+                                  max_size / 1024.0 / 1024.0);
+                        compute_buffer_deferred_to_tiling = true;
+                        return false;
+                    }
+                }
+                LOG_DEBUG("%s: max op size = %.2f MB", get_desc().c_str(), max_size / 1024.0 / 1024.0);
+            } else {
             size_t max_size = ggml_backend_buft_get_max_size(buft);
             if (max_size > 0) {
                 ggml_gallocr* probe = ggml_gallocr_new(buft);
@@ -1937,6 +1952,7 @@ protected:
                     return false;
                 }
             }
+            }
         }
 
         compute_allocr = ggml_gallocr_new(buft);
```

worked perfectly with my card's 4GiB limit. With an SDXL 1024x960 gen, I've got:
> [DEBUG] ggml_extend.hpp:1931 - vae: untiled compute op size 4320.00 MB exceeds backend support; deferring to tiling

while the slightly smaller 960x960 worked without tiling, despite the graph as a whole getting much larger than 4GiB:
> [DEBUG] ggml_extend.hpp:1938 - vae: max op size = 4050.00 MB
[DEBUG] ggml_extend.hpp:1969 - vae compute buffer size: 9031.86 MB(VRAM)

Of course, that would assume that "unsupported" means "too large"; but a truly unsupported operation would end up failing in the same way as before.

RapidMark:
Thanks @wbruna — I went and measured; your `supports_op` approach is the right call, so I've adopted it.

On the weights: the planned compute size scales exactly with output area (416 / 1664 / 3745 / 6657 MB at 256² / 512² / 768² / 1024²) with no constant offset, so the VAE weights aren't actually being counted in it — there was nothing to exclude there. The real difference your approach makes is using the device's true per-buffer limit instead of the suballocation cap, so it only tiles when a single op genuinely won't fit.

`supports_op` matched in every case — it stays untiled where the untiled decode really does fit and succeed, and defers to tiling exactly where an untiled decode actually OOMs. So I've switched the Vulkan path to the per-op check and kept the `get_max_size` probe for the other backends; I also rebuilt and ran the CUDA and ROCm paths to confirm no change there.

I also added a `max_buffer_size=<bytes>` key to `extra_tiling_args`: in auto mode it tiles when an untiled VAE compute buffer would exceed it, so a user can cap VRAM regardless of the device limit (works on any backend). Your `-1`/`0` cases are already covered by the existing flags (`--no-vae-tiling-fallback` disables, default is auto), so this just adds the `> 0` override and leaves `--vae-tiling` untouched — no compatibility break.

pwilkin:
@RapidMark we'll probably do the other way around - you'll add your solution since it's targeted and I'll rework mine without it, just linked in case I had some cases covered which you missed (eg. might want temporal tiling on LTX instead of standard tiling).

wbruna (commented):
Working fine for me on Vulkan.

The total allocation check (with ROCm, since the Vulkan test triggers much sooner) seems less reliable: I've hit a failure by just increasing the resolution until hitting full VRAM:

<details>

```
[INFO ] stable-diffusion.cpp:4512 - sampling completed, taking 131.77s
[INFO ] stable-diffusion.cpp:4524 - generating 1 latent images completed, taking 131.77s
[INFO ] stable-diffusion.cpp:4197 - decoding 1 latents
[DEBUG] model_loader.cpp:987  - loading 138/244 tensors from /opt/sdif/models/VAE/ae_bf16.safetensors
  |##################################################| 138/138 - 472.50MB/s
[INFO ] model_loader.cpp:1224 - loading tensors completed, taking 0.20s (read: 0.00s, memcpy: 0.00s, convert: 0.04s, copy_to_backend: 0.04s)
[DEBUG] model_manager.cpp:218  - model manager prepared params backend buffer ( 94.57 MB, 138 tensors, VRAM)
[DEBUG] ggml_extend.hpp:2061 - vae compute buffer size: 14978.25 MB(VRAM)
[DEBUG] ggml_extend.hpp:61   - ROCm pool[0]: alloc of 1209.60 MiB failed, flushing 935.90 MiB of cached buffers and retrying
[ERROR] ggml_extend.hpp:70   - ROCm error: out of memory
[ERROR] ggml_extend.hpp:70   -   current device: 0, in function alloc at ggml/src/ggml-cuda/ggml-cuda.cu:449
[ERROR] ggml_extend.hpp:70   -   err
ggml/src/ggml-cuda/ggml-cuda.cu:103: ROCm error
[New LWP 2249503]
[New LWP 2249484]
[New LWP 2249483]
[New LWP 2249481]
[New LWP 2249480]
[New LWP 2249468]
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
__syscall_cancel_arch () at ../sysdeps/unix/sysv/linux/x86_64/syscall_cancel.S:56
warning: 56     ../sysdeps/unix/sysv/linux/x86_64/syscall_cancel.S: Arquivo ou diretório inexistente
#0  __syscall_cancel_arch () at ../sysdeps/unix/sysv/linux/x86_64/syscall_cancel.S:56
56      in ../sysdeps/unix/sysv/linux/x86_64/syscall_cancel.S
#1  0x00007f089909b668 in __internal_syscall_cancel (a1=<optimized out>, a2=<optimized out>, a3=<optimized out>, a4=<optimized out>, a5=a5@entry=0, a6=a6@entry=0, nr=61) at ./nptl/cancellation.c:49
warning: 49     ./nptl/cancellation.c: Arquivo ou diretório inexistente
#2  0x00007f089909b6ad in __syscall_cancel (a1=<optimized out>, a2=<optimized out>, a3=<optimized out>, a4=<optimized out>, a5=a5@entry=0, a6=a6@entry=0, nr=61) at ./nptl/cancellation.c:75
75      in ./nptl/cancellation.c
#3  0x00007f08991067c7 in __GI___wait4 (pid=<optimized out>, stat_loc=<optimized out>, options=<optimized out>, usage=<optimized out>) at ../sysdeps/unix/sysv/linux/wait4.c:30
warning: 30     ../sysdeps/unix/sysv/linux/wait4.c: Arquivo ou diretório inexistente
#4  0x000055f37b9c617b in ggml_print_backtrace ()
#5  0x000055f37b9c62ce in ggml_abort ()
#6  0x000055f37af02282 in ggml_cuda_error(char const*, char const*, char const*, int, char const*) ()
#7  0x000055f37af19640 in ggml_cuda_pool_leg::alloc(unsigned long, unsigned long*) ()
#8  0x000055f37af17e41 in ggml_cuda_op_mul_mat_cublas(ggml_backend_cuda_context&, ggml_tensor const*, ggml_tensor const*, ggml_tensor*, char const*, float const*, char const*, float*, long, long, long, long, ihipStream_t*) ()
#9  0x000055f37af16b8a in ggml_cuda_op_mul_mat(ggml_backend_cuda_context&, ggml_tensor const*, ggml_tensor const*, ggml_tensor*, void (*)(ggml_backend_cuda_context&, ggml_tensor const*, ggml_tensor const*, ggml_tensor*, char const*, float const*, char const*, float*, long, long, long, long, ihipStream_t*), void (*)(float const*, int const*, void*, ggml_type, long, long, long, long, long, long, long, long, ihipStream_t*)) ()
#10 0x000055f37af11532 in ggml_cuda_mul_mat(ggml_backend_cuda_context&, ggml_tensor const*, ggml_tensor const*, ggml_tensor*) ()
#11 0x000055f37af0c555 in ggml_cuda_graph_evaluate_and_capture(ggml_backend_cuda_context*, ggml_cgraph*, bool, bool, void const*) ()
#12 0x000055f37af0901c in ggml_backend_cuda_graph_compute(ggml_backend*, ggml_cgraph*) ()
#13 0x000055f37b9dcf0e in ggml_backend_graph_compute ()
#14 0x000055f37ad79dcf in std::optional<sd::Tensor<float> > GGMLRunner::execute_graph<float>(ggml_cgraph*, int, bool, bool, bool, bool, std::unordered_set<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >, std::hash<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::equal_to<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > >, std::allocator<std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > > > const*) ()
#15 0x000055f37ad7c664 in std::optional<sd::Tensor<float> > GGMLRunner::compute<float>(std::function<ggml_cgraph* ()>, int, bool, bool, bool, bool) ()
#16 0x000055f37ad7ca83 in AutoEncoderKL::_compute(int, sd::Tensor<float> const&, bool) ()
#17 0x000055f37ac0fd0a in VAE::decode(int, sd::Tensor<float> const&, sd_tiling_params_t, bool, bool, bool, bool) [clone .isra.0] ()
#18 0x000055f37ac121e9 in StableDiffusionGGML::decode_first_stage(sd::Tensor<float> const&, bool) [clone .isra.0] ()
#19 0x000055f37ac1e14c in generate_image ()
#20 0x000055f37aad7db4 in main ()
```

</details>

But it worked OK when I reduced available VRAM by running something else in parallel:

<details>


```
[INFO ] stable-diffusion.cpp:4524 - generating 1 latent images completed, taking 102.56s
[INFO ] stable-diffusion.cpp:4197 - decoding 1 latents
[DEBUG] model_loader.cpp:987  - loading 138/244 tensors from /opt/sdif/models/VAE/ae_bf16.safetensors
  |##################################################| 138/138 - 472.50MB/s
[INFO ] model_loader.cpp:1224 - loading tensors completed, taking 0.20s (read: 0.01s, memcpy: 0.00s, convert: 0.04s, copy_to_backend: 0.03s)
[DEBUG] model_manager.cpp:218  - model manager prepared params backend buffer ( 94.57 MB, 138 tensors, VRAM)
[ERROR] ggml_extend.hpp:70   - ggml_backend_cuda_buffer_type_alloc_buffer: allocating 12637.90 MiB on device 0: cudaMalloc failed: out of memory
[ERROR] ggml_extend.hpp:70   - ggml_gallocr_reserve_n_impl: failed to allocate ROCm0 buffer of size 13251797248
[ERROR] ggml_extend.hpp:2054 - vae: failed to allocate the compute buffer
[ERROR] ggml_extend.hpp:2504 - vae alloc compute buffer failed
[WARN ] vae.hpp:225  - vae: untiled decode buffer exceeded the backend limit; retrying with tiling
[DEBUG] ggml_extend.hpp:882  - num tiles : 3, 5 
[DEBUG] ggml_extend.hpp:883  - optimal overlap : 0.375000, 0.406250 (targeting 0.340000)
[DEBUG] ggml_extend.hpp:884  - processing 15 tiles
[DEBUG] ggml_extend.hpp:2061 - vae compute buffer size: 1664.25 MB(VRAM)
  |===>                                              | 1/15 - 1.01s/it[DEBUG] ggml_extend.hpp:61   - ggml_backend_cuda_graph_compute: CUDA graph warmup complete
  |==================================================| 15/15 - 1.01it/s
[DEBUG] vae.hpp:263  - computing vae decode graph completed, taking 15.02s
[INFO ] stable-diffusion.cpp:4217 - latent 1 decoded, taking 15.03s
[INFO ] stable-diffusion.cpp:4221 - decode_first_stage completed, taking 15.03s
[INFO ] stable-diffusion.cpp:4661 - generate_image completed in 119.36s
[DEBUG] model_manager.cpp:759  - model manager releasing params backend buffer ( 94.57 MB, 138 tensors, VRAM)
[INFO ] main.cpp:463  - save result image 0 to './test_1781612279.png' (success)
[INFO ] main.cpp:535  - 1/1 images saved
```

</details>

Another attempt that worked:

<details>

```
[INFO ] stable-diffusion.cpp:4524 - generating 1 latent images completed, taking 120.51s
[INFO ] stable-diffusion.cpp:4197 - decoding 1 latents
[DEBUG] model_loader.cpp:987  - loading 138/244 tensors from /opt/sdif/models/VAE/ae_bf16.safetensors
  |##################################################| 138/138 - 472.50MB/s
[INFO ] model_loader.cpp:1224 - loading tensors completed, taking 0.20s (read: 0.00s, memcpy: 0.00s, convert: 0.04s, copy_to_backend: 0.07s)
[DEBUG] model_manager.cpp:218  - model manager prepared params backend buffer ( 94.57 MB, 138 tensors, VRAM)
[DEBUG] ggml_extend.hpp:2061 - vae compute buffer size: 14042.11 MB(VRAM)
[DEBUG] ggml_extend.hpp:61   - ROCm pool[0]: alloc of 1134.00 MiB failed, flushing 877.50 MiB of cached buffers and retrying
[DEBUG] ggml_extend.hpp:61   - ROCm pool[0]: retry succeeded
[DEBUG] vae.hpp:263  - computing vae decode graph completed, taking 7.39s
[INFO ] stable-diffusion.cpp:4217 - latent 1 decoded, taking 7.39s
[INFO ] stable-diffusion.cpp:4221 - decode_first_stage completed, taking 7.39s
[INFO ] stable-diffusion.cpp:4661 - generate_image completed in 129.83s
[DEBUG] model_manager.cpp:759  - model manager releasing params backend buffer ( 94.57 MB, 138 tensors, VRAM)
[INFO ] main.cpp:463  - save result image 0 to './test_1781612278.png' (success)
[INFO ] main.cpp:535  - 1/1 images saved
```

</details>

RapidMark:
@wbruna thanks for the feedback.

I looked into this and discovered that the probe was only checking ghe backends single-buffer size, which CUDA and ROCm report as unlimited, which means it did nothing and we fell back to catching the reserve failure. Also, your csae where the reserve fits, your OOM never hits that, so it aborts.

My fix was to also to check the device free memory and defer the tiling before reserving. 

Running final tests on RDNA2-4+iGPU and will post updated PR soon

wbruna (commented):
Seems to be working fine on ROCm now!

But looking at the code, I think instead of:

```c++
if (sd_backend_is(runtime_backend, "Vulkan")) {
    // supports_op limit, returns false if hit
} else {
    // check (over_buffer_cap || over_free_vram), returns false if hit
}
```

we should do something like this:

```c++
// check over_free_vram, returns false if hit
if (sd_backend_is(runtime_backend, "Vulkan")) {
    // supports_op limit, returns false if hit
} else {
    // check over_buffer_cap, returns false if hit
}
```

because we could hit a free VRAM limit on Vulkan even if each individual op fits the 4G limit.

RapidMark:
Good call — pulled the free-VRAM check out so it applies on Vulkan too, and folded the duplicate probe into one measurement. Verified tiling still kicks in on AMD Vulkan, NVIDIA Vulkan, and ROCm.

examples/common/common.cpp
```diff
@@ -1155,6 +1148,12 @@ ArgOptions SDGenerationParams::get_options() {
          "process vae in tiles to reduce memory usage",
          true,
          &vae_tiling_params.enabled},
```
leejet, line 1150: I think all parameters related to tiling fallback should be placed in `--extra-tiling-args`.

src/core/ggml_extend.hpp
```diff
@@ -1978,10 +1985,77 @@ struct GGMLRunner {
     }
 
     bool alloc_compute_buffer(ggml_cgraph* gf) {
```
leejet, line 1987: I think this mixes allocation, graph memory estimation, and VAE auto-tiling policy in one place.

`alloc_compute_buffer()` should ideally only allocate the compute buffer, or fail if allocation fails. It should not need to know that a failure may be intentionally used by VAE to trigger tiling.

Could we split this into a query-style helper on `GGMLRunner`, for example:

    struct GraphBufferUsage {
        size_t compute_buffer_bytes = 0;
        size_t params_buffer_bytes = 0;
    };

    std::optional<GraphBufferUsage> get_graph_buffer_usage(ggml_cgraph* gf, bool keep_compute_params);

This helper can use the temporary gallocr probe to estimate the compute buffer size, and also report the graph params that need to be prepared for this graph. Then `vae.hpp` can make the VAE-specific decision explicitly:

- if compute buffer exceeds `max_buffer_size`, use tiled VAE
- if compute + params + margin exceeds free VRAM, use tiled VAE
- if the backend cannot support the graph/op size, use tiled VAE
- otherwise call the normal compute path

That would keep `alloc_compute_buffer()` side-effect-free and policy-neutral, and avoid the need for `set_probe_compute_buffer_fits()`, `probe_max_bytes_`, and `compute_buffer_deferred_to_tiling`.

examples/common/common.cpp
```diff
@@ -965,7 +958,7 @@ ArgOptions SDGenerationParams::get_options() {
          &extra_sample_args},
         {"",
          "--extra-tiling-args",
-         "extra VAE tiling args, key=value list. LTX video VAE supports temporal_tile_frames (default: 4), temporal_tile_overlap (default: 1)",
+         "extra VAE tiling args, key=value list. max_buffer_size (bytes) forces the auto fallback to tile when an untiled VAE compute buffer would exceed it. LTX video VAE supports temporal_tile_frames (default: 4), temporal_tile_overlap (default: 1)",
```
leejet, line 961: I don't think `extra_tiling_args=max_buffer_size` is necessary here. We already have `--max-vram` as the user-facing VRAM budget knob, including per-backend/device assignment. Adding a second VAE-specific memory cap in bytes creates overlapping semantics and makes it unclear which memory budget should be authoritative.

For auto VAE tiling, I think the preflight should use the existing runner/backend VRAM budget instead:

- if `--max-vram` is set for this backend, compare the estimated VAE graph memory against that budget;
- otherwise fall back to current free VRAM / backend max-buffer checks;
- always keep the backend hard-limit checks, such as Vulkan `supports_op` or `ggml_backend_buft_get_max_size`.

That would avoid introducing another hidden memory-limit parameter under `extra_tiling_args`.

leejet (changes requested):

PR #1639, opened 2026-06-11
feat: audio-vae-cpu, save/load latent, optimized scripts, LoRA docs

Daniel-Uzcategui:
### `--audio-vae-cpu`
Run audio VAE on CPU to free GPU VRAM for video VAE. Audio VAE is lightweight and runs faster on CPU anyway, so this is a net win for VRAM-constrained setups.

### `--save-latent` / `--load-latent`
Save the denoised latent tensor before VAE decode, or load a pre-computed latent to skip text encoding and denoising. Combined with `--audio-vae-cpu`, enables fast iteration on VAE parameters without re-running denoising.

### Optimized generation scripts
Batch scripts for 480p/720p/768p generation with tuned VAE tiling parameters.

### LoRA documentation
`LORA_DUAL_CHAR.md` — guide for dual-character LoRA usage (two subjects from one LoRA).

### Other changes
- `skip_model_check` parameter on `resolve_and_validate` for latent-loading workflows
- `cast_audio_activation_for_weight` for CPU audio VAE compatibility
- Allow `hires_denoising_strength=0` (skip denoising) for load-latent workflows
- Updated `TEST_MATRIX.md` with complete generation log

🤖 Generated with [Claude Code](https://claude.com/claude-code)


wbruna (changes requested):
Please review [the contribution conventions](https://github.com/leejet/stable-diffusion.cpp/blob/master/CONTRIBUTING.md) for the project, especially the part about AI-assisted code.

leejet:
Thanks for the contribution, but I’m going to close this PR.

The main reasons are:

1. I don’t think stable-diffusion.cpp needs latent import/export support at this stage. This would add a new user-facing file format and maintenance surface, and I don’t want to commit the project to supporting it right now.

2. `--audio-vae-cpu` is not the direction we want to take. This kind of device placement should be handled through the existing / upcoming `--backend` mechanism instead of adding another special-purpose CLI option.

3. The PR contains too many unrelated AI-generated artifacts and experimental files. That makes it hard to review and does not match the level of focus expected for a mergeable PR.

4. More broadly, this PR appears to include substantial AI-generated code, and I’m not confident that the contributor has demonstrated enough understanding of what the generated code is doing or the maintenance implications.

For future contributions, please keep PRs narrowly scoped, avoid committing generated artifacts, and make sure any AI-assisted code is fully understood, tested, and explained by the contributor.

PR #1661, opened 2026-06-15
Switch to npx to avoid global pnpm install

parsecgames:
## Summary

This change uses the recommended npx pnpm to avoid installing pnpm globally (npm i -g pnpm). Tested on Linux.

## Checklist

- [ x] I have read and confirmed this PR follows the [contribution guidelines](https://github.com/leejet/stable-diffusion.cpp/blob/master/CONTRIBUTING.md).


leejet:
Thanks for the PR.

I’d prefer not to change the default frontend build path from `pnpm` to `npx pnpm` for now. The current CMake setup intentionally expects `pnpm` to be installed and available, which keeps the build behavior explicit and predictable.

Users who want to use `npx` can still build the frontend manually with their own workflow and then use the generated `index.html`/header output.

Closing this for now, but thanks for the contribution.

Issue #1431, opened 2026-04-16
[Bug] Vulkan Flash Attention not working


daniandtheweb:
### Git commit

a564fdf642780d1df123f1c413b19961375b8346

### Operating System & Version

Artix Linux - Linux 7.0.0-1-cachyos

### GGML backends

Vulkan

### Command-line arguments used

./sd-cli -M img_gen -p "a cat" --sampling-method euler --steps 20 -W 1024 -H 1024 -b 1 --cfg-scale 5 -s -1 --clip-skip -1 --embd-dir /home/daniandtheweb/Workspace/sd.cpp-webui/models/embeddings/ --lora-model-dir /home/daniandtheweb/Workspace/sd.cpp-webui/models/loras/ -t 0 --rng cuda --sampler-rng cuda --lora-apply-mode auto -o /home/daniandtheweb/Workspace/sd.cpp-webui/outputs/txt2img/1765151582.png --diffusion-model /home/daniandtheweb/Workspace/sd.cpp-webui/models/unet/anima-preview3-base.safetensors --vae /home/daniandtheweb/Workspace/sd.cpp-webui/models/vae/qwen_image_vae.safetensors --llm /home/daniandtheweb/Workspace/sd.cpp-webui/models/text_encoders/qwen_3_06b_base.safetensors --scheduler kl_optimal --vae-tile-overlap 0.5 --vae-tile-size 32x32 --cache-mode spectrum --preview proj --preview-path /home/daniandtheweb/Workspace/sd.cpp-webui/outputs/txt2img/1765151582_preview.png --preview-interval 1 --offload-to-cpu --vae-tiling --fa --vae-conv-direct --mmap --color

### Steps to reproduce

Running Ernie or Anima models on Vulkan with Flash Attention enabled causes the generation to fail.
When Flash Attention is disabled the generation works correctly.

### What you expected to happen

<img width="1024" height="1024" alt="Image" src="https://github.com/user-attachments/assets/bb19f9f5-a81e-4d5f-8842-da2f8ca3b602" />

### What actually happened

<img width="128" height="128" alt="Image" src="https://github.com/user-attachments/assets/e3a47e91-058f-4ba4-bf95-b706792a6ba5" />

### Additional context / environment details

This issue started with 8f2967c006c43c08e72e6748cb8e44993edf0265.

daniandtheweb:
This is probably related to the issue mentioned here: #509.

Naster17:
Same

[...]

Issue #1647, opened 2026-06-13
[Bug] Vulkan GGML Backend Performance Regression on commit 9b0fceb


sz1kormar:
### Git commit

A major performance regression occurs in the Vulkan GGML backend when moving from
commit 19bdfe2 → commit 9b0fceb.
On an AMD RX 580 running Windows 10 22H2 through sd.cpp‑webui, iteration time increases from ~9.5s/it to 22–23s/it, more than doubling generation time on identical workloads.

### Operating System & Version

Windows 10 22H2

### GGML backends

Vulkan

### Command-line arguments used

sd-cli.exe -M img_gen -p "anime, depth of field, 1girl, closed eyes, white shirt, open shirt, cross necklace, nurse cap, pov, 1boy, holding another's arm, <lora:anima-turbo-lora-v0.1:1>" --sampling-method er_sde --steps 8 -W 1152 -H 896 -b 1 --cfg-scale 1 -s 20514 --clip-skip -1 --embd-dir D:\Programs\sd.cpp-webui\models/embeddings/ --lora-model-dir D:\Programs\sd.cpp-webui\models/loras/ -t 0 --rng std_default --sampler-rng std_default --lora-apply-mode auto -o D:\Programs\sd.cpp-webui\outputs\txt2img\158.png --diffusion-model D:\Programs\sd.cpp-webui\models\unet\anima_baseV10.safetensors --vae D:\Programs\sd.cpp-webui\models\vae\qwenimagevae_v7.safetensors --llm D:\Programs\sd.cpp-webui\models\text_encoders\qwen_3_06b_base.safetensors --type f16 --scheduler sgm_uniform --vae-tile-overlap 0.5 --vae-tile-size 64x64 --cache-mode spectrum --vae-tiling --diffusion-conv-direct --vae-conv-direct --mmap --color

### Steps to reproduce

1. Launch sd.cpp‑webui using build 19bdfe2.

2. Run any image generation task using the Vulkan backend.

3. Update to build 9b0fceb.

4. Run the exact same command and settings.

5. Compare iteration times.

### What you expected to happen

Performance should match or exceed previous Vulkan backend throughput.

### What actually happened

BEFORE regression:  
sd-master‑19bdfe2‑bin‑win‑vulkan‑x64

Performance: ~9.5 seconds per iteration

AFTER regression:  
sd-master‑9b0fceb‑bin‑win‑vulkan‑x64

Performance: 22–23 seconds per iteration

[...]

PR #1678, opened 2026-06-17
fix: workaround for Anima with Vulkan and Flash Attention

wbruna:
## Summary

Avoids broken images on Anima with Vulkan + Flash Attention.

## Related Issue / Discussion

#1647 (unrelated) has the original discussion, and some tests; I don't think this was reported on a specific issue.

## Additional Information

> sd-cli --backend vulkan1 -p "1girl, pointing to a blackboard where it is written: FLASH ATTENTION <lora:anima-turbo-lora-v0.1:1>" --sampling-method euler --scheduler sgm_uniform --steps 8 -W 1024 -H 1024 --cfg-scale 1  --vae-tiling --vae-conv-direct  -s 1 --diffusion-model anima-base-v1.0.safetensors --lora-model-dir . --vae Qwen_Image-VAE.safetensors --llm Qwen3-0.6B-UD-Q4_K_XL.gguf --fa

| master (7f0e728b7d42f2490dfa5dd9539082d904f2f6b2) | PR |
|---|---|
| master flash attention | PR flash attention |

Not sure how to choose the best scale; values between 1/8 and 1/128 seemed to work fine on my card (RX 7600 XT, radv, amdgpu).

## Checklist

### DIFF
--- a/src/model/diffusion/anima.hpp
+++ b/src/model/diffusion/anima.hpp
@@ -227,6 +227,7 @@ namespace Anima {
             k4 = k_norm->forward(ctx, k4);
 
             ggml_tensor* attn_out = nullptr;
+            float scale           = (sd_backend_is(ctx->backend, "Vulkan") && ctx->flash_attn_enabled) ? 1.0f / 32.0f : 1.0f;
             if (pe_q != nullptr || pe_k != nullptr) {
                 if (pe_q == nullptr) {
                     pe_q = pe_k;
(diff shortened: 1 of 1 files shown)


daniandtheweb:
This may be related to an issue I've been having for a while, I'll test it later today, hopefully it's the fix for it.

daniandtheweb:
I can confirm that this fixes #1431 for Anima. Maybe some similar fix is also required for Ernie image, my original issue started happening on both models with the same commit.

leejet:

jeffbolznv:
Hi, I think https://github.com/ggml-org/llama.cpp/pull/24909 should avoid the need for this workaround. I get a reasonable/stable image after this change, though it looks quite different than the image in the OP.

wbruna:
> Hi, I think [ggml-org/llama.cpp#24909](https://github.com/ggml-org/llama.cpp/pull/24909) should avoid the need for this workaround. I get a reasonable/stable image after this change, though it looks quite different than the image in the OP.

Works for me 🙂 Applying your fix and reverting mine, I get:



Which is very similar to the same gen without Flash Attention:

Issue #1199, opened 2026-01-17
Read prompt from text file?


fcore117:
Hello, i searched hours and cannot find what is command to read prompt from txt file? i tried --prompt ".\prompt.txt" but it wont work and do not read prompt that is inside txt file. Thanks

PR #1693, opened 2026-06-22
feat: add --prompt-file and --negative-prompt-file flags

grencez:
as an alternative to passing long strings to the --prompt and --negative-prompt flags.

## Summary
What: Provide alternative flags that populate the `prompt` and `negative_prompt` variables from file content.
Why: Putting the prompts in files makes them easier to edit and leak less info on a multi-user system.

Granted, there are other ways to accomplish those goals, but I found this to be the best for my use case. I've been maintaining a patch locally for ~2 months now.

## Related Issue / Discussion
https://github.com/leejet/stable-diffusion.cpp/issues/1199 was filed by someone else near the start of the year, so I'm probably not the only one who wants this feature.

## Additional Information
I tested manually with good cases:
- file exists
- /dev/stdin

I also tested some error cases to make sure it doesn't crash or silently succeed:
- file not found
- no read permission
- broken symlink
- directory instead of file

## Checklist

### DIFF
--- a/examples/common/common.cpp
+++ b/examples/common/common.cpp
@@ -6,6 +6,7 @@
 #include <cstdlib>
 #include <ctime>
 #include <filesystem>
+#include <fstream>
 #include <iomanip>
 #include <iostream>
 #include <regex>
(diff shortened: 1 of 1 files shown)


leejet:

PR #1697, opened 2026-06-22
fix?: enforce flow shift for all schedulers

stduhpf:
## Summary

Applies flow-shift even on sigma-space schedulers.

## Related Issue / Discussion

https://github.com/leejet/stable-diffusion.cpp/pull/1669(discussion)

## Additional Information

Example: bong_tangent  scheduler (sigma-space)

sd-cli.exe --diffusion-model  ..\ComfyUI\models\unet\z_image_turbo-Q8_0.gguf --vae ..\ComfyUI\models\vae\flux\ae.safetensors  --llm ..\ComfyUI\models\clip\llms\Qwen3-4b.Q6_K.gguf --cfg-scale 1.0 -v --offload-to-cpu   --preview proj --steps 8 --rng cpu --lora-model-dir H:\ComfyUI\models\loras\z-img -W 1024 -H 1024 --color --vae-conv-direct  --sampling-method res_2s -p "A lovely cat holding a small 16:9 picture frame" --scheduler bong_tangent  --flow-shift [x]`

| | --flow-shift 3.0 |  --flow-shift 1.0 |
| -- | -- | -- |
| master | output - Copy (360) | output |
| PR | output | output - Copy (361) |

smoothstep scheduler (timestep-space)

| | --flow-shift 3.0 |  --flow-shift 1.0 |
| -- | -- | -- |
| master | output | output |
| PR | output | output |

## Checklist

### DIFF
--- a/src/runtime/denoiser.hpp
+++ b/src/runtime/denoiser.hpp
@@ -571,13 +571,19 @@ struct Denoiser {
     virtual sd::Tensor<float> inverse_noise_scaling(float sigma,
                                                     const sd::Tensor<float>& latent) = 0;
 
+    virtual float apply_shift_warp(float sigma) {
+        return sigma;
+    }
+
     virtual std::vector<float> get_sigmas(uint32_t n, int image_seq_len, scheduler_t scheduler_type, SDVersion version, const char* extra_sample_args = nullptr) {
         auto bound_t_to_sigma = std::bind(&Denoiser::t_to_sigma, this, std::placeholders::_1);
         std::shared_ptr<SigmaScheduler> scheduler;
+        bool shifted_scheduler = false;
         switch (scheduler_type) {
             case DISCRETE_SCHEDULER:
                 LOG_INFO("get_sigmas with discrete scheduler");
                 scheduler = std::make_shared<DiscreteScheduler>();
+                shifted_scheduler = true;
                 break;
             case KARRAS_SCHEDULER:
                 LOG_INFO("get_sigmas with Karras scheduler");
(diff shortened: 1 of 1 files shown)


leejet:
Thanks for looking into this, but I don't think this is the right fix.

The current distinction between scheduler types is intentional: timestep-space schedulers call `t_to_sigma()`, and for flow denoisers hat is where the flow shift is applied. Sigma-space schedulers define their own sigma curve directly from `sigma_min()` / `sigma_max()` or from a fixed eference schedule, so applying an additional element-wise flow warp after schedule generation changes the meaning of those schedulers.

This is also not equivalent to diffusers' FlowMatch scheduler behavior with Karras sigmas. Diffusers applies dynamic shifting before the ptional Karras conversion, then Karras rebuilds the schedule from the shifted endpoints. This PR instead warps every generated sigma after the cheduler has already constructed its curve, which changes the intermediate points in a different way.

So I don't think we should merge this global post-processing approach. If we want diffusers-like behavior for Karras, it should be implemented in the scheduler-specific path by shifting the endpoint range before Karras reconstruction, not by warping all scheduler outputs afterward.

stduhpf:
I get your point and you're probably right. 

But on the other hand, this makes the flow-shift parameter silently innefective with a lot of schedulers, which might feel inconsistent and confusing for users who don't care about the implementation details of the schedulers nor strict compatibility with diffusers.

leejet:
I understand the concern, but I don't think this UX inconsistency is a problem we should solve at this level.

For most users, schedulers are already black-box sampling strategies. `--flow-shift` is not guaranteed to have the same type or strength of effect across all schedulers.

For sigma-space schedulers, the schedule itself is the definition of the sigma curve. Applying an element-wise shift after the scheduler has generated that curve changes the scheduler's meaning.

So I think we should keep the current behavior: timestep-space schedulers apply flow shift through `t_to_sigma()`, while sigma-space schedulers use their own sigma curve, with any shift effect limited to the endpoints when applicable.

Issue #960, opened 2025-11-09
[Bug] Flash-attention gives very bad results on ROCm/Hipblas


stduhpf:
### Git commit

694f0d923578262c4f12ae93ded8e56c116085fe

### Operating System & Version

Windows 10

### GGML backends

HIP

### Command-line arguments used

.\buildhip\bin\sd.exe --diffusion-model  ..\ComfyUI\models\unet\Flux\dev\flux1-dev-iq4_nl.gguf --t5xxl ..\ComfyUI\models\clip\t5\t5xxl_q8_0.gguf  --clip_l ..\ComfyUI\models\clip\clip_l\clip_l.safetensors --vae ..\ComfyUI\models\vae\flux\ae.safetensors  -p "Photograph of the Alps. Vivid summer afternoon. Everything is perfectly in focus." --cfg-scale 1 --sampling-method euler -v --preview proj -W 2048 -H 1024 --vae-tiling --vae-tile-size 64 --steps 20--clip-on-cpu --scheduler smoothstep --diffusion-fa

### Steps to reproduce

Just run a DiT model at high resolution with flashAttention enabled.

### What you expected to happen

A clean image that closely matches the image I can get without flash attention 

### What actually happened

I get results that are either Blurry, overly contrasted or both, sometimes even black images.

Flux.1[Dev]  2048x1024:

| HIP fa ON | HIP fa OFF | Vulkan fa ON|
| --- | --- | --- |
|<img width="2048" height="1024" alt="Image" src="https://github.com/user-attachments/assets/24952586-ce4c-4581-a6e7-74cfddb29ec9" /> | <img width="2048" height="1024" alt="Image" src="https://github.com/user-attachments/assets/2f2b04ae-2f49-4ba0-8995-bdef0232d701" /> | <img width="2048" height="1024" alt="Image" src="https://github.com/user-attachments/assets/f17da4aa-bf36-4b0e-b636-af4dbb17b242" /> |

Flux.1 Krea [Dev] 1536x2048 (too big to run without FA at reasonable speeds on my GPU):

| HIP fa ON | Vulkan fa ON|
| --- | --- |
| <img width="1536" height="2048" alt="Image" src="https://github.com/user-attachments/assets/f9aa64d3-6335-4814-9805-c4da19ab1dc3" /> | <img width="1536" height="2048" alt="Image" src="https://github.com/user-attachments/assets/707f4e53-3131-4bf7-8fbd-642234a196d3" /> |

[...]

PR #1705, opened 2026-06-23
feat: add krea2 support

leejet:
## Summary

- Add krea2 support

## Related Issue / Discussion

N/A

## Additional Information

```
.\bin\Release\sd-cli.exe --diffusion-model krea2_raw-Q8_0.gguf --llm ..\..\ComfyUI\models\text_encoders\Qwen3-VL-4B-Instruct-Q4_K_M.gguf --vae ..\..\ComfyUI\models\vae\wan_2.1_vae.safetensors -p "a lovely cat"  --diffusion-fa -v --offload-to-cpu
```

## Checklist

output


stduhpf:
How are you so fast? I was only halfay through with it

leejet:
In fact, if I hadn’t initially used Qwen3-4B instead of Qwen3-VL-4B, which cost me two hours of debugging, it could have been done faster.

GreenShadows:
But judging by this image, something still seems off, there's a lot of noise. It looks like SD XL.

stduhpf:
@GreenShadows It's because it's running the "raw" model, and only with only 20 steps (recommended ~50).

Here's an example with the "turbo" model:

`.\build\bin\sd-cli.exe --diffusion-model krea2-turbo-Q8_0.gguf --llm Qwen_Qwen3-VL-4B-Instruct-Q5_K_M.gguf --vae ..\ComfyUI\models\vae\wan-qwen\wan_2.1_vae.safetensors --preview proj --cfg-scale 1 --steps 8 --offload-to-cpu --diffusion-fa -p "A close-up portrait of a young East Asian woman with straight black hair, loose strands sweeping across her fair skin, and an intense gaze. She wears a light grey collared shirt with a black tie. A vibrant bouquet of pink and orange lilies with lush green leaves sits in the blurred right foreground. The background is a solid, striking crimson red. Soft, directional studio lighting highlights her facial features, creating a high-contrast composition with a shallow depth of field." -W 1184 -H 1776 --vae-tiling --vae-tile-size 64`

output - Copy (358)

GreenShadows:
Hmm. That looks much better, although there's still some subtle noise compared to what I saw a little while ago in Comfy UI, maybe because you specified "depth of field" in the prompt, or the scheduler isn't optimal.

stduhpf:
For comparison, the official demo image using that same prompt looks like this

flowers

It is noticably sharper with better skin detail, but it's also 4Mpx ,while mine is only 2Mpx... Could definitely be related to the scheduler though.

Looks like the official scheduler is resolution aware (a bit like the one for Ideogram4, see https://github.com/leejet/stable-diffusion.cpp/pull/1669). 
edit: actually for the turbo model, it's not supposed to use the resolution aware thing.

stduhpf:
50 steps: 
output

stduhpf:
@GreenShadows You might be onto something, even with the same prompt and same resolution, the result is much softer and with more saturated colors compared to the official sample image.

| Official | Sd.cpp |
| -- | -- |
| flowers | output |

Edit: might be just that: https://github.com/leejet/stable-diffusion.cpp/issues/960

stduhpf:
I tried running **without** flash attention to see if that was the softness issue, I get black images at "high" (>= ~1 Mpx) resolutions. On ROCm backend. 

Edit : And OOM on Vulkan so i cannot test it with that either

leejet:
I’m not entirely sure whether the official example involves any special processing, such as prompt enhancement. Additionally, the sampling parameters may also have an impact.

stduhpf:
Took forever to run, but I managed to generate a 2Mpx image with CPU backend to compare

| CPU backend no fa | ROCm backend fa | 
| --- | --- |
| output | output - Copy (358) |

(All sampling parameters and seed are the same)

The CPU image looks a lot more detailed and less oversaturated, comparable to the official reference in my opinion.

This definitely looks like I was just experiencing issue #960 again. I never looked back at it but it seems it hasn't resolved itself.

Issue #857, opened 2025-09-25
[Metal]: Regression after 1ba30ce: Unsupported Op with nitro models


newfla:
Up to [master-306-2abe945](https://github.com/leejet/stable-diffusion.cpp/releases/tag/master-306-2abe945) Chroma models won't work when building for Metal backend with the infamous GGML unsupported op abort signal.
Starting from [1ba30ce](https://github.com/leejet/stable-diffusion.cpp/commit/1ba30ce00518543a66b18b05f10908026323d57f) Chroma model started working but nitro/diff* models crash with the GGML unsupported op signal.

The models files that I'm using are described here:
[Chroma](https://github.com/newfla/diffusion-rs/blob/b092e76646c7587b56147b063b46bd59b715bea2/src/preset_builder.rs#L346)
[Nitro](https://github.com/newfla/diffusion-rs/blob/b092e76646c7587b56147b063b46bd59b715bea2/src/preset_builder.rs#L395)

Green-Sky:
There has been another llama.cpp sync to ggml https://github.com/ggml-org/ggml/commit/84fd368c7b7b11c07089d78457c4992bc6f96b6e . Can you try with an updated ggml submodule?

newfla:
Unfortunately the regression is still present 😕

newfla:
Hi, 
Metal backend seems heavily broken: I have found that sxdl-turbo is now broken.

CMD:
`./sd -m /Users/fbiz/.cache/huggingface/hub/models--stabilityai--sdxl-turbo/blobs/e869ac7d6942cb327d68d5ed83a40447aadf20e0c3358d98b2cc9e270db0da26 --vae /Users/fbiz/.cache/huggingface/hub/models--madebyollin--sdxl-vae-fp16-fix/blobs/235745af8d86bf4a4c1b5b4f529868b37019a10f7c0b2e79ad0abca3a22bc6e1 --guidance 0.0 --steps 4 --cfg-scale 1.0 -p "banana"`

Relevant stack trace:

`[INFO ] model.cpp:2293 - loading tensors completed, taking 11.43s (process: 0.01s, read: 11.22s, memcpy: 0.00s, convert: 0.00s, copy_to_backend: 0.08s)
[...]

PR #1731, opened 2026-07-02
fix: enable Wan/TAEHV video generation on the Metal backend (#850)

### LINKED ISSUE #850: unsupported op 'IM2COL_3D' on Mac
I tried to run Wan2.2-TI2V-5B on a Mac Mini, it hits an assert

```
ggml/src/ggml-metal/ggml-metal.m:2068: unsupported op 'IM2COL_3D'
```

is there any way to avoid this problem? I could run Flux dev successfully, by the way

mmandelker-code:
## Summary

Wan 2.1/2.2 video generation aborts on the Metal backend with `unsupported op 'IM2COL_3D'` (#850). There are two separate problems:

1. `ggml_ext_conv_3d` builds `ggml_conv_3d`, which decomposes into `GGML_OP_IM2COL_3D` — not implemented by the Metal backend.
2. The Wan VAE, TAEHV, and LTX audio VAE build left-padding `GGML_OP_PAD` ops, and Metal's PAD kernel only implements right-padding (its `supports_op` requires op params 0/2/4/6 to be zero).

This PR fixes both on the sd.cpp side, without touching `ggml` (per CONTRIBUTING, ggml updates are maintainer-only). Both fixes are gated on `ggml_backend_supports_op`, so graphs on backends that already work are byte-for-byte unchanged:

- `ggml_ext_conv_3d` now takes the runner backend and falls back to `ggml_conv_3d_direct` (`GGML_OP_CONV_3D`, which Metal implements) only when the backend does not support the probed `IM2COL_3D` op. The probe matters: CUDA implements `IM2COL_3D` but **not** `GGML_OP_CONV_3D`, so switching unconditionally would break CUDA video generation.
- `ggml_ext_pad_ext` now takes the runner backend; when the native (left-)pad op is unsupported, it decomposes into a right-pad by `lp+rp` followed by `ggml_roll` by `lp` (`GGML_OP_ROLL` is unconditionally supported on Metal; the `shift < ne` assertion always holds because `ne` grew by `lp+rp`).
- Direct `ggml_pad_ext` calls with left-padding in `wan_vae.hpp`, `tae.hpp`, and `ltx_audio_vae.hpp` are routed through `ggml_ext_pad_ext` and pass the runner backend.

Once the missing Metal kernels land in ggml (see below), the gates automatically stop firing — no cleanup is required to benefit.

## Related Issue / Discussion

- Fixes #850 (also reported for Qwen-Image in #857 — the conv3d/pad situation is the same family of missing Metal ops, though this PR only exercises the video paths).
- ggml-side alternative: ggml-org/llama.cpp#16669 adds the missing Metal kernels (`IM2COL_3D`, PAD left-padding). It is still open/unmerged, and per CONTRIBUTING ggml bumps are maintainer-only, so this PR makes video generation work with the currently vendored ggml. The two are complementary.

## Additional Information

Verified on an M4 Pro (24 GB, macOS 15 / Darwin 25.5), Metal backend:

- Wan 2.2 TI2V-5B (Q8_0), umt5-xxl Q4_K_M: **t2v and i2v** both complete end-to-end (previously: `ggml_abort` with `unsupported op 'IM2COL_3D'` before the first diffusion step).
- Decode verified with both **TAEHV** (`taew2_2.safetensors`) and the **full Wan 2.2 VAE**.
- Example: 33 frames, 480x480, 20 steps, `--diffusion-fa --offload-to-cpu` → ~5 min end-to-end.

Not covered: the two LTX audio VAE left-pad sites received the same mechanical change and compile, but LTX audio was not run (the model does not fit in 24 GB). On backends with native left-pad support the gate keeps the original single PAD op there, so only Metal behavior changes.

## Checklist

### DIFF
--- a/src/core/ggml_extend.hpp
+++ b/src/core/ggml_extend.hpp
@@ -1038,6 +1038,7 @@ __STATIC_INLINE__ ggml_tensor* ggml_ext_linear(ggml_context* ctx,
 }
 
 __STATIC_INLINE__ ggml_tensor* ggml_ext_pad_ext(ggml_context* ctx,
+                                                ggml_backend_t backend,
                                                 ggml_tensor* x,
                                                 int lp0,
                                                 int rp0,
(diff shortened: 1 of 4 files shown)


leejet:

PR #1737, opened 2026-07-03
feat: (fix?) try known prefixes to find correspondance for applying imatrix

stduhpf:
## Summary

Allows to apply imatrix to model parts without creating an all in one checkpoint.

## Checklist


src/model_loader.cpp
```diff
@@ -1165,6 +1165,33 @@ bool ModelLoader::load_tensors(on_new_tensor_cb_t on_new_tensor_cb,
                         }
                         std::string processed_name = convert_tensor_name(tensor_storage.name, imatrix_version);
                         std::vector<float> imatrix = get_imatrix_collector().get_values(processed_name);
+                        if (imatrix.empty()) {
```
leejet, line 1168: I think it would be better to pass in a `--imat-prefix` argument to specify the prefix, rather than trying to hardcode the traversal.

leejet (changes requested):

PR #1743, opened 2026-07-05
feat: add DPM++ 2M SDE (Brownian tree) sampler

fszontagh:
## Summary

Adds `dpm++2m_sde_bt` - the DPM++ 2M SDE variant that k-diffusion / ComfyUI expose as `sample_dpmpp_2m_sde_gpu`. Same math as the `dpm++2m_sde` merged in #1742, but the noise per step comes from a seeded Brownian tree instead of a fresh `randn_like`, giving step-count-stable composition: the same seed at 20 and 40 steps produces the same image, only fine detail differs.

The k-diffusion `_gpu` suffix is legacy naming (torchsde used to require CUDA tensors) - the algorithm is scalar-side and backend-agnostic. Named `_bt` here for honesty, with `k_dpmpp_2m_sde_gpu` / `dpm++ 2m sde gpu` aliased in the sdapi mapper for A1111 / Comfy compatibility.

## Related Issue / Discussion

Follow-up to #1742.

## Additional Information

Verified end to end on SD 1.5 (`v1-5-pruned-emaonly-fp16`), 512x512, karras, seed 42, prompt `"a cat"`. Same seed, only step count varies. Runtime overhead of the Brownian tree is ~1% (SD 1.5 20-step: 9.01s vs 8.93s for the non-BT variant).

### Step-count stability (this is the point)

| | 20 steps | 40 steps |
| --- | --- | --- |
| `dpm++2m_sde_bt` (this PR) | stab_dpm++2m_sde_bt_20 | stab_dpm++2m_sde_bt_40 |
| `dpm++2m_sde` (baseline) | stab_dpm++2m_sde_20 | stab_dpm++2m_sde_40 |

Top row: same tortoiseshell portrait, same eyes, same background - refined detail only. Bottom row: two unrelated cats.

### Design notes

- `BrownianTreeNoiseSampler` is a recursive Brownian bridge with fixed max depth (24 levels). Each interval seed derives from the base seed via splitmix64 + branch salt, so the noise at any point is a pure function of the seed and the position within `[sigma_min, sigma_max]`, independent of which schedule sigmas are queried. This is what gives the step-count stability.
- Base seed is drawn from the sampler RNG at loop entry, so it inherits the user seed and stays deterministic.
- W(sigma) values are memoised across the sampling loop; each schedule point is reused as `sigma_a` on the next step.
- Enum value is appended; existing integer values stay stable.

## Checklist


leejet:

PR #1750, opened 2026-07-05
imatrix: reject a repeated entry with an inconsistent value count (#1749)

### LINKED ISSUE #1749: Heap out-of-bounds write in load_imatrix via a duplicate tensor name
I was fuzzing the imatrix loader and hit a heap out-of-bounds write in `IMatrixCollector::load_imatrix` (`src/runtime/imatrix.cpp`).

`load_imatrix` resizes an entry's `values`/`counts` vectors only when they are empty. Entries are keyed by tensor name in an unordered_map, so an imatrix file with two entries that share the same tensor name reaches the accumulation loop with the vectors sized to the first entry while looping over the second entry's `nval`. When the second `nval` is larger, the loop writes past the end of both heap buffers:

```cpp
if (e.values.empty()) {
    e.values.resize(nval, 0);
    e.counts.resize(nval, 0);
}
...
for (int i = 0; i < nval; i++) {
    e.values[i] += tmp[i];   // OOB write when nval > e.values.size()
    e.counts[i] += ncall;    // OOB write
}
```

The written value, the count, and the out-of-bounds length are all read from the file. A crafted `.imatrix` with `n_entries=2`, both named `blk.0`, the first with `nval=1` and the second with a large `nval`, corrupts the heap and crashes sd-cli (the imatrix loads via `--imat-in` before any model). Confirmed on current master (`dff0e88`).

The nice thing is that `collect_imatrix` in the same file already does the right thing: it has an `else if (e.values.size() != ...)` check that `load_imatrix` is missing. So the fix is just to mirror that check. I will open a small PR for it.

I have the crafted `.imatrix` PoC and can share it if useful.

professor-moody:
Fixes #1749.

`load_imatrix` resizes an entry's `values`/`counts` vectors only when they are empty, so a repeated tensor name with a larger `nval` writes past the end of both buffers in the accumulation loop. This mirrors the check `collect_imatrix` already does in the same file: reject a repeated entry whose value count does not match, instead of writing out of bounds. Valid imatrix files are unaffected; a crafted file with a mismatched repeated entry is cleanly rejected with an error.

### DIFF
--- a/src/runtime/imatrix.cpp
+++ b/src/runtime/imatrix.cpp
@@ -287,6 +287,10 @@ bool IMatrixCollector::load_imatrix(const char* fname) {
         if (e.values.empty()) {
             e.values.resize(nval, 0);
             e.counts.resize(nval, 0);
+        } else if (e.values.size() != (size_t)nval) {
+            LOG_ERROR("inconsistent size for a repeated entry (%d vs %d)\n", (int)e.values.size(), nval);
+            stats_ = {};
+            return false;
         }
 
         std::vector<float> tmp(nval);


leejet:

PR #1751, opened 2026-07-05
docs: explain CPU streaming combo in performance.md

fszontagh:
## Summary

Adds a section to `docs/performance.md` explaining how `--offload-to-cpu`, `--max-vram`, and `--stream-layers` compose. These flags exist to let a model that doesn't fit in VRAM still run, but the exact interaction between them isn't spelled out anywhere in one place - and `--stream-layers` silently no-ops when the diffusion params backend isn't CPU (`stable-diffusion.cpp:785`), which is easy to miss.

The new section:

- Describes what each flag does on its own.
- Documents the `--max-vram -1` "auto-detect, spare 1 GiB" semantic (currently only in the arg parser help string).
- Notes the `--stream-layers` requirement (CPU params backend).
- Gives the composed recipe for "run the biggest model this card can host".
- Orders the flags from fastest to smallest-VRAM so users pick the right step for their situation.

## Related Issue / Discussion

None.

## Additional Information

Docs-only. No code, no tests. Flag semantics unchanged.

## Checklist

### DIFF
--- a/docs/performance.md
+++ b/docs/performance.md
@@ -53,6 +53,26 @@ Per-module assignments can target only the largest modules:
 
 See [backend selection](./backend.md) for full syntax.
 
+## Run models that don't fit in VRAM (CPU streaming).
+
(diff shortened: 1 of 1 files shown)


wbruna:
It may be also worth mentioning that, with most model files, `--mmap` can be combined with `--offload-to-cpu` to reduce RAM pressure.

PR #1768, opened 2026-07-10
feat: hot-reload ControlNet - swap without rebuilding the context

fszontagh:
## What

Adds three C API entry points so a long-lived process can swap the ControlNet weights on an existing `sd_ctx_t` without tearing it down:

```c
SD_API bool sd_ctx_load_control_net(sd_ctx_t*, const char* path);
SD_API bool sd_ctx_unload_control_net(sd_ctx_t*);
SD_API bool sd_ctx_has_control_net(const sd_ctx_t*);
```

`sd_ctx_load_control_net` acts as both first-load and replace. Reuses the same construction path as `--control-net`, so hot-loaded ControlNets are indistinguishable at runtime.

## Why

Loading the diffusion model and text encoders takes tens of seconds. Switching ControlNet (canny to depth, etc.) shouldn't require rebuilding the ctx.

## How

- `StableDiffusionGGML` gains `load_control_net_from_file(path)` and `unload_control_net()` sharing the existing construction code.
- `ModelManager` gains `unregister_param_tensors(desc)` so re-registering the same tensor names does not collide.
- `diffusion_conv_direct` and `control_net_params_mem_size` promoted from locals to members so the reload path honors the ctx's original settings.

The helpers are shaped so the same pattern generalizes to other module families (VAE, LoRA, upscaler) later; only ControlNet is exposed here.

## Caveats

- Not thread-safe against generation in flight. The caller serializes, same as the rest of `sd_ctx_*`.
- The ControlNet must be compatible with the base model. Version detection is delegated to the existing `ControlNet` ctor.

## Verification

- `sd-cli --control-net ...` init path unchanged.
- Standalone smoke test cycles load / swap / unload / reload / free cleanly.

### DIFF
--- a/include/stable-diffusion.h
+++ b/include/stable-diffusion.h
@@ -428,18 +428,9 @@ SD_API const char* sd_get_system_info();
 SD_API bool sd_ctx_supports_image_generation(const sd_ctx_t* sd_ctx);
 SD_API bool sd_ctx_supports_video_generation(const sd_ctx_t* sd_ctx);
 
-// Hot-swap the ControlNet weights on an already-initialized context.
-// Loads the file at `path`, replaces any currently loaded ControlNet, and re-
-// registers it with the model manager. The file must be compatible with the
-// diffusion model the context was created with. Safe to call only when no
-// generation is in flight. Returns true on success.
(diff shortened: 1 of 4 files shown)


leejet:

PR #1771, opened 2026-07-10
fix: avoid f16 overflow in Z-Image quantized matmuls on ROCm

pwilkin:
Fixes white image generation on ROCm backend by forcing 32-bit precision.

I don't know how many other image engines are affected, but might be worth a look.


leejet:

PR #1775, opened 2026-07-10
feat: Add Krea2OstrisEdit support

stduhpf:
## Summary

Adds support for Krea2 fine-tunes with editing capabilities trained with https://github.com/ostris/ai-toolkit 

Ref: [Krea2OstrisEdit](https://huggingface.co/ostris/Krea2OstrisEdit)

## Related Issue / Discussion

N/A 

### Example:

reference image: 
VG

prompt: "`a white yeti with horns reading a book that is titled "Ostris + Krea2 Style Reference"`"

style_reference LoRA: https://huggingface.co/ostris/krea2_turbo_style_reference
(also filter bypasss lora applied to the weights directly)

seed: 9499

| master (no image conditioning) | No LoRA (with mmproj) | No LoRA (without mmproj) | style_reference LoRA (with mmproj) | style_reference LoRA (without mmproj) |
|output - Copy (464) | output | output | output - Copy (467) | output - Copy (468) |

## Checklist

### DIFF
--- a/src/conditioning/conditioner.hpp
+++ b/src/conditioning/conditioner.hpp
@@ -2142,7 +2142,7 @@ struct LLMEmbedder : public Conditioner {
 
             prompt = "<|im_start|>system\nDescribe the image by detailing the color, shape, size, texture, quantity, text, spatial relationships of the objects and background:<|im_end|>\n<|im_start|>user\n";
             if (llm->enable_vision && conditioner_params.ref_images != nullptr && !conditioner_params.ref_images->empty()) {
-                std::string img_prompt = "";
+                std::string img_prompt        = "";
                 const std::string placeholder = "<|image_pad|>";
 
                 for (int i = 0; i < conditioner_params.ref_images->size(); i++) {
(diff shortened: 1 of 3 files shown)


stduhpf:
Bonus: llm-vision only (disabled reference latents for the DiT):
| no LoRA | style_reference LoRA |
| -- | -- |
| output | output - Copy (470) |

Maybe this could be an option? (what would be the best UI for that?)

stduhpf:
... and now I find this one, which seems to use different setup:
model: https://huggingface.co/conradlocke/krea2-identity-edit
implentation: https://github.com/lbouaraba/comfyui-krea2edit


- refs are concatenated before the target image rather than after, but since non-causal attention permutation invarient, I'm pretty sure this doesn't matter as long as the RoPE ids are the same
- Uses a slightly different prompt format (probably not that important)
- Does not force timestep=0 for the refs

Just disabling the separate timestep for the refs seems to be enough to make it work pretty well

leejet:

PR #1780, opened 2026-07-11
refactor/feat: Override edit models params + support more edit models

stduhpf:
## Summary

- Unifies all the DiT-based edit models configurations under the new EditModeParams struct. 
- adds the `--ref-image-mode` cli argument to override the configuration
- Enables support for https://github.com/lbouaraba/comfyui-krea2edit / https://huggingface.co/conradlocke/krea2-identity-edit (with the `--ref-image-mode "preset=krea2_edit"` argument), and https://github.com/Mirumo0u0/ComfyUI-Cosmos-Reference/blob/main/cosmos_temporal_reference.py

## Related Issue / Discussion

cont: https://github.com/leejet/stable-diffusion.cpp/pull/1775

implements https://github.com/leejet/stable-diffusion.cpp/issues/1783 and https://github.com/leejet/stable-diffusion.cpp/issues/1787

## Additional Information

The  `--ref-image-mode` uses the same csv format as the `--extra-xxx-args` arguments.

Here are the supported fields:

| Key | Allowed Values | Description |
| :--- | :--- | :--- |
| **`preset`** | `flux_kontext`, `flux2`, `qwen_layered`, `qwen`, `z_image_omni`, `krea2_ostris_edit`,  `krea2_edit`, `cosmos_reference`, `default` | Sets a predefined group of parameters. If unset, the preset is chosen based on the model architectiure to match previous behavior. |
| **`use_vlm`** | `true`, `false` | Determines if references are passed to the Vision Language Model encoder (if applicable). |
| **`pass_to_dit`** | `true`, `false` | Determines if VAE-encoded references are passed to the DiT. |
| **`ref_index_mode`** | `fixed`, `increase`, `decrease` | Sets the behavior of the reference images RoPE index. |
| **`force_timestep_0`** | `true`, `false` | Forces timestep = 0 for the reference  image tokens during diffusion (Note: Only supported by Krea2 architecture). |
| **`resize_vae_refs`** | `true`, `false` | Enables/disables resizing of VAE references. |
| **`vae_refs_max_size`** | `[integer]` | Sets the maximum size for VAE references. |
| **`cond_refs_resize_mode`** | `longest_side`, `area`, `none` | Determines how condition references are resized. |
| **`cond_refs_max_size`** | `[integer]` | Sets the maximum size for condition references. |
| **`cond_refs_min_size`** | `[integer]` | Sets the minimum size for condition references. |
| **`cond_refs_size`** | `[integer]` | **Shortcut:** Sets both `cond_refs_min_size` and `cond_refs_max_size` to this same integer. |

### newly supported models

| Base Model | LoRa | prompt | input image |output image |
| -- | -- | -- | -- | -- |
| Krea2-Turbo | https://huggingface.co/conradlocke/krea2-identity-edit | "Remove the text"| example | output |
| anima-base-v1.0 | https://civitai.com/models/2752978/extend-image-image-edit-anima-ed | "extend image, " | padded-2 | image |
| anima-base-v1.0 | https://civitai.com/models/2650553/anima-edit?modelVersionId=3089149 | "Change the background to be outside with a few trees and grass, it's very vibrant. Make the character hold up a sign that says \`"AnimaEditV2, Now in SD.CPP!\`". The sign should be covering her chest and navel. Also, change her expression to be excited and remove the blush. Keep the clothing the exact same." | padded-3 | output - Copy (486) |

## Checklist


MrDrMcCoy:
Would you be so kind as to include an update to the docs that covers all this?

leejet:

stduhpf:
@leejet I've just realized that `--ref-image-mode force_timestep_0=true` is pretty much synonymous to  `--model-args qwen_image_zero_cond_t=true` (excapt that the first one is for krea2 arch, while the other one is for Qwen-Image arch). Those two should probably be unified (and maybe qwen_image_zero_cond_t should be packaged in a new preset for Qwen Image Edit 2511?). thoughts?

PR #1802, opened 2026-07-21
Update ROCm releases to 7.14.0

superm1:
## Summary
7.14.0 is a production release from AMD.  Drop the previous stable artifacts and only create production artifacts.

Increase hardware support to more devices included with 7.14.0.

## Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -449,8 +449,8 @@ jobs:
     runs-on: windows-2022
 
     env:
-      ROCM_VERSION: "7.13.0"
-      GPU_TARGETS: "gfx906;gfx908;gfx90a;gfx942;gfx950;gfx1010;gfx1011;gfx1012;gfx1030;gfx1031;gfx1032;gfx1033;gfx1034;gfx1035;gfx1036;gfx1100;gfx1101;gfx1102;gfx1150;gfx1151;gfx1152;gfx1200;gfx1201"
+      ROCM_VERSION: "7.14.0"
+      GPU_TARGETS: "gfx1010;gfx1011;gfx1012;gfx1030;gfx1031;gfx1032;gfx1033;gfx1034;gfx1035;gfx1036;gfx1100;gfx1101;gfx1102;gfx1103;gfx1150;gfx1151;gfx1152;gfx1153;gfx1200;gfx1201"
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


kenvandine:
The `windows-latest-rocm` job's "Pack artifacts" step used to explicitly copy `rocblas*.dll` / `hipblas*.dll` and the `rocblas\library` / `hipblaslt\library` kernel directories from `$HIP_PATH\bin` into `build\bin` before zipping. That copy block was removed here — the step now just runs `7z a ... .\build\bin\*` directly.

I don't see anything else that would put those files into `build\bin` on Windows: `cmake --build .` doesn't copy runtime DLL dependencies, and there's no post-build copy/install step in either `stable-diffusion.cpp`'s `CMakeLists.txt` or `ggml`'s `ggml-hip/CMakeLists.txt` (checked both at this PR's head commit / pinned ggml submodule).

Since `HIP_PATH` now points into the pip-installed `rocm-sdk` wheel tree instead of the old tarball's `C:\TheRock\build`, the released `sd-*-bin-win-rocm-7.14.0-x64.zip` may end up missing `rocblas.dll`/`hipblas*.dll` and rocBLAS's adjacent `library` folder (Tensile kernels), which would break `sd.exe` for anyone running it without ROCm already installed. Worth double-checking the packaged zip actually contains those files before this merges.

superm1:
> The `windows-latest-rocm` job's "Pack artifacts" step used to explicitly copy `rocblas*.dll` / `hipblas*.dll` and the `rocblas\library` / `hipblaslt\library` kernel directories from `$HIP_PATH\bin` into `build\bin` before zipping. That copy block was removed here — the step now just runs `7z a ... .\build\bin\*` directly.
> 
> I don't see anything else that would put those files into `build\bin` on Windows: `cmake --build .` doesn't copy runtime DLL dependencies, and there's no post-build copy/install step in either `stable-diffusion.cpp`'s `CMakeLists.txt` or `ggml`'s `ggml-hip/CMakeLists.txt` (checked both at this PR's head commit / pinned ggml submodule).
> 
> Since `HIP_PATH` now points into the pip-installed `rocm-sdk` wheel tree instead of the old tarball's `C:\TheRock\build`, the released `sd-*-bin-win-rocm-7.14.0-x64.zip` may end up missing `rocblas.dll`/`hipblas*.dll` and rocBLAS's adjacent `library` folder (Tensile kernels), which would break `sd.exe` for anyone running it without ROCm already installed. Worth double-checking the packaged zip actually contains those files before this merges.

That was an intentional change because of how multiarch and rocm-sdk works.   When a user installs the rocm-sdk they should pull down these DLL's.  This also mirrors how another ggml project's ROCM 7.14 PR (llama.cpp) works.

superm1:
@ramkrishna2910 can you please validate I got this right on Windows?

JohnLoveJoy:
Gfx900 and Gfx90c are now supported by ROCm as well.

.github/workflows/build.yml
```diff
@@ -472,34 +472,68 @@ jobs:
...
+          New-Item -Path "C:\TheRock\build" -ItemType Directory -Force | Out-Null
+          python -m venv C:\TheRock\build\.venv
+          & C:\TheRock\build\.venv\Scripts\Activate.ps1
+
+          write-host "Upgrading pip"
+          python -m pip install --upgrade pip
+
+          write-host "Installing ROCm wheels for multi-arch support"
+          # Install ROCm wheels for multi-arch support (this may take several minutes)
+          python -m pip install --index-url https://repo.amd.com/rocm/whl-multi-arch/ "rocm[libraries,devel]"
```
leejet: The installed ROCm version is not pinned.

`ROCM_VERSION` is currently only used by the cache key and artifact name, while
`pip install "rocm[libraries,devel]"` installs the latest version available from
the stable index. After a newer ROCm release is published, a cache miss could
produce binaries built against that newer version but still publish them as
`rocm-7.14.0`.

Please pin the package explicitly on both platforms, for example:

    "rocm[libraries,devel]==${ROCM_VERSION}"

.github/workflows/build.yml
```diff
@@ -449,7 +449,7 @@ jobs:
     runs-on: windows-2022
 
     env:
-      ROCM_VERSION: "7.13.0"
+      ROCM_VERSION: "7.14.0"
       GPU_TARGETS: "gfx906;gfx908;gfx90a;gfx942;gfx950;gfx1010;gfx1011;gfx1012;gfx1030;gfx1031;gfx1032;gfx1033;gfx1034;gfx1035;gfx1036;gfx1100;gfx1101;gfx1102;gfx1150;gfx1151;gfx1152;gfx1200;gfx1201"
```
leejet: ROCm 7.14 adds support for multiple gfx1153 Ryzen APUs, and the official multi-arch packages include a device-gfx1153 component.

However, gfx1153 is currently missing from both the Windows and Linux GPU_TARGETS lists.

Since this PR is intended to increase hardware support for devices introduced with ROCm 7.14, please add gfx1153 to both target lists. Otherwise the stable-diffusion.cpp HIP kernels in the released binary will not include a gfx1153 code object.

leejet (changes requested):

superm1:
> Gfx900 and Gfx90c are now supported by ROCm as well.

Yes; but only for Linux.  I've adjusted the jobs for only enabling GFX9 on Linux.

> The installed ROCm version is not pinned.

Now pinned.

> However, gfx1153 is currently missing from both the Windows and Linux GPU_TARGETS lists.

Added.

PR #1810, opened 2026-07-23
feat: optional cooperative cancellation for in-flight image generation

robertphyatt:
## Summary

Adds an optional cooperative cancellation primitive so a host application can abandon an **in-flight `generate_image`** call and get an empty result back promptly, instead of waiting for the full denoise plus VAE decode to finish.

Motivation: in an interactive host, a long generation sometimes needs to be dropped the moment the user does something else. Today there is no way to interrupt a generation once it is running — only to let it finish and discard the result.

## API

Three new functions in `stable-diffusion.h`:

```c
SD_API void sd_request_cancel(void);  // ask the running generation to abandon at its next step
SD_API void sd_reset_cancel(void);    // clear the flag; call before starting a generation
SD_API bool sd_is_cancelled(void);    // read the flag
```

Existing behaviour is unchanged for anyone who never calls `sd_request_cancel()`.

## How it works

- A file-scope `std::atomic<bool>` flag.
- `sample_k_diffusion` polls it once at the top of each step loop. On cancel the loop returns `false` — the same clean-abort path each sampler already takes when `denoised == nullptr` — which unwinds to a `NULL` result naturally.
- Two additional short-circuits in `generate_image_internal` (after text encoding, and before the VAE decode) catch a cancel that lands between the per-step checks, freeing `work_ctx` on the way out to match the function's existing `NULL`-return cleanup.
- A cancelled generation therefore yields an empty result rather than an error, so the two sampling-failure `LOG_ERROR` calls are suppressed while a cancel is in effect.

This is deliberately host-side C++ rather than a ggml abort callback, so it behaves identically across CPU, CUDA, Vulkan and Metal builds.

## Notes and open questions

- The flag is **process-wide**, which is simple and correct for a caller that runs one generation at a time. It does not target a specific `sd_ctx_t`.
- If you would prefer a **per-`sd_ctx` cancel** so concurrent contexts can be cancelled independently, I am happy to rework it that way, with tests in whatever style you prefer. I did not want to build the larger variant speculatively — let me know your preference and I will adjust before you spend review time on this shape.
- `generate_video` is untouched here; the same primitive extends to it easily if that is wanted.
- The branch is based on `545fac4` rather than current `master`, since that is the commit this was built and exercised against. Happy to rebase.

## Testing

Exercised against real FLUX.2-Klein generations from a host application on macOS/Metal:

- a normal generation still returns a full image;
- a cancel requested mid-generation (at step 6 of 28) returns an empty result promptly with no crash;
- a cancel flag set before a generation short-circuits it in ~380 ms against a ~10.4 s uncancelled baseline.

I have not added tests to this repo — glad to add some in your preferred style if you would like them before merging.

### DIFF
--- a/include/stable-diffusion.h
+++ b/include/stable-diffusion.h
@@ -343,6 +343,15 @@ typedef void (*sd_preview_cb_t)(int step, int frame_count, sd_image_t* frames, b
 
 SD_API void sd_set_log_callback(sd_log_cb_t sd_log_cb, void* data);
 SD_API void sd_set_progress_callback(sd_progress_cb_t cb, void* data);
+
+// Optional cooperative cancellation of an in-flight generation. Backend-agnostic:
+// the sampler loop polls sd_is_cancelled() once per step and abandons cleanly, so a
+// cancelled generation returns an empty result rather than an error. Call
+// sd_reset_cancel() before starting a generation; sd_request_cancel() is safe to call
+// from another thread while one is running.
+SD_API void sd_request_cancel(void);
+SD_API void sd_reset_cancel(void);
+SD_API bool sd_is_cancelled(void);
 SD_API void sd_set_preview_callback(sd_preview_cb_t cb, enum preview_t mode, int interval, bool denoised, bool noisy, void* data);
 SD_API int32_t sd_get_num_physical_cores();
 SD_API const char* sd_get_system_info();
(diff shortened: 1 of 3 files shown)


robertphyatt:
Withdrawing this — I hadn't realised `sd_cancel_generation()` (#1124) already landed; my branch was based on an older commit that predated it. Upstream's per-`sd_ctx` design with the three cancel modes is a better fit than the process-wide flag proposed here, so there's nothing worth keeping from this PR. Apologies for the noise, and thanks for the existing implementation.

PR #1838, opened 2026-07-30
feat: add text context CLI parameter

happyyzy:
## Summary

Adds `--text-ctx` to limit text and cross-attention context before graph execution.

This bounds attention shapes and prevents oversized attention buffers from corrupting image quality. Related Adreno OpenCL backend work: https://github.com/ggml-org/llama.cpp/pull/26331

## Verification

`cmake --build build-android-opencl-current-repro --target sd-cli -j 8`

### DIFF
--- a/examples/common/common.cpp
+++ b/examples/common/common.cpp
@@ -1045,6 +1045,10 @@ ArgOptions SDGenerationParams::get_options() {
          "ignore last layers of CLIP network; 1 ignores none, 2 ignores one layer (default: -1). "
          "<= 0 represents unspecified, will be 1 for SD1.x, 2 for SD2.x",
          &clip_skip},
+        {"",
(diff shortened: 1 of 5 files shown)


leejet:
Thanks for the PR.

I don't think we should add a `--text-ctx` limitation in the application layer to work around a backend limitation.

The context length should be handled properly by the backend. Adding an application-level parameter to truncate/limit the input introduces another user-facing constraint and makes the behavior dependent on backend implementation details.

If the issue is caused by oversized attention buffers on a specific backend (e.g. OpenCL/Adreno), the fix should be implemented in that backend instead, so that all users can benefit from the correct behavior without needing to manually adjust model inputs.

For this reason, I don't think this is the right approach to merge.

PR #1861, opened 2026-08-08
Spare 1-D norm weights from a blanket --type

danielhanchen:
A blanket \`--type\` currently quantizes 1-D weights whenever their length happens to divide the block size. A 1-D weight is a per-channel gain (LayerNorm/RMSNorm), never a matmul weight, so every channel of a block shares one scale and one min while a gain vector has no reason to be locally smooth.

Until now these survived only by accident: FLUX q_norm/k_norm are \`[128]\` and \`128 % 256 != 0\`, so the existing block-size check spared them. MiniMax-H3 norms are \`[5376]\` and \`5376 % 256 == 0\`, so a blanket \`--type q4_K\` crushed 106 norm scales to 4 bits.

The failure is quiet, which is what makes it worth guarding. The model still loads and still renders a plausible-looking video, so a "does it run" check passes. Measured against a bf16 render of the same prompt and seed:

| build | PSNR | SSIM | LPIPS |
|---|---|---|---|
| stock, blanket q4_K | 9.87 | 0.074 | 0.981 |
| with this rule | 22.22 | 0.841 | 0.292 |

An LPIPS of 0.981 is essentially uncorrelated with the reference.

The rule sits beside the existing \`.bias\` and \`.scale\` exclusions and costs almost nothing in size: on H3 the 106 spared tensors are 5376 floats each.

### DIFF
--- a/src/model_loader.cpp
+++ b/src/model_loader.cpp
@@ -1511,6 +1511,18 @@ bool ModelLoader::tensor_should_be_converted(const TensorStorage& tensor_storage
     if (type != GGML_TYPE_COUNT) {
         if (ggml_is_quantized(type) && tensor_storage.ne[0] % ggml_blck_size(type) != 0) {
             // Pass, do not convert
+        } else if (ggml_is_quantized(type) && tensor_storage.n_dims <= 1) {
+            // Pass, do not convert. A 1-D weight is a per-channel scale (LayerNorm/RMSNorm gain),
+            // never a matmul weight, so quantizing it buys almost nothing and costs a lot: every
+            // channel of the block shares one scale and one min, and a gain vector has no reason
+            // to be locally smooth. Until now these survived only by accident, when their length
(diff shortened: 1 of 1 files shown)


leejet:
Thanks for the PR. But I don't think we should hard-code an exemption for 1-D tensors here.

If a user explicitly specifies `--type q4_K`, I think we should preserve their ability to quantize any tensor that is technically quantizable. Whether certain tensors, such as norm weights, should remain in BF16 is a quantization-policy decision rather than a format constraint.

We already have `--tensor-type-rules` for this purpose. If some tensors are known to be sensitive to quantization, users or model-specific recipes can explicitly keep them in BF16 through those rules.

Adding a blanket `n_dims <= 1` check also means an explicit `--tensor-type-rules` request to quantize such a tensor can be silently overridden by the loader, which makes the override mechanism less predictable.

In my opinion, the converter should enforce technical constraints on quantization, while leaving quality-related choices to `--type` and `--tensor-type-rules`.

PR #1862, opened 2026-08-08
Stop MiniMax-H3 aborting on the default cfg-scale and on --vae-on-cpu

danielhanchen:
Two deterministic aborts on MiniMax-H3, both of which a user hits on a first run.

## 1. The default cfg-scale aborts

\`--cfg-scale\` defaults to 7.0, and H3 aborts at any value above 1.0:

```
ggml/src/ggml.c: GGML_ASSERT(!ggml_is_transposed(a)) failed
```

SIGABRT, exit 134. So a \`vid_gen\` command that simply omits the flag crashes instead of rendering. H3 is distilled with guidance baked in and has no negative-prompt semantics; its empty uncond prompt encodes to zero tokens, which is where the assert comes from. Measured: cfg 1.0 renders, cfg 1.5 and cfg 4.0 both abort.

Since no cfg above 1.0 has a legitimate use on this model, this clamps to 1.0 with a warning rather than failing.

## 2. --vae-on-cpu aborts

```
ggml/src/ggml-cpu/ops.cpp:6321: GGML_ASSERT(src0->type == GGML_TYPE_F16) failed
```

The trigger is the **audio** VAE, not the video VAE. \`ggml_conv_1d\` and \`ggml_conv_1d_dw\` build an F16 im2col, and the CPU backend then asserts the kernel is F16, while the audio VAE conv kernels reach it as F32. Bisected on the flags: \`--vae-on-cpu\` with \`--audio-vae\` aborts, the same command without \`--audio-vae\` renders. Converting the audio VAE checkpoint to fp16 does not help, because the requirement is imposed inside the graph rather than by the file.

This casts F32 conv1d kernels to F16 in-graph, and only on CPU backends, so GPU backends keep their F32 kernels and no checkpoint is degraded on disk.

Verified end to end on a CUDA build with a q4_K H3 denoiser, the Qwen3-VL encoder and both VAEs: cfg 7 now warns and renders; \`--vae-on-cpu\` renders, alone and combined with \`--offload-to-cpu --clip-on-cpu\`.

### DIFF
--- a/src/stable-diffusion.cpp
+++ b/src/stable-diffusion.cpp
@@ -4222,6 +4222,15 @@ struct GenerationRequest {
             guidance->img_cfg = 1.f;
         }
 
+        // MiniMax-H3 is distilled with guidance baked in and has no negative
+        // prompt semantics; its empty uncond prompt cannot even be encoded
+        // (zero tokens), so any cfg != 1 would abort deep inside ggml.
+        if (sd_version_is_minimax_h3(sd_ctx->sd->version) && guidance->txt_cfg != 1.f) {
+            LOG_WARN("MiniMax-H3 is a distilled, CFG-free model; forcing cfg-scale from %.2f to 1.0",
+                     guidance->txt_cfg);
+            guidance->txt_cfg = 1.f;
+        }
+
         if (guidance->img_cfg != guidance->txt_cfg) {
             *use_uncond = true;
         }
--- a/src/model/vae/ltx_audio_vae.hpp
+++ b/src/model/vae/ltx_audio_vae.hpp
@@ -182,6 +182,17 @@ namespace LTXV {
         }
     };
 
+    // ggml_conv_1d / ggml_conv_1d_dw build an F16 im2col, and the CPU backend
+    // additionally requires the kernel itself to be F16. F32 kernels (as loaded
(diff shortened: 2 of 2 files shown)


stduhpf:
> Since no cfg above 1.0 has a legitimate use on this model, this clamps to 1.0 with a warning rather than failing.

That's quite an assumption. We could get un-distilled models in the future (like we got for Flux in the past), it would be best to find a way to make CFG work without crashing rather than overriding the users settings. (Having a default guidance settings for each model based on architecture could be a nicer solution).

Also those are two unrelated changes that would probably make more sense separated in two PRs.

src/stable-diffusion.cpp
```diff
@@ -4222,6 +4222,15 @@ struct GenerationRequest {
             guidance->img_cfg = 1.f;
         }
 
+        // MiniMax-H3 is distilled with guidance baked in and has no negative
+        // prompt semantics; its empty uncond prompt cannot even be encoded
+        // (zero tokens), so any cfg != 1 would abort deep inside ggml.
+        if (sd_version_is_minimax_h3(sd_ctx->sd->version) && guidance->txt_cfg != 1.f) {
```
leejet, line 4228: Please remove this logic, as there may be non-distilled MiniMax weights or LoRAs in the future.

leejet (changes requested):

PR #1874, opened 2026-08-12
feat: (taehv) support taeh3

stduhpf:
## Summary

Support for https://github.com/madebyollin/taehv/blob/main/safetensors/taeh3.safetensors

## Additional Information

Encoding images tested and works, video decoding works but uses too much memory (OOM on ROCm backend for 2s videos, even with --vae-conv-direct, I think sequential decoding should be implemented to lower the memory requirements. Works on CPU backend for longer videos, but not as quick)

## Checklist

### DIFF
--- a/src/model/vae/tae.hpp
+++ b/src/model/vae/tae.hpp
@@ -656,7 +656,7 @@ public:
     }
 
     ggml_tensor* encode(GGMLRunnerContext* ctx, ggml_tensor* x) {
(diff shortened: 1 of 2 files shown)


stduhpf:
I have managed to get the parallel version of taehv decoding working, but the graph size is scaling with the number of frames... I will try to make it work cleanly without exploding graph size and open another PR.

Edit: maybe the larger graph size isn't that bad after all.

PR #1875, opened 2026-08-12
Fix signed shape-product wrap in safetensors reader (DoS via NULL deref)

x14ngch3n:
## Summary

`read_safetensors_file` (`src/model_io/safetensors_io.cpp`) parses each tensor dimension with `shape[i].get<int64_t>()` and stores it directly into `ne[]` with no range or overflow check. `TensorStorage::nelements()` (`src/model_io/tensor_storage.h`) multiplies the dims as a signed `int64` product, and the reader uses `nbytes()` (derived from that product) for its only size check:

```cpp
tensor_size_ok = (tensor_storage.nbytes() == tensor_data_size);   // safetensors_io.cpp:346
// tensor_data_size = end - begin, already bounded against the file at :255
```

A crafted `.safetensors` with `shape = [4294967296, 4294967296]` makes the `int64` product overflow to **exactly 0**, so `nbytes() == 0`. Pairing it with an empty, in-bounds data range (`data_offsets [x, x]`) makes `0 == 0` pass the size check, and the tensor is accepted with huge `ne[]` but `nbytes() == 0`.

## Impact — ASAN-confirmed NULL deref (DoS)

`ggml_new_tensor` computes its allocation size from the same `ne[]` and also wraps to 0, returning a tensor whose `data` pointer is **NULL** (`ggml_nbytes == 0, data == 0x0`). The first consumer access then dereferences NULL → crash.

```
read_safetensors_file: ACCEPTED (BUG)
tensor 't': nelements()=0 nbytes()=0 ne=[4294967296,4294967296]
ggml_new_tensor OK: ggml_nbytes=0 data=0x0
==95833==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000
==95833==The signal is caused by a READ memory access.
==95833==Hint: address points to the zero page.
```

**Impact scope:** denial of service (crash) via a malicious model file. Only an exact-zero wrap is reachable — a near-wrap to a small nonzero product is infeasible with IEEE-double-representable dims (`a·b = N·2^64 + k` with both `a,b < 2^53`), so the consequence is a NULL-deref crash, not a heap out-of-bounds. Negative dimensions were also accepted at parse time and reached the `int64` product.

## The fix

Validate each parsed dimension (reject `<= 0`) and compute the element-count product with `__builtin_mul_overflow`, rejecting the tensor up front if any dim is non-positive or the product overflows. This mirrors the invariants the downstream ggml/tensor paths assume hold.

```cpp
for (int i = 0; i < n_dims; i++) {
    ne[i] = shape[i].get<int64_t>();
    if (ne[i] <= 0) { set_error(...); return false; }
}
// overflow-checked product
int64_t product = 1;
for (int i = 0; i < n_dims; i++) {
    if (__builtin_mul_overflow(product, ne[i], &product)) { set_error(...); return false; }
}
if (product <= 0) { set_error(...); return false; }
```

## Verification

AddressSanitizer (`-O1`) harnesses against this reader:
- `shape = [2^32, 2^32]` (was: ACCEPTED → NULL deref) now **REJECTED** — "tensor 't' has an overflowed/invalid shape product".
- `shape = [-1, 2]` (negative dim) now **REJECTED** — "dimension 0 is -1".
- legitimate tensor `[2, 2]` still **accepted** (`nelements=4`) — no regression.

Repro harness and PoC tensors available on request.

### DIFF
--- a/src/model_io/safetensors_io.cpp
+++ b/src/model_io/safetensors_io.cpp
@@ -272,23 +272,37 @@ bool read_safetensors_file(const std::string& file_path,
         int64_t ne[SD_MAX_DIMS] = {1, 1, 1, 1, 1};
         for (int i = 0; i < n_dims; i++) {
             ne[i] = shape[i].get<int64_t>();
-            if (ne[i] <= 0) {
+            if (ne[i] < 0) {
                 set_error(error, "invalid tensor shape for '" + name + "' (dimension " +
-                                     std::to_string(i) + " is " + std::to_string(ne[i]) + ")");
+                                     std::to_string(i) + " is negative)");
                 return false;
             }
         }
 
-        {
-            int64_t product = 1;
-            bool overflow   = false;
-            for (int i = 0; i < n_dims && !overflow; i++) {
-                if (__builtin_mul_overflow(product, ne[i], &product)) {
-                    overflow = true;
-                }
+        // nelements() is the product of the dims; nbytes() is
+        // nelements()*ggml_type_size(type)/ggml_blck_size(type). Both multiply
+        // silently, so a 2^62-element F32 tensor overflows only at the byte size
+        // (x4 -> 0) and reproduces the NULL-buffer deref this reader rejects.
+        // Check both with overflow detection; a zero extent is a valid empty
+        // tensor, so skip it rather than build a zero-byte storage.
+        int64_t nelems = 1;
+        bool overflow  = false;
+        for (int i = 0; i < n_dims && !overflow; i++) {
+            if (__builtin_mul_overflow(nelems, ne[i], &nelems)) {
+                overflow = true;
             }
-            if (overflow || product <= 0) {
-                set_error(error, "tensor '" + name + "' has an overflowed/invalid shape product");
+        }
+        if (overflow) {
+            set_error(error, "tensor '" + name + "' has an overflowed shape product");
+            return false;
+        }
+        if (nelems == 0) {
+            continue;
+        }
+        {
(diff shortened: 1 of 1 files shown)


x14ngch3n:
Thanks for the review — both points addressed in the latest push (4ed4053).

**P1 (byte-size overflow):** Added a second `__builtin_mul_overflow` check on `nelems * ggml_type_size(type)` before the `TensorStorage` is constructed. The element-count check alone does not cover `TensorStorage::nbytes() = nelements()*ggml_type_size(type)/ggml_blck_size(type)`, so a tensor whose element count fits in `int64_t` (e.g. a 2^62-element F32 tensor) overflowed only at the byte-size multiply (×4 → 0) and re-opened the zero-size path. The new check mirrors `nbytes()`'s numerator, so if it passes the internal multiply cannot wrap.

**P2 (zero dimensions):** Now rejects only negative dimensions (`ne[i] < 0`). A zero extent is a valid safetensors empty tensor (equal data offsets), so the `product <= 0` clause is dropped and a genuinely empty tensor (`nelems == 0`) is skipped — same pattern as the existing `U8` skip — rather than building a zero-byte storage that ggml would turn into a NULL buffer.

Builds clean (`Release`).

leejet:
Thanks for reporting this and working on a fix. The negative-dimension, shape-product overflow, and byte-size overflow checks are now covered by #1956 (6b47fec0).

The remaining change skips empty tensors, which can discard metadata such as `v_pred`, so we should preserve master's behavior here.

Closing this PR as superseded by #1956.

PR #1879, opened 2026-08-13
feat: support sequential execution for taehv

stduhpf:
## Summary

Keep the VRAM usage of video decoding for TAEHV lower by decoding (or encoding) frame-by-frame.

## Additional Information

I left in the code for parallel decoding, but it's always disabled, on all my tests it seemed like sequential is always faster and uses lower memory, but on some more powerful hardware it might not be the case.

## Checklist


Green-Sky:
Try doing this fancy new "pr stack" thingy.

PR #1883, opened 2026-08-16
fix: mark graph cuts with both a prefix and a suffix

### LINKED ISSUE #1865: MiniMax-H3: segmented offload (max_vram) deterministically produces an unrelated image (SD1.5 unaffected)
### Summary

With MiniMax-H3, setting `sd_ctx_params_t.max_vram` (segmented parameter offload) makes generation complete normally in the usual wall time but produce a **completely unrelated image** — deterministically. The same seed that renders "a red paper lantern drifting over dark water" without a budget renders grey storm clouds with birds with one. Byte-identical mp4 output across separate processes on different days, so it is a wrong computation, not a race.

SD1.5 under the same mechanism is fine (control below), so this looks specific to the H3 pipeline.

### Environment

- commit `ea7f0c8` (the MiniMax-H3 support commit), CUDA 13.1 backend, Windows 11, RTX 5070 Ti 16 GB
- via the C API (managed binding over `stable-diffusion.h`; `generate_video` path). `SDNET_DEBUG` param dumps confirm the struct reaches native intact.

### Model set

- `minimax_h3_fl2va_pruned-Q4_K_M.gguf` (leejet/MiniMax-H3-GGUF)
- `qwen3vl_32b_minimax_h3-Q4_K_M.gguf` (llm slot)
- `minimax_h3_video_vae_fp16.safetensors`, `minimax_h3_audio_vae_fp32.safetensors` (Comfy-Org)
- text encoder on CPU (`te=cpu` in the backend spec), flash attention on, VAE tiling on

### Repro

t2v, 864x480, 56 frames, 8 steps, cfg 1.0, seed 7, prompt "a red paper lantern drifting over dark water at night, warm glow, gentle ripples".

| `max_vram` | `stream_layers` | Result | Wall time |
|---|---|---|---|
| 0 (off) | off | correct (lantern over water) | ~6.6 min |
| 9 GiB | off | **wrong scene** (storm clouds, birds) | ~6.8 min |
| 9 GiB | on | **wrong scene, byte-identical to the row above** | ~6.7 min |

`stream_layers` is exonerated — the budget alone does it. The wrong output is stable: three runs across two days produced byte-identical files. Frame count, duration and the generated audio track are all correct; only the picture content is wrong, so nothing errors and nothing looks failed.

Weights report as `total params memory size = 35398.76MB (VRAM 16534.02MB, RAM 18864.74MB)` without a budget — i.e. the model already oversubscribes this 16 GB card via CUDA VMM and renders correctly that way; the corruption only appears when segmentation takes over placement.

### Control (mechanism works elsewhere)

SD1.5 Q4 (1.5 GB weights), 512x512 image, same build: `max_vram = 0.25 GiB` (6x oversubscribed) produces **byte-identical** output to the unsegmented render, at ~2x the time. So segmented offload itself is sound for the single-model case; the suspicion lands on H3's multi-component graph (second VAE / `token_refiner` blocks) interacting with the segment planner.

Happy to run diagnostics or test patches — this card depends on segmentation being usable for H3, since without it every step pays full VMM re-paging.

wbruna:
## Summary

Avoids collisions between ggml-generated names and graph cut marks, since most of those are made by adding a suffix.

I've opted for a minimal suffix to avoid issues with `GGML_MAX_NAME`, but it should work with any fixed string (provided it's not a suffix of `GGML_RUNNER_CUT_PREFIX` too).

Should fix #1865 .

## Checklist

### DIFF
--- a/src/core/ggml_graph_cut.cpp
+++ b/src/core/ggml_graph_cut.cpp
@@ -453,11 +453,12 @@ namespace sd::ggml_graph_cut {
         if (tensor == nullptr || tensor->name[0] == '\0') {
             return false;
         }
-        return std::strncmp(tensor->name, GGML_RUNNER_CUT_PREFIX, std::strlen(GGML_RUNNER_CUT_PREFIX)) == 0;
+        return starts_with(tensor->name, GGML_RUNNER_CUT_PREFIX) &&
+            ends_with(tensor->name, GGML_RUNNER_CUT_SUFFIX);
     }
 
     std::string make_graph_cut_name(const std::string& group, const std::string& output) {
-        return std::string(GGML_RUNNER_CUT_PREFIX) + group + "|" + output;
+        return std::string(GGML_RUNNER_CUT_PREFIX) + group + "|" + output + GGML_RUNNER_CUT_SUFFIX;
     }
 
     void mark_graph_cut(ggml_tensor* tensor, const std::string& group, const std::string& output) {
@@ -751,7 +752,9 @@ namespace sd::ggml_graph_cut {
 
             plan.has_cuts = true;
             std::string full_name(node->name);
-            std::string payload = full_name.substr(std::strlen(GGML_RUNNER_CUT_PREFIX));
+            size_t prefix_len   = std::strlen(GGML_RUNNER_CUT_PREFIX);
+            size_t suffix_len   = std::strlen(GGML_RUNNER_CUT_SUFFIX);
+            std::string payload = full_name.substr(prefix_len, full_name.size() - prefix_len - suffix_len);
             size_t sep          = payload.find('|');
             std::string group   = sep == std::string::npos ? payload : payload.substr(0, sep);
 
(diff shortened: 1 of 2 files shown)


farawayso:
Great fix! 👏

This is a clean, robust approach. Adding a fixed suffix (|) and requiring both prefix + suffix matching effectively turns cut-tensor identification from a fragile prefix-only heuristic into a proper format contract.

Compared to the one-liner blacklist approach in #1865 (excluding names containing " ("), this whitelist-based method is safer: ggml will never synthesize a trailing | suffix, so there's no risk of false positives. And | is a nice choice — it's already the group/output separator, so prefix:group|output| reads consistently, and the single-byte cost stays well within GGML_MAX_NAME.

Verified the logic against the #1865 root cause: view/cont derivatives end with ), not |, so ends_with correctly filters them out — the cache collision that produced byte-identical unrelated clips on H3 is resolved.

Thanks for the PR!

PR #1890, opened 2026-08-20
Fix ROCm ccache build caching and update to ROCm 10.0.0

superm1:
## Summary

Using mtime doesn't work with wheels, need to compare content.

## Related Issue / Discussion

This was reported in llama.cpp, it should be same problem here.

## Checklist


superm1:
@leejet with these changes CI will be much faster for multiple builds on the same day.

PR #1895, opened 2026-08-21
feat: support numbering for preview images

vmobilis:
## Summary

Allow to use %d placeholder in a filename template of preview images.
Hires previews will use numbering, continuous with base resolution.

Video previews are unaffected.

## Additional Information

Tested only on Android.
And thank you.

## Checklist

### DIFF
--- a/examples/cli/main.cpp
+++ b/examples/cli/main.cpp
@@ -400,20 +400,15 @@ void step_callback(int step, int frame_count, sd_image_t* image, bool is_noisy, 
     // is_noisy is set to true if the preview corresponds to noisy latents, false if it's denoised latents
     // unused in this app, it will either be always noisy or always denoised here
     if (frame_count == 1) {
-        std::string name = cli_params->preview_path;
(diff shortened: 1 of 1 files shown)


leejet (changes requested):
Two small things:

1. Could we only call format_frame_idx() when the filename actually contains a format specifier? Otherwise paths containing %% change even when numbering isn't requested.
2. Could we keep the preview index consistent with --output? --output starts from 000, while previews currently start from 001.

vmobilis:
@leejet, you're right, done:
1. added `%d` check;
2. changed `++counter` to `counter++`;
3. added default `.png` extension.

vmobilis:
@leejet, after some thought, I've applied substitution to whole pathname (not just filename), because the `save_results()` below does the same at [line 525](https://github.com/leejet/stable-diffusion.cpp/blob/50d640568388f876b0d63ee6ddb6bc86d997ec64/examples/cli/main.cpp#L525) and it would be less intuitive to use different rules for similar paths (as with 0 and 1-based counters).

leejet:

PR #1899, opened 2026-08-22
feat: support Brownian tree noise in all noise injection samplers

wbruna:
## Summary

Adds a new extra sampling parameter `noise_sampler=brownian_tree` to optionally use Brownian tree noise in LCM and all ancestral samplers. 

## Related Issue / Discussion

#1743 

## Checklist

### DIFF
--- a/examples/common/common.cpp
+++ b/examples/common/common.cpp
@@ -1008,7 +1008,7 @@ ArgOptions SDGenerationParams::get_options() {
          &hires_upscaler},
         {"",
          "--extra-sample-args",
-         "extra sampler/scheduler/guidance args, key=value list. CFG supports guidance_schedule; APG supports apg_eta, apg_momentum, apg_norm_threshold, apg_norm_threshold_smoothing; SLG supports slg_uncond; lcm supports noise_clip_std, noise_scale_start, noise_scale_end; flux supports base_shift, max_shift; ltx2 supports max_shift, base_shift, stretch, terminal; euler_ge supports gamma; beta scheduler supports alpha, beta; logit_normal supports mu, std, logsnr_min, logsnr_max, resolution_aware; lms supports lms_max_order, lms_shift, lms_divisions",
+         "extra sampler/scheduler/guidance args, key=value list. CFG supports guidance_schedule; APG supports apg_eta, apg_momentum, apg_norm_threshold, apg_norm_threshold_smoothing; SLG supports slg_uncond; lcm supports noise_clip_std, noise_scale_start, noise_scale_end; flux supports base_shift, max_shift; ltx2 supports max_shift, base_shift, stretch, terminal; euler_ge supports gamma; beta scheduler supports alpha, beta; logit_normal supports mu, std, logsnr_min, logsnr_max, resolution_aware; lms supports lms_max_order, lms_shift, lms_divisions; ancestral and lcm samplers support noise_sampler with value iid (default for most samplers) or brownian_tree",
          (int)',',
          &extra_sample_args},
         {"",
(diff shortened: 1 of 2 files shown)


vmobilis:
@wbruna, <s>maybe it's possible to make it more generic – as an option for `--sampler-rng` or even `--rng`?

Because it looks like `noise_sampler` is duplicating their functionality.</s>

Probably is not so good idea, because it does not substitute, but rather extends the rng.

vmobilis:
@wbruna, hello, I digged a little more, currently the `BrownianTreeNoiseSampler()` explicitly  uses `std::make_shared<STDDefaultRNG>()`:

https://github.com/wbruna/stable-diffusion.cpp/blob/sd_brownian_tree/src/runtime/denoiser.hpp#L2724
```
auto rng = std::make_shared<STDDefaultRNG>();
```

(Which is the `std` option of `--sampler-rng`):
https://github.com/wbruna/stable-diffusion.cpp/blob/sd_brownian_tree/src/stable-diffusion.cpp#L659
```
if (rng_type == STD_DEFAULT_RNG) {
    return std::make_shared<STDDefaultRNG>();
}
```
<br>

Can you pass the application-wide `rng`, as in `IIDGaussianNoiseSampler()`?

https://github.com/wbruna/stable-diffusion.cpp/blob/sd_brownian_tree/src/runtime/denoiser.hpp#L2702
```
return sd::Tensor<float>::randn(shape, rng);
```

Then it should allow to use `--sampler-rng` choice of `std`, `cpu` and `cuda` in Brownian extension.

wbruna:
Sounds sensible to me. But maybe @fszontagh adopted it to match results from ComfyUI? If so, we may want to keep it as the default for `dpm++2m_sde_bt`? (but then the rng choice would need to be moved from model loading time to an inference parameter - which also sounds sensible to me, but maybe a bit too much for this PR 🙂)

vmobilis:
@wbruna, the `std` option should pass exactly the same `<STDDefaultRNG>` as currently is in  `BrownianTreeNoiseSampler()`, so with `--sampler-rng std` the `noise_sampler=brownian_tree` should behave exactly as before, if I'm not mistaken.

And probably there is no need to move anything, because you are already using the app-wide `rng` in `IIDGaussianNoiseSampler()` and only need to pass it to `BrownianTreeNoiseSampler()` as well.

fszontagh:
@wbruna Not a ComfyUI parity thing - mine is a hand-written bridge, ComfyUI uses `torchsde.BrownianTree`, so the noise never matched anyway. No need to keep `STDDefaultRNG` as the default for `dpm++2m_sde_bt`.

One caveat: `bridge()` re-seeds per node, so passing the shared app RNG would re-seed it mid-sampling. Better to clone it. The hardcoded `std` also keeps the tree device-independent, which is worth preserving.

vmobilis:
@wbruna, hello, maybe something like https://github.com/wbruna/stable-diffusion.cpp/pull/3?

I didn't test all variants yet, but so far `iid` and `brownian_tree` look same and others produce different images.

The changes are minimal, the rng in `BrownianTreeNoiseSampler()` is [copied](https://github.com/wbruna/stable-diffusion.cpp/pull/3/changes#diff-c111bcb1e34ce9c65ef128c7c5a62e502b4bae06de90feaa66c458a031e0432dR2727), as @fszontagh suggested.

P.s. If you don't need it, just ignore, I've simply shown what I meant. 😉

leejet:
Thanks for the discussion. I’ve added `brownian_tree_rng` with these options:

- `cpu` (default)
- `cuda`
- `std_default`
- `sampler_rng`

Each tree owns an independent RNG. `sampler_rng` clones the selected sampler RNG once, so reseeding tree nodes cannot reset the shared RNG.

The default is now `cpu`, using the same host-side implementation regardless of the inference backend. This changes existing Brownian outputs; `std_default` preserves the previous node-noise generation. The Brownian bridge and tree-seed derivation remain unchanged.

I also made tree initialization lazy, so samplers that never request noise no longer consume RNG state unnecessarily.

PR #1915, opened 2026-08-27
feat: additional `--preview-interval` values

vmobilis:
## Summary

Additional values for `--preview-interval` option of `sd-cli`:
- value `0` should preview only the last step of base-resolution sampling.
Can be used in `--hires` generations to save (in `--preview-path`) the base-resolution result together with the high-resolution final image, skipping simultaneous high-resolution sampling + decoding.
With https://github.com/leejet/stable-diffusion.cpp/pull/1895, can save all base-resolution results of batch generations.
- values `less than 0` should decode only the `-N`th single image (to preview a sample and not to slow the generation down).

## Additional Information

For example, `--steps 20 --hires --preview-interval 20` and `--steps 20 --hires --preview-interval 0` make identical "base + high" images, but with less memory.
Peak memory usage (SD 1.5, 256×512 -> 448×896, CPU):
<table>
<tr><td>--preview</td><td>--preview-interval 0</td><td>--preview-interval 20</td></tr>
<tr><td>tae</td><td>3.03 Gb</td><td>3.72 Gb</td></tr>
<tr><td>vae</td><td>3.12 Gb</td><td>4.33 Gb</td></tr>
</table>

Also, I have enabled "last step" preview for "high noise" video pass, but I didn't test any video generations and not sure if this is much useful.

The `sd-server` should be unaffected because it doesn't set the preview interval and uses default "every step" value.

## Checklist

### DIFF
--- a/examples/cli/main.cpp
+++ b/examples/cli/main.cpp
@@ -36,6 +36,7 @@ struct SDCliParams {
     SDMode mode             = IMG_GEN;
     std::string output_path = "output.png";
     int output_begin_idx    = -1;
+    int compression_quality = 90;
     std::string image_path;
     std::string metadata_format = "text";
 
@@ -99,6 +100,10 @@ struct SDCliParams {
              "--output-begin-idx",
              "starting index for output image sequence, must be non-negative (default 0 if specified %d in output path, 1 otherwise)",
              &output_begin_idx},
+            {"",
+             "--compression-quality",
+             "compression quality of video and JPEG / WebP images (90 by default)",
+             &compression_quality},
         };
 
         options.bool_options = {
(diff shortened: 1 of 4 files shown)


vmobilis:
@leejet, could you please explain? If I uderstood correctly, in [d9b6e27/src/runtime/preview_interval.h](https://github.com/leejet/stable-diffusion.cpp/blob/d9b6e27e9f3203bc6d4ac35ff5fb5ef4b7db1032/src/runtime/preview_interval.h):
- the `logical_sample_step(step)` is `abs(step)`;
- the `sample_step_is_complete()` is `step > 0 || (terminal_sigma_is_zero && abs(step) == total_steps)`
(Because `step == 0` ≠ `total_steps > 0`).

So the logic is same, except the case when `step < 0`.
<s>When is it negative, I didn't find it?</s>

Got it, I forgot samplers with 2 denoising passes, like "auto denoised_opt = model(x, sigmas[i], <i><b>-(i + 1)</b></i>);".
Thank you very much!

PR #1966, opened 2026-09-12
perf: cache MiniMax H3 text conditioning

xledx:
## Summary

Cache MiniMax H3 text-only conditioning results so repeated requests with the same prompt can reuse the previously computed `SDCondition`.

The cache is intentionally conservative and is only enabled when:

- the model is MiniMax H3
- there are no effective LoRAs
- unconditional conditioning is not requested
- there are no image or MiniMax H3 presentation references

Non-cacheable H3 requests invalidate the cached condition.

## Related Issue / Discussion

None.

## Additional Information

Validated with MiniMax H3 on an Intel Mac, CPU-only, using `sd-server` so consecutive requests shared the same model context.

On a repeated identical text-only request, the runtime log reported:

```text
H3 conditioning cache hit
get_learned_condition completed, taking 0.00s


xledx:
Retested on the latest `master`.

* Confirmed the conditioning cache hit on an Intel Mac, CPU-only.
* `get_learned_condition`: **199.97s → 0.00s**
* Wall-clock generation time: **6m36s → 2m59s**
* Same prompt, seed, and generation settings were used for both runs.

The patch also cherry-picks cleanly onto the latest `master`, and both `sd-cli` and `sd-server` build successfully.

farawayso:
Validated the parameter surface: this is a pure internal cache of get_learned_condition (4 files, no coupling to --stream-layers/--max-vram), so it should apply cleanly on the current default-streaming CLI. One ROCm gfx1100 data point would de-risk it further — our 7900XTX setup hits the same get_learned_condition cost on CPU-side text conditioning. Happy to build from this branch and report the get_learned_condition before/after on ROCm if useful for the merge decision.

leejet:

PR #1995, opened 2026-09-18
fix: prevent clip_preprocess center crop from exceeding the resized image

akhenakh:
## Summary

`clip_preprocess` computed the aspect-preserving resize size by truncating `scale * dim` to `int64_t`. For some aspect ratios the single-precision result lands a hair below an integer (e.g. `736.0f / 730.0f * 730.0f == 735.999…`), so the resized side is **one pixel smaller than the crop target**. The center crop then reads one element past the end of the resized tensor and throws an uncaught exception:

```
sd::Tensor error: Tensor index out of range: shape=[650, 735, 3, 1]
terminate called after throwing an instance of 'std::invalid_argument'
  what():  Tensor index out of range: shape=[650, 735, 3, 1]
SIGABRT: abort
```

Reproduction (Qwen3VL / Krea2 reference images): a second reference of 645×730 is snapped to `w_bar=640, h_bar=736`; `scale = max(640/645, 736/730) = 1.008219`, so `resized_height = (int64_t)(scale * 730) = (int64_t)735.999… = 735`, while the crop loops up to `target_height = 736`. Any caller of `clip_preprocess` can hit this (Qwen3VL/Krea2 vision encoder, CLIP vision, PhotoMaker) — it only depends on the image aspect ratio.

The fix rounds the resized dimensions **up** and clamps them to at least the crop target, so the crop window is always inside the resized image:

```cpp
int64_t resized_width  = static_cast<int64_t>(std::ceil(scale * static_cast<float>(image.shape()[0])));
int64_t resized_height = static_cast<int64_t>(std::ceil(scale * static_cast<float>(image.shape()[1])));
resized_width  = std::max<int64_t>(resized_width, target_width);
resized_height = std::max<int64_t>(resized_height, target_height);
```

## Additional Information

Verification (exact failing setup):

- Model: `Krea-2-Turbo-Q8_0.gguf` + Qwen3VL text encoder, `--backend vulkan1,te=cpu`
- Args: `-W 960 -H 720 --steps 10 --cfg-scale 1 --img-cfg-scale 1 --flow-shift 3`
- Refs: `-r ref0.png` (960×720, i.e. 720×960 H×W) `-r ref1.png` (645×730) with `--ref-image-args preset=krea2_ostris_edit`
- Prompt: an edit instruction with two `Picture N:` references.

Before the fix:

```
[INFO] image.cpp:271  - EDIT mode
[INFO] conditioner resize ... ref image 1 from 730x645 to 736x640
sd::Tensor error: Tensor index out of range: shape=[650, 735, 3, 1]
[abort]
```

After the fix (same command, same refs):

```
[INFO] image.cpp:516  - get_learned_condition completed, taking 26.60s
[INFO] image.cpp:886  - sampling completed, taking 100.81s
[INFO] main.cpp:504  - save result image 0 to '...png' (success)
exit 0
```

- Platform: Arch Linux, RX 7900 XTX (RADV), Mesa 26.2.2, Vulkan 1.4.354.
- Formatting: `clang-format` clean (`.clang-format`, Chromium base).

## Checklist


leejet (changes requested):
I think the crash fix is valid, but using `ceil` here changes the resize result for inputs that were previously handled correctly.

For example, in the reported `645x730 -> 640x736` case, only the resized height needs to be clamped from `735` to `736`. Changing to `ceil` also changes the width from `650` to `651`, which unnecessarily alters the CLIP preprocessing result.

Could we keep the existing truncation behavior and only enforce the crop invariant instead?

```cpp
int64_t resized_width =
    static_cast<int64_t>(scale * static_cast<float>(image.shape()[0]));
int64_t resized_height =
    static_cast<int64_t>(scale * static_cast<float>(image.shape()[1]));

resized_width = std::max<int64_t>(resized_width, target_width);
resized_height = std::max<int64_t>(resized_height, target_height);
```

This should fix the out-of-bounds issue while minimizing behavior changes for existing inputs.

akhenakh:
thanks for catching the width regression. Updated to the clamp-only version

leejet (approved):

PR #2007, opened 2026-09-20
fix: preserve reference image dimensions in server requests

### LINKED ISSUE #2004: [Bug] sd-server decodes `ref_images` at the request's output size, while sd-cli keeps the reference's own size — models conditioned on a low-resolution reference cannot be served
### Git commit

17860c0

CacyOs

### GGML backends

HIP

sd-server --diffusion-model pid_1.5_flux2_1024_to_4096_4step_bf16.safetensors --llm gemma_2_2b_it_elm_bf16.safetensors --tokenizer tokenizer.json --vae flux2_ae.safetensors --vae-format flux2 --rng cpu --listen-ip 127.0.0.1 --listen-port 7861

### Steps to reproduce

The same generation gives two different pictures depending on whether it is run
through `sd-cli` or posted to `sd-server`, because the server resizes the
reference image to the request's output size before encoding it.

In the server's own log:

```
media_io.cpp:592 - resize input image from 384x384 to 1536x1536
[INFO] image.cpp - encode_first_stage completed, taking 6.30s
```

sd-cli on the identical inputs:

```
[INFO] image.cpp - encode_first_stage completed, taking 0.42s
```

For a model conditioned on a **low-resolution** reference — PiD takes a
reference at a quarter of the output side — this is not a cosmetic difference.
The model is handed a blurred, upscaled reference instead of the small sharp one
it expects, and returns a smear.

One tile, a 288² crop of a 512² source, 1152² output, seed 42, 4 steps, cfg 1,
`pid_1.5_flux2_1024_to_4096_4step`:

| run | Laplacian SD (sharpness) | result |
|---|---|---|
| `sd-cli` | 17.1 | sharp fur, whiskers, eyes |
| `sd-server`, same inputs | 11.4 | blocky smear, eyes barely there |
| plain bicubic ×4 of the crop | 2.8 | soft, for scale |

A reference image should reach the model at its own size unless the caller asks
otherwise — i.e. the `ref_images` decode should not take the output size as the
expected size, or `auto_resize_ref_image: false` should be honored at decode
time rather than only at the VAE step.

With this, I need to start a new sd-cli instance each time I want to use pid, paying the price of loading the model multiple times.
Since I have images that can be bigger than 1024x1024, I first do a tiling step, and then scale each tile, and then rebuild the full scaled image.

## Where it comes from

`examples/common/common.cpp`, `SDGenerationParams::from_json_str()`:

```cpp
if (!parse_image_array_json_field(j, "ref_images", 3, width, height, ref_images)) {
    LOG_ERROR("invalid ref_images");
    return false;
}
```

`width` and `height` here are the request's **output** size, and
`parse_image_array_json_field` passes them to `decode_base64_image` as the
expected dimensions, so every reference image is resampled to the output size at
decode time. `sd-cli` has no equivalent step and encodes the file as it is.

## What does not turn it off

- `"auto_resize_ref_image": false` in the request body. It only appends
  `resize_before_vae=0` to `ref_image_args` in `to_sd_img_gen_params_t()`, which
  acts later, inside the VAE step — the decode has already resized the image.
- `--ref-image-args "resize_before_vae=off"` at launch, for the same reason.
- The sdapi `extra_images` route and the OpenAI-style edits route: both pass the
  output size to the same decode once a size is given, and PiD always gives one.

_No response_

- **sd.cpp**: first observed on `master-859-7f410a3`, still reproducing on
  `master-872-cc515a0`, and the code path below is unchanged on master today
- **Backend**: ROCm / HIP, gfx1151 (AMD Radeon 8060S, Ryzen AI MAX+ 395), Linux
- **Model**: `Comfy-Org/PixelDiT` → `pid_1.5_flux2_1024_to_4096_4step_bf16.safetensors`

mikemikimike:
## Summary

- Preserve the source dimensions of `ref_images`, sdapi `extra_images`, and OpenAI edit uploads during decoding.
- Keep output-size decoding for init and mask images unchanged.

Issue: https://github.com/leejet/stable-diffusion.cpp/issues/2004
Fixes #2004

## Testing

- `cmake -S . -B build -DSD_BUILD_EXAMPLES=ON -DSD_SERVER_BUILD_FRONTEND=OFF -DSD_WEBP=OFF -DSD_WEBM=OFF -DCMAKE_BUILD_TYPE=Release`
- `cmake --build build --parallel 4`
- `sd-cli --help`
- `sd-server --help`


leejet (changes requested):
Please do not commit any test-related files.

mikemikimike:
@leejet The current head is `1a6d542b6df1d4f3b70eedb27ee574e3b52ecff9`. `gh pr diff --name-only` contains only `examples/common/common.cpp`, `examples/server/routes_openai.cpp`, and `examples/server/routes_sdapi.cpp`; the previously added test files are no longer present. No code changes were made in this pass. Please re-review.

examples/server/routes_openai.cpp
```diff
@@ -164,7 +164,7 @@ static bool build_openai_edit_request(const httplib::Request& req,
             reinterpret_cast<const char*>(bytes.data()),
             static_cast<int>(bytes.size()),
             img_w, img_h,
-            width, height, 3);
+            0, 0, 3);
```
leejet, line 167: The first reference image is also copied into init_image below. With this change, that copy retains the original dimensions and is later stretched to the output size by sd_image_to_tensor(), while the mask still goes through center-cropping and resizing during decoding.

When the uploaded image and mask have a different aspect ratio from the requested output, they therefore become misaligned, causing inpainting to affect the wrong region. I reproduced this with a 128×64 image/mask pair and a 64×64 output: they align before this change but not afterward.

Please keep the reference at its original size while separately applying the existing output-size crop/resize preprocessing to init_image, so it stays aligned with the mask.

leejet (changes requested):

leejet:

PR #2011, opened 2026-09-21
fix: honor reference image resize opt-out in server requests

AzizMuminov:
## Summary

- Make `auto_resize_ref_image: false` preserve native dimensions for native `ref_images` and SDAPI `extra_images`.
- Preserve the existing default behavior when `auto_resize_ref_image` is enabled or omitted.
- Keep the OpenAI edits route unchanged because it also establishes generic `init_image`/img2img semantics.

## Related Issue / Discussion

Related to #2004.

The server previously resized base64 reference images to the requested output size during decode. This occurred before the later `resize_before_vae` option was evaluated, so setting `auto_resize_ref_image: false` could not restore the original pixels. For PiD, which bypasses the later reference-VAE resize path, the prematurely upscaled reference was encoded directly.

This intentionally takes the compatibility-preserving approach: the existing opt-out now works, while default server requests retain their current behavior. It differs from #2007, which proposes preserving native reference dimensions by default.

## Verification

- Parser-level probe using a 1x1 PNG and a 1536x1152 request:
  - default behavior: reference is resized to 1536x1152;
  - `auto_resize_ref_image: false`: reference remains 1x1 through the generation request.
- `cmake --build build_issue2004 --target sd-server -j 4` with `SD_METAL=ON`.
- `git diff --check`.

Full PiD image-quality verification was not run because the required PiD model weights are not available locally.


leejet:

PR #2021, opened 2026-09-21
Add alpha channel input for Qwen Image 2.1 and relative docs

CarlGao4:
## Summary

Image inputs are no longer force-converted to RGB during decode in the CLI and server; they keep their native channel count. The generation pipeline already normalizes channels per model (`get_image_channels()` + `ensure_image_tensor_channels()`), so RGBA-capable models now receive the alpha channel while RGB-only models are unaffected.

Relative docs are also added.

## Additional Information

The decoder change affects all image and video input paths. Due to local VRAM limits video generation paths were not exercised locally and should be tested before merging. Theoretically this is not affected.

It seems that sometimes requiring the model to remove background may create some white squares, which I haven't find out the reason yet. Maybe comparing different quantizations (I have no enough VRAM too) or test the results of original `diffusers` could answer this.

## Checklist


CarlGao4:
Ohh I accidentally pushed [1014d0f](https://github.com/leejet/stable-diffusion.cpp/pull/2021/commits/1014d0f7c110a5f330d1cafa3bf27148a825189e) here, but if stop building docker images on branches is acceptable I'll just keep it here

leejet:

stduhpf:
I'm noticing the same opaque white patches in the areas that are supposed to transparent in [assets/qwen/qwen-image-2.1-alpha-out2.png](https://github.com/leejet/stable-diffusion.cpp/blob/master/assets/qwen/qwen-image-2.1-alpha-out2.png) and in my own attempts at generating transparent images. Is this a flaw of the model in general, or a bug in sd.cpp's implementation in particular?

CarlGao4:
@stduhpf What quantization are you using?

stduhpf:
Q8_0

CarlGao4:
I tried the demo on [huggingface space](https://huggingface.co/spaces/Qwen/Qwen-Image-2.1-workflow) which use diffusers directly without quantization. The white patches didn't exist.
image
workflow_bc47f08aa5bbc689

I'll try F16 on sd.cpp then

CarlGao4:
Interesting. I ran with original weights (transformer, vae, text encoder) without quantization, the results are same:

test-f16dit-f32vae-f16encoder

CarlGao4:
Almost there. Some last tests then I'll come to an answer and open another PR

PR #2033, opened 2026-09-23
feat: optimize cfg special cases with guidance schdeule

stduhpf:
## Summary

Avoid running unnecessary diffusion model forward passes when using guidance schedule with scale == 1 (or rather scale == img_cfg)

## Checklist


leejet (changes requested):
The CFG simplification is valid, but skipping these branches also skips work required by other features:

1. **Preserve APG behavior**  
   With APG momentum enabled, removing the unconditional prediction at scale 1 makes `AdaptiveProjectedGuidance::forward()` return before updating its momentum buffer. Subsequent guided steps therefore produce different results. Removing `uncond` in three-condition APG also changes the output formula after projection/clipping. Please restrict this optimization to plain CFG until these cases are handled.

2. **Keep the ControlNet computation needed by `img_uncond`**  
   When `text_cfg == img_cfg != 1`, skipping the `uncond` branch also skips computing negative-prompt controls. The remaining `img_uncond` pass then receives the positive-prompt controls instead. Its required ControlNet computation must remain even when the unconditional diffusion pass is skipped.

3. **Invalidate caches for skipped conditions**  
   With a schedule such as `2 → 1 → 2`, the positive cache can be refreshed while the negative cache remains unchanged. When guidance resumes, EasyCache/UCache can reuse the older negative residual based on the newer positive cache state. Please invalidate the skipped condition’s cache or force a refresh when it becomes active again.

stduhpf:
Like this?

leejet:
Thanks for the update! The APG guard and ControlNet fix look good.

I made a small follow-up locally to invalidate only the skipped condition’s cache, preserving the positive cache and current-step state. I also moved negative ControlNet preparation to the cache-miss path so resumed conditions receive the correct controls.

